PR #8, opened 2018-08-15
Develop

leem42:
added scss, push vars into state, updated readme to show markdown usage, fixed rendering list of markdown component

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ yarn-error.log*
 # visual studio specific 
 jsconfig.json
 .vscode
+
+# compiled css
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -19,3 +19,6 @@ yarn-error.log*
 # visual studio specific 
 jsconfig.json
 .vscode
+
+# compiled css
```
jay-hodgson: shouldn't the compiled css (from scss) be in the dist output directory?

jay-hodgson:
Looks like this change broke the travis build, can you investigate?

PR #9, opened 2018-08-20
Develop

leem42:
Fixed scrolling for references to only scroll with the markdown component that its contained


leem42:
merged versions to reflect patch update

PR #20, opened 2018-09-14
Develop

leem42:
Added a facets filter, synapse table view components, very basic horizontal stacked column chart component, created a HOC to tie together those all together, added more scss from SWC


src/lib/containers/Facets.js
```diff
@@ -1,25 +1,45 @@
...
+
+    render() {
+        const {element} = this.props
+        let children = []
+        let selectedFacets = this.props.selectedFacets
+        element.facetValues.forEach(
+            facetValue => {
+                children.push(
+                    <div key={facetValue.value + " " + facetValue.count}>
+                        <input defaultChecked={facetValue.isSelected} onClick={this.props.clickHandler({selectedFacets: selectedFacets, value: facetValue.value, columnName: element.columnName})} id={element.columnName + " " + facetValue.value + " " + facetValue.count} type="checkbox"/>
```
jay-hodgson: does the id need to be set to this?
element.columnName + " " + facetValue.value + " " + facetValue.count
if required, I think it precludes the possibility of having multiple tables being rendered on the same page (at least, it would be a violation of the html spec to have multiple elements with the same id).

src/lib/containers/FacetsExperimental.js
```diff
@@ -0,0 +1,119 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import * as SynapseConstants from 'lib/utils/SynapseConstants';
+
+export default class FacetsExperimental extends React.Component {
```
jay-hodgson: where does FacetsExperimental fit in?  I think I understand QueryWrapper, Facets, StackedRowHomebrew, SynapseTable.

src/lib/containers/Facets.js
```diff
@@ -36,112 +59,81 @@ export default class Facets extends React.Component {
...
-                sql,
-                limit: 1,
-                selectedFacets: selectedFacetsFormatted
-            },
-        };
-
-        this.props.updateQueryRequest(queryRequest)
+        this.setState({selectedFacets})
+        let queryRequest = this.props.getLastQueryRequest()
+        console.log("last query request ", queryRequest)
```
jay-hodgson: (console log)

PR #21, opened 2018-09-18
Develop

leem42:
Added simple tests for facets/table/query wrapper, added documentation to each of those


src/lib/containers/SynapseTable.js
```diff
@@ -1,21 +1,47 @@
...
+     *
+     * @memberof SynapseTable
+     */
+    handlePaginationClick = (eventType) => (event) => {
+        let queryRequest = this.props.getLastQueryRequest()
+        let currentOffset = queryRequest.query.offset
+        // if its a "previous" click subtract from the offset
+        // otherwise its next and we paginate forward
+        if (eventType === PREVIOUS)  {
+            currentOffset -= 25
```
jay-hodgson: 25 should probably be a const.

PR #24, opened 2018-09-25
Develop

leem42:
Made non SWC css classes use SRC prefix, changed inline styles to use classes, added column filter feature/bold column values on select, added advanced search, updated one snapshot

### DIFF
--- a/src/lib/containers/QueryWrapper.js
+++ b/src/lib/containers/QueryWrapper.js
@@ -95,7 +95,7 @@ export default class QueryWrapper extends React.Component {
         return (
             <div> 
                 {React.Children.map(this.props.children, child =>{
-                    return React.cloneElement(child, {facetCount: this.state.facetCount, alias: this.props.alias, executeQueryRequest: this.executeQueryRequest, getLastQueryRequest: this.getLastQueryRequest, data: this.state.data})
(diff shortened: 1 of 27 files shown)


src/lib/containers/QueryWrapper.js
```diff
@@ -95,7 +95,7 @@ export default class QueryWrapper extends React.Component {
         return (
             <div> 
                 {React.Children.map(this.props.children, child =>{
-                    return React.cloneElement(child, {facetCount: this.state.facetCount, alias: this.props.alias, executeQueryRequest: this.executeQueryRequest, getLastQueryRequest: this.getLastQueryRequest, data: this.state.data})
+                    return React.cloneElement(child, {synapseId: "syn15661198", facetCount: this.state.facetCount, alias: this.props.alias, executeQueryRequest: this.executeQueryRequest, getLastQueryRequest: this.getLastQueryRequest, data: this.state.data})
```
jay-hodgson: why is there a hard-coded synapse id in the query wrapper?

leem42: In the SynapseTable component it has to use the table id in the url to go to synapse for the advanced search button, so I could parse the select clause of the SQL instead, but I believe that we don't want to do this?

jay-hodgson: dev user should not have to modify QueryWrapper.  
Please either parse out of query (you're right that this could get complicated, but this ID should always be after the FROM keyword), or pass it in.

PR #31, opened 2018-10-09
Develop

leem42:
Made cards responsive, Added Icons to each of them
Synced colors between table and facets
Added menu
Added algo for black/white text decision
Updated docs

### DIFF
--- a/src/lib/containers/Facets.js
+++ b/src/lib/containers/Facets.js
@@ -3,13 +3,14 @@ import PropTypes from 'prop-types'
 import { library } from '@fortawesome/fontawesome-svg-core'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faTimesCircle } from '@fortawesome/free-solid-svg-icons'
+import { faPlus } from '@fortawesome/free-solid-svg-icons'
 const cloneDeep = require("lodash.clonedeep")
-const uuidv4 = require("uuid/v4")
 const SELECT_ALL = "select all"
 const DESELECT_ALL = "deselect all"
 
 // Add all icons to the library so you can use it in your page
 library.add(faTimesCircle)
+library.add(faPlus)
 
 
 /**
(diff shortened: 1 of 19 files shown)


src/lib/containers/Facets.js
```diff
@@ -14,28 +22,85 @@ const DESELECT_ALL = "deselect all"
...
+        if (event.target.tagName === "SPAN") {
+            event.target.className += " SRC-primary-background-color"
+        }
+    }
+    
+    handleHoverExit(event) {
+        if (event.target.tagName === "SPAN") {
+            let className = event.target.className
+            if (className.indexOf(" SRC-primary-background-color") !== -1 ) {
+                className = className.substring(0, className.indexOf(" SRC-primary-background-color"))
```
jay-hodgson: assumption here that the background color class is the last one listed (otherwise all other css classes will be lost).

src/lib/containers/Facets.js
```diff
@@ -14,28 +22,85 @@ const DESELECT_ALL = "deselect all"
...
-        let selectedFacets = this.props.selectedFacets
+
+        element.facetValues.sort((a,b) => {return b.count - a.count})
 
         element.facetValues.forEach(
             (facetValue, index) => {
                 let uniqueId = element.columnName + " " + facetValue.value + " " + facetValue.count
                 // caution when using uuId's to not cause extra re-renders from this always changing
-                let uuId = uuidv4()
+                let newR = this.props.RGB[0] * (1.3 - (1.0 / (index + 1)))
```
jay-hodgson: how many facet values does it take for this algorithm to level out pure white (or black)?

src/lib/containers/Facets.js
```diff
@@ -14,28 +22,85 @@ const DESELECT_ALL = "deselect all"
...
+                if (check) {
+                    style = {
+                        background: `rgb(${newR},${newG},${newB})` 
+                    }
+                } else {
+                    style = {
+                        background: `#C4C4C4`
+                    }
+                }
+                // https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color
```
jay-hodgson: nice!

src/lib/containers/QueryWrapper.js
```diff
@@ -74,44 +80,94 @@ export default class QueryWrapper extends React.Component {
...
-                    return React
-                                .cloneElement(  child, 
-                                                {
-                                                    executeQueryRequest: this.executeQueryRequest,
-                                                    getLastQueryRequest: this.getLastQueryRequest,
-                                                    data: this.state.data}
-                                            )
-                })} 
+            <div className="container-fluid"> 
+                <div className="col-xs-2">
```
jay-hodgson: if not showing the menu, you'll have a 2 columns dedicated to whitespace.

src/lib/containers/StackedRowHomebrew.js
```diff
@@ -269,19 +152,47 @@ export default class StackedRowHomebrew extends React.Component {
                             <div ref={measureRef}>
                                 {x_data.map(
                                     (obj, index) => {
+                                        let rgb = []
```
jay-hodgson: should have a shared function for these calculations (get new RGB and deciding white/black text)

src/lib/containers/Facets.js
leem42: As of the moment there is always going to be an almost pure black color chosen (index === 0 will set each RGB components to near zero which is why I used 1.3 to offset it from being pure black). Pure white isn't achievable with this algorithm, as index gets really large the original color will be chosen (it wont the the exact original color b/c of the 1.3 but close..). So I guess the color spectrum here is [black, original color chosen * 1.3]

So to answer the question:  pure black- 1 facet value, pure white- N/A

 I saw other ideas online that I can try for this too.

PR #32, opened 2018-10-09
Develop

leem42:
Added icons for study complete/active
Added filesize info for dataset card
Optimized menu tab switching- doesn't re-render if the chicklets aren't chosen on the current view.
Added loading bar option for stackedRowHomebrew
Updated README


src/lib/containers/StackedRowHomebrew.js
```diff
@@ -113,9 +113,9 @@ export default class StackedRowHomebrew extends React.Component {
      */
     render () {
         let {data} = this.props
-        // while data hasn't queued up display loading
-        if (data.length === 0) {
-            return (<div className="container">loading</div>)
+        // while loading
+        if (data.length === 0 || (this.props.isLoading && this.props.loadingScreen !== undefined )) {
+            return (this.props.loadingScreen ? this.props.loadingScreen : (<div className="container"> </div>))
```
jay-hodgson: Default loading UI screen?  "Loading..."?

src/lib/containers/row_renderers/Dataset.js
```diff
@@ -60,7 +61,7 @@ class Dataset extends React.Component {
                       columns.map(
                           column => {
                               return (
-                                  <div key={uuidv4()} className="col-md-2 col-sm-2">
+                                  <div key={uuidv4()} className="col-md-3 col-sm-3">
```
jay-hodgson: if specifying col-sm-3, that applies to all larger screens (col-md-3 unnecessary).

PR #33, opened 2018-10-10
Develop

leem42:
Refactored color calculation
improved styling
made deselect all


src/lib/containers/QueryWrapper.js
```diff
@@ -60,7 +62,7 @@ export default class QueryWrapper extends React.Component {
 
         if (prevProps.initQueryRequest.query.sql !== this.props.initQueryRequest.query.sql) {
             this.setState({isChecked: []})
-            this.executeQueryRequest(this.props.initQueryRequest, false)
+            this.executeQueryRequest(null, true)
```
jay-hodgson: I understand that you are resetting the request in this case, but why do you now send null instead of the initial query request?

PR #39, opened 2018-10-23
Develop

leem42:
added logic for user profile (view not in yet)
changed stacked bar chart behavior -- no loading on facet selection
added clickable links to each of the cards
added white or black text heuristic


src/lib/containers/SynapseTable.js
```diff
@@ -254,13 +254,42 @@ export default class SynapseTable extends React.Component {
...
+        let totalAllFalseCase = 0
+        let totalStandardCase = 0
+        for (let key in x_data) {
+             if (x_data.hasOwnProperty(key)) { 
+                 allFalse = allFalse || x_data[key].isSelected
+                 totalAllFalseCase += x_data[key].count 
+                 totalStandardCase += x_data[key].isSelected ?  x_data[key].count : 0
+            } 
+        }
+        let total = allFalse === false ? totalAllFalseCase: totalStandardCase
```
jay-hodgson: This allFalse logic is bewildering.  Think we need a rename and simplification here.
allFalse => anyTrue?
Then 
`let total = allFalse === false ? totalAllFalseCase: totalStandardCase` would become this, right?
`let total = anyTrue ? totalStandardCase : totalAllFalseCase`

src/lib/containers/UserBadgeView.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+const uuidv4 = require("uuid/v4")
+
+const UserBadgeView = ({data, loadingBar}) => {
```
jay-hodgson: Very nice, and since you have all of the profile data Ljubo can decide what should be shown (and style it).

PR #47, opened 2018-10-30
PORTALS-145 -> 151

leem42:


src/lib/containers/Menu.js
```diff
@@ -33,6 +35,21 @@ export default class Menu extends React.Component {
         }
     }
 
+    handleHover = (info) => (event) => {
+        if (!info.isSelected && event.target.tagName === "DIV") {
```
jay-hodgson: I'd pull this logic into another function and share in hover enter/exit functions (in case it changes in the future)

src/lib/style/Portal.css
```diff
@@ -410,3 +408,21 @@ button.SRC-roundBorder {
 
 .SRC-selected-table-icon {
   color: white; }
+
+.SRC-center-text {
+  text-align: center; }
+
+.SRC-text-title {
+  font-size: 18px !important; }
```
jay-hodgson: !important (x3)?!

PR #50, opened 2018-11-06
Ts/Develop

leem42:
Removed Extraneous Files
Removed Relative imports
Added Typescript Support
Added JSON response types for api calls to synapse


src/lib/containers/widgets/SynapsePlot.tsx
```diff
@@ -1,67 +1,81 @@
...
 import Plot from 'react-plotly.js';
-import PropTypes from 'prop-types'
-import * as SynapseConstants from '../../utils/SynapseConstants'
-import {getFullQueryTableResults} from '../../utils/SynapseClient'
+import * as SynapseConstants from '../../utils/SynapseConstants';
+import { getFullQueryTableResults } from '../../utils/SynapseClient';
 
-class SynapsePlot extends React.Component {
+type SynapsePlotProps = {
+    token: string | undefined
```
leem42: token? : string

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -0,0 +1,481 @@
...
+let synapseMath = require('markdown-it-synapse-math');
+
+type MarkdownSynapseProps = {
+    errorMessageView?: JSX.Element,
+    token?: string,
+    ownerId?: string,
+    wikiId?: string,
+    markdown?: string,
+    hasSynapseResources?: boolean
+    updateLoadState?: any,
```
jay-hodgson: missing comma and extra comma do not break build?

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -0,0 +1,481 @@
...
+    processMath() {
+        if (!this.markupRef.current) {
+            return;
+        }
+        // use regex to grab all elements
+        let mathExpressions = this.markupRef.current.querySelectorAll('[id^="mathjax-"]');
+        // go through all obtained elements and transform them with katex
+        mathExpressions.forEach((element: any) => {
+            katex.render(element.textContent, element, {
+                // @ts-ignore: The docs for katex report conflicting information
```
jay-hodgson: did you test math processing after ts conversion?

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -0,0 +1,481 @@
...
+        switch (widgetType) {
+            case 'buttonlink':
+                return this.renderSynapseButton(widgetparamsMapped);
+            case 'image':
+                return this.renderSynapseImage(widgetparamsMapped);
+            case 'plot':
+                return this.renderSynapsePlot(widgetparamsMapped);
+            case 'toc':
+                return this.renderSynapseTOC(originalMarkup);
+            default:
```
jay-hodgson: Based on wiki widget load statistics, next up should be support for:
- User/Team widget (small version of User/Team card)
- Join Team widget (complex because team may have access requirements)
- Team Member Count (easiest of them all!)

src/lib/containers/MarkdownSynapse.tsx
leem42: will do that now

leem42: confirmed that this will render math (had to add katex css files to the project), also confirmed that the type declarations are in fact inconsistent with the actual usage of katex so the ignore comment is necessary

PR #51, opened 2018-11-14
PORTALS-180 PORTALS-165

leem42:
Fixed study card links to use id field of table data
Fixed Show More button to have the right styling

### DIFF
--- a/src/lib/containers/row_renderers/Study.tsx
+++ b/src/lib/containers/row_renderers/Study.tsx
@@ -31,18 +31,19 @@ export default class Study extends React.Component<StudyProps, StudyState> {
 
     render() {
         const { data, schema } = this.props;
-        const projectName = data[schema.projectName];
-        const projectLeads = data[schema.projectLeads] && data[schema.projectLeads].split(";").join(" / ");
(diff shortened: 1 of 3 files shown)


src/lib/containers/row_renderers/Study.tsx
```diff
@@ -31,18 +31,19 @@ export default class Study extends React.Component<StudyProps, StudyState> {
...
-        const projectName = data[schema.projectName];
-        const projectLeads = data[schema.projectLeads] && data[schema.projectLeads].split(";").join(" / ");
-        let summary = data[schema.summary];
-        const diseaseFocus = data[schema.diseaseFocus];
-        const tumorType = data[schema.tumorType];
-        const projectStatus = data[schema.projectStatus];
-        const fundingAgency = data[schema.fundingAgency];
-        const dataStatus = data[schema.dataStatus];
-        const summarySource = data[schema.summarySource];
+        const projectName: string = data[schema.projectName];
```
jay-hodgson: We should push for projectName to be renamed to name (in the future, if this is built on top of a view then "name" is already a default column that is surfaced for each entity type).

PR #52, opened 2018-11-20
Ts/develop

leem42:
Added project/study/consortia cards for amp-ad along with their assets
linked files under bar chart to their respective place in synapse
added hide/show menu in app demo per the design of facets that have a remembered state -- this introduces breaking changes into the query wrapper


src/lib/containers/Menu.tsx
```diff
@@ -1,126 +1,153 @@
 import * as React from "react";
```
jay-hodgson: QueryWrapperNavBar???

src/lib/style/Portal.scss
```diff
@@ -428,6 +428,15 @@ li.SRC-table-anchor-chosen, li.SRC-table-anchor-chosen > a {
    background-color: #D53F77 !important;
  }
 
+ .SRC-primary-background-color-hover:hover {
```
jay-hodgson: Improve css selector specificity (usually I'd have a top level element id based selector, but you can't do that!) to remove all !important

jay-hodgson: try specifying element (like body) instead (this may override style for Michael B)

PR #58, opened 2018-11-29
update package-lock.json

jay-hodgson:
to include jquery types too, to fix travis build


jay-hodgson:
this takes us back to a node-sass build error

PR #60, opened 2018-11-30
Ts/develop

leem42:
Added View More button for cards
Different scheme for color gradient 
Added tooltip for bart
made show more text inline, added correct facet count selection on the show more button
added bold text for key contributors
made rough heuristic for calculating column width based on column name

### DIFF
--- a/src/lib/containers/QueryWrapper.tsx
+++ b/src/lib/containers/QueryWrapper.tsx
@@ -33,6 +33,7 @@ export type QueryWrapperChildProps = {
     isLoadingNewData?: boolean
     executeQueryRequest?: (param: QueryBundleRequest) => void
     executeInitialQueryRequest?: () => void
+    getNextPageOfData?: (queryRequest: any) => void 
     getLastQueryRequest?: () => QueryBundleRequest
     isChecked?: boolean []
     data?: QueryResultBundle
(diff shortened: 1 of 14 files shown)


src/lib/containers/QueryWrapper.tsx
```diff
@@ -152,6 +154,29 @@ export default class QueryWrapper extends React.Component<QueryWrapperProps, Que
...
+     * @memberof QueryWrapper
+     */
+    getNextPageOfData(queryRequest: any) {
+        this.setState({
+            isLoading: true
+        })
+        SynapseClient.getQueryTableResults(queryRequest, this.props.token)
+            .then(
+                (data: QueryResultBundle )=> {
+                    let oldData: QueryResultBundle = cloneDeep(this.state.data)
```
jay-hodgson: the name 'oldData' initially threw me off, since it really contains the new data query result rows, but I don't have a better suggestion (since it is initially a copy of the old state data)...

src/lib/containers/StackedRowHomebrew.tsx
```diff
@@ -264,29 +257,36 @@ export default class StackedRowHomebrew extends React.Component<StackedRowHomebr
...
-                                            <text className="SRC-text-title" fontFamily={"bold sans-serif"} fill={textColor} x={svgWidth / 2} y={svgHeight / 2 + 3}>
-                                                {index < 3 && obj.count}
-                                            </text>
-                                            {(this.state.index === index || initRender) && (
-                                                <text fill={originalColor} x={0} y={svgHeight + 15} className="SRC-text-shadow SRC-text-large">
-                                                    {"\u25BE"}
-                                                </text>
-                                            )}
-                                        </svg>
+                                                    <text className="SRC-text-title" fontFamily={"bold sans-serif"} fill={textColor} x={svgWidth / 2} y={svgHeight / 2 + 3}>
```
jay-hodgson: Setting the font weight makes sense, but changing the font family?  I'm surprised Ljubo would want a different font here.  In any case, could you create a css class name for this?

PR #62, opened 2018-12-04
Ts/develop

leem42:
Fixed alignment of menu and bar chart
made color selection stateless
removed links under bar chart


src/lib/containers/ColorGradient.tsx
```diff
@@ -1,40 +1,26 @@
 import {COLOR_PALETTE_EVEN, COLOR_PALETTE_ODD} from './colorPalette'
 
-export default class ColorGradient {
+export function getColorPallette (start: number, end: number): {colorPallete: string [], textColors: string []} {
```
jay-hodgson: *palette

PR #71, opened 2019-01-08
Ts/develop

leem42:
Style fixes, small html changes

### DIFF
--- a/src/lib/containers/StackedRowHomebrew.tsx
+++ b/src/lib/containers/StackedRowHomebrew.tsx
@@ -143,7 +143,7 @@ export default class StackedRowHomebrew extends
     const { index, hoverText } = this.state
     const hoverTextDisplay = index === -1 ? (xData[0] && xData[0].value) : hoverText
     return (
-      <span id="barChartHoverText">
+      <span>
         <span className="SRC-text-title">
           {this.props.filter}
         </span> :
--- a/src/lib/containers/StackedRowHomebrew.tsx
+++ b/src/lib/containers/StackedRowHomebrew.tsx
@@ -143,15 +143,15 @@ export default class StackedRowHomebrew extends
     const { index, hoverText } = this.state
     const hoverTextDisplay = index === -1 ? (xData[0] && xData[0].value) : hoverText
     return (
-      <React.Fragment>
-        <span className="SRC-text-cap">
+      <span id="barChartHoverText">
+        <span className="SRC-text-title">
           {this.props.filter}
         </span> :
-        <span>
+        <span className="SRC-text-title">
           {' '}
           {hoverTextDisplay === 'org.sagebionetworks.UNDEFINED_NULL_NOTSET' ? 'unannotated' : hoverTextDisplay}
         </span>
-      </React.Fragment>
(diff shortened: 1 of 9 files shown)


src/lib/containers/StackedRowHomebrew.tsx
```diff
@@ -141,16 +141,17 @@ export default class StackedRowHomebrew extends
 
   public getHoverText(xData: any) {
     const { index, hoverText } = this.state
-    const hoverTextDisplay = index === -1 ? xData[0] && xData[0].value : hoverText
+    const hoverTextDisplay = index === -1 ? (xData[0] && xData[0].value) : hoverText
     return (
-      <React.Fragment>
-        <span className="SRC-text-cap">
+      <span id="barChartHoverText">
```
jay-hodgson: Danger!  The [spec](https://www.w3.org/TR/html401/struct/global.html#h-7.5.2) indicates that the id must be unique across the doc.  If someone uses more than one of these in the same page, they'll have an issue.  If this is for css selectors, could you define a standard where portal devs need to add all SRC components inside an element of known ID (like id="syn-react-components")?  For our portals, this could be the body's id!

src/lib/style/Portal.scss
```diff
@@ -813,7 +827,7 @@ checkmark styling end
   display: none;
 }
 
-.SRC-paddingTopNoMargin {
+div.col-xs-2.SRC-paddingTopNoMargin {
```
jay-hodgson: Instead of targeting the grid layout class, perhaps add your own new class (in case the layout changes)?

src/lib/containers/StackedRowHomebrew.tsx
leem42: Ah right, since all default create-react-apps have their body tag as "root" I can target that, I will make sure to document this. Are you okay with this?

jay-hodgson: Sounds reasonable to me!  And if we get to a place where this project build can be exported for use in vanilla js apps in the browser (integration ready for Agora and SWC) then we'll just need to remember to designate an element with id="root"!

src/lib/containers/StackedRowHomebrew.tsx
```diff
@@ -341,6 +343,7 @@ export default class StackedRowHomebrew extends
           </p>
           <p
             className="SRC-noMargin SRC-padding-chart SRC-text-chart"
+            id="fileCount"
```
jay-hodgson: is this (id="fileCount") necessary?

src/lib/style/Portal.scss
```diff
@@ -533,11 +533,16 @@ li.SRC-table-anchor-chosen, li.SRC-table-anchor-chosen > a {
...
+  .SRC-text-title {
+    font-size: 18px;
+  }
+  .SRC-text-chart {
+    font-size: 16px;
+  }
 }
 
-.SRC-text-chart {
+.SRC-text-chart #fileCount {
```
jay-hodgson: oh, I see.  perhaps another css class for the selector?

PR #75, opened 2019-01-28
Hide Null Fields from cards

leem42:
Hide null fields on cards
Progress on UserBadge (will not be shipping with this PR)


jay-hodgson:
add test for null/unspecified/undefined/unset value case?

PR #76, opened 2019-01-28
PORTALS-367

jay-hodgson:


src/lib/containers/demo_components/Login.tsx
```diff
@@ -84,6 +84,7 @@ class Login extends React.Component<Props, State> {
     clickEvent.preventDefault() // avoid page refresh
     SynapseClient.login(this.state.username, this.state.password)
             .then((data: any) => {
+              SynapseClient.setSessionTokenCookie(data.sessionToken)
```
leem42: Would this need a check at all? Will setSessionTokenCookie ever fail? Or is the proof of concept?

src/lib/containers/demo_components/Login.tsx
```diff
@@ -187,6 +189,7 @@ class Login extends React.Component<Props, State> {
     if (code) {
       SynapseClient.oAuthSessionRequest(this.props.authProvider, code, `${this.props.redirectURL}?provider=${this.props.authProvider}`)
                 .then((synToken: any) => {
+                  SynapseClient.setSessionTokenCookie(synToken.sessionToken)
```
leem42: Same as above

src/lib/utils/SynapseClient.tsx
```diff
@@ -9,18 +9,25 @@ import { FaceFacetColumnValuesRequest } from './jsonResponses/Table/FacetColumnR
...
+    return text ? JSON.parse(text) : {}
+  })
+}
 const fetchWithExponentialTimeout = (url: string, options: any, delayMs: any, retries: number): Promise<any> => {
   return fetch(url, options)
     .then((resp) => {
       if (resp.status > 199 && resp.status < 300) {
         // ok!
-        return resp.json()
+        return parseJSON(resp)
```
leem42: Should all returns from this function have this same format? [e.g. this line below](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/76/files?utf8=%E2%9C%93&diff=split#diff-568e1d888d041ef55c0cd01da75ef9c7R45)

src/lib/utils/SynapseClient.tsx
```diff
@@ -81,14 +90,22 @@ export const doPost = (
...
   if (sessionToken) {
     options.headers.sessionToken = sessionToken
   }
   return fetchWithExponentialTimeout(endpoint + url, options, 1000, 5)
 }
-export const doGet = (url: string, sessionToken: string | undefined, endpoint: string) => {
+export const doGet = (
+                      url: string,
+                      sessionToken: string | undefined,
+                      initCredentials: string | undefined,
```
leem42: Is initCredentials a string or an object?

src/lib/utils/SynapseClient.tsx
```diff
@@ -113,7 +133,7 @@ export const getQueryTableResultsFromJobId = (
   sessionToken: string | undefined = undefined,
   endpoint: string = DEFAULT_ENDPOINT
 ): Promise<QueryResultBundle> => {
-  return doGet(`/repo/v1/entity/${entityId}/table/query/async/get/${jobId}`, sessionToken, endpoint)
+  return doGet(`/repo/v1/entity/${entityId}/table/query/async/get/${jobId}`, sessionToken, undefined, endpoint)
```
leem42: Shouldn't all the API calls receive a new parameter `initCredentials` that gets passed down? Otherwise it never gets set?

src/lib/containers/demo_components/Login.tsx
jay-hodgson: should not need to (response is empty, but instructs browser to set/clear cookie)

src/lib/utils/SynapseClient.tsx
jay-hodgson: probably

src/lib/utils/SynapseClient.tsx
jay-hodgson: only session cookie calls need the include credentials

PR #81, opened 2019-02-06
Ts/develop

leem42:
Fixed bug with cards loading and jumping around the page-- removed uuid calculation
Added more tests
Removed unused state/prop variables
Shared query logic between querywrapper and cardcontainer
put functions rewritten for rollup specifically in their own folder


src/lib/containers/CardContainer.tsx
```diff
@@ -57,81 +57,36 @@ export type CardContainerProps = {
...
   constructor(props: CardContainerProps) {
     super(props)
     this.handleViewMore = this.handleViewMore.bind(this)
-    this.getBufferData = this.getBufferData.bind(this)
-    this.gatherData = this.gatherData.bind(this)
     this.state = {
-      cardLimit: PAGE_SIZE,
-      hasLoadedBufferData: false,
-      hasMoreData: true
+      cardLimit: PAGE_SIZE
```
jay-hodgson: remove

PR #92, opened 2019-03-26
PORTALS-412: add click wall behind dropdown menu to capture clicks outside menu

jay-hodgson:
hide menu on click


leem42:
great!

PR #93, opened 2019-03-26
SWC-4767: fix build, and add prop to hide tooltip

jay-hodgson:


jay-hodgson:
Tried to use the disable prop for the ReactTooltip (https://www.npmjs.com/package/react-tooltip), but it's still shown.  Setting an empty data-tip seems to do the trick.  Please do not merge unless travis job succeeds.

src/__tests__/lib/containers/UserCardTests.tsx
```diff
@@ -156,11 +155,9 @@ describe('it creates the correct UI for the medium card', () => {
       }
     ] as MenuAction []
     const { wrapper } = createMediumComponent({ ...props, menuActions })
-    expect(wrapper.render().find('SRC-menu-wall')).toHaveLength(0)
```
leem42: looks like "." was missing from SRC-menu-wall class selector, adding that back in to be ".SRC-menu-wall" fixed the issue, I think we could keep check in there.

src/lib/containers/UserCardSmall.tsx
```diff
@@ -31,19 +32,21 @@ export const UserCardSmall: React.SFC<UserCardSmallProps> = (
...
-  }
-  if (userProfile.userName && !userProfile.userName.includes('TEMPORARY-')) {
-    label += ` (${userProfile.userName})`
-  }
-  if (userProfile.position) {
-    label += ` <br/>${userProfile.position}`
-  }
-  if (userProfile.location) {
-    label += ` <br/>${userProfile.location}`
+  if (!hideTooltip) {
```
leem42: we could do a conditional render on ReactToolTip using the hideToolTip variable, e.g. {
   !hideToolTip && <ReactToolTip..../>
} instead of putting that on the label

src/__tests__/lib/containers/UserCardTests.tsx
jay-hodgson: Good catch, I'll re-add it!

src/lib/containers/UserCardSmall.tsx
jay-hodgson: Yes, I tried that.  It still showed the tooltip!

leem42: ohh funky, just to clarify, my comment would not use the disable prop on the ReactTooltip, it would be a conditional render on the whole Reactooltip itself, that did not work?

jay-hodgson: Understood, and that's what I initially tried (mimicking conditional adding of span using hideText), but was surprised to still see the tooltip.

leem42: Gotcha, okay, well then all looks good hear, ill merge.

PR #101, opened 2019-04-02
SWC 4797, SWC 4799, PORTALS-424, SWC-4792

leem42:
SWC 4797 - photos getting stretched 
SWC 4799 - card line height
PORTALS-424 - text in card
SWC-4792 - letter is off center

fixed bug with prepublish script in package.json
split card content in half with metadata inside metadata for cards


src/lib/containers/UserCardSmall.tsx
```diff
@@ -51,12 +51,11 @@ export const UserCardSmall: React.SFC<UserCardSmallProps> = (
   if (preSignedURL) {
     marginLeft = '3px'
     img = (
-      <img
+      <div
+        className="SRC-userImgSmall"
+        style={{ borderRadius: '50%', backgroundImage: `url(${preSignedURL})` }}
```
leem42: can get removed

PR #113, opened 2019-04-11
PORTALS-434

leem42:


src/__tests__/lib/containers/Facets.test.tsx
```diff
@@ -87,224 +79,108 @@ describe('it performs basic functionality', () => {
...
-    //           ])
-    //         })
-    //       ])
-    //     })
-    //   })
-    // )
+  it('Onload all inputs are considered selected', async () => {
+    const { wrapper } = await createMountedComponent(cloneDeep(props))
+    const checkbox = wrapper.find(CheckboxGroup)
+    const inputs = checkbox.find(`input.SRC-hidden.SRC-facet-checkboxes.${FACET_SELECTED_CLASS}`)
```
jay-hodgson: suggest just using facet selected class (since class set may change)

src/__tests__/lib/containers/facetUtils.test.ts
```diff
@@ -0,0 +1,314 @@
+import {
+  getIsValueSelected,
+  readFacetValues,
+  SyntheticHTMLInputElement
+} from '../../../lib/utils/modules/facetUtils'
```
jay-hodgson: check on capitalization standard

PR #116, opened 2019-04-12
PORTALS-452

leem42:
added node-sass per the react docs, don't ignore css files anymore


src/lib/index.tsx
```diff
@@ -10,8 +10,8 @@ import UserCard from './containers/UserCard'
 import Login from './containers/Login'
 import CardContainer from './containers/CardContainer'
 import CardContainerLogic from './containers/CardContainerLogic'
-import './style/Cards.scss'
-import './style/Portal.scss'
+import './style/Cards.css'
```
leem42: node-sass up above doesn't actually fix the styling, its just recommended in the react docs. And even though .css is imported instead of .scss, scss still gets used during development since the file gets watched and written to css

PR #125, opened 2019-04-25
added isConsistent prop and show progress message when index builds

leem42:


src/lib/utils/SynapseClient.tsx
```diff
@@ -144,15 +145,19 @@ export const getQueryTableResultsFromJobId = (
...
   sessionToken: string | undefined = undefined,
-  endpoint: string = DEFAULT_ENDPOINT
+  endpoint: string = DEFAULT_ENDPOINT,
+  updateParentState?: any
 ): Promise<QueryResultBundle> => {
   return doGet(`/repo/v1/entity/${entityId}/table/query/async/get/${jobId}`, sessionToken, undefined, endpoint)
     .then((resp: any) => {
       // is this the job status?
       if (resp.jobState && resp.jobState !== 'FAILED') {
+        updateParentState && updateParentState({
```
jay-hodgson: Seems like this logic could be improved.  The service returns a http status code of 202 (accepted) in this case (accepted for processing but is incomplete).

leem42: The status code of 202 is returned for when the index is building and or if the table query is in 
progress without the index building, so I don't know that it adds clarity. Additionally, right now the doGet method doesn't pass forward the status code of the request (which could change of course) but that would have to get reworked.

PR #129, opened 2019-05-21
PORTALS-494

jay-hodgson:


src/lib/utils/LoginUtils.tsx
```diff
@@ -0,0 +1,48 @@
+import { SynapseClient } from '.'
```
leem42: Two small things: 
1. this file can be made a .ts file instead of .tsx since there's no jsx
2. I can't recall if I had an issue with this in the past but the build process might not like this _import from '.'_ syntax, instead it might want _import from './index'_

src/lib/assets/ButtonContent.tsx
```diff
@@ -1,7 +1,15 @@
 import * as React from 'react'
 
-export default ({ children, icon }: { children: any, icon: any }) => {
-  const style = { paddingRight: 10, fontWeight: 500, paddingLeft: icon ? 0 : 10, paddingTop: 10, paddingBottom: 10 }
+export default ({ children, icon }: { children: any, icon?: any }) => {
+  const style : React.CSSProperties = {
```
leem42: This is cool, I should use the React.CSSProperties type elsewhere in SRC

PR #131, opened 2019-05-22
add LoginUtils back in (and try to properly export), fix redirect url…

jay-hodgson:
…, and make this work in dev environment


src/lib/index.tsx
```diff
@@ -28,4 +29,4 @@ const SynapseComponents = {
   Markdown: MarkdownSynapse
 }
 
-export { SynapseClient, SynapseConstants, SynapseComponents }
+export { SynapseClient, SynapseConstants, SynapseComponents, SynapseLoginUtils }
```
jay-hodgson: move LoginUtils to SynapseClient

PR #133, opened 2019-05-24
PORTALS-496

leem42:


src/lib/containers/QueryWrapperMenu.tsx
```diff
@@ -99,9 +122,19 @@ export default class QueryWrapperMenu extends React.Component<QueryWrapperMenuPr
   public render() {
     const menuDropdown = this.renderFacetMenu()
     const queryWrapper = this.renderQueryChildren()
-
+    const name = window.location.hash.substring('#/Explore/'.length) || 'Grants'
```
jay-hodgson: substring of 10??? ('#/Explore/'.length)

leem42: I can change that, I was thinking 'that would be too hardcoded' but now that I think about it, thats no more hard coded than the actual string itself

leem42: whoops, I definitely didn't mean for grants to be the default value

PR #135, opened 2019-05-28
Added PUT call to refresh session token

leem42:
Also added if case for 204 status with empty response body


src/lib/utils/SynapseClient.tsx
```diff
@@ -516,6 +558,18 @@ export const getSessionTokenFromCookie = () => {
   const sessionToken = localStorage.getItem(DEV_ENV_SESSION_LOCAL_STORAGE_KEY)
   return Promise.resolve(sessionToken)
 }
+/**
+ * Get the current session token from a cookie.  Note that this will only succeed if your app is running on
+ * a .synapse.org subdomain.
+ */
+export const refreshSessionToken = () => {
```
jay-hodgson: I'm confused, this looks like a copy of getSessionTokenFromCookie (above it)

src/lib/utils/SynapseClient.tsx
```diff
@@ -144,6 +148,44 @@ export const doGet = (
...
+    options.headers.sessionToken = sessionToken
+  }
+  return fetchWithExponentialTimeout(endpoint + url, options, 1000, 5)
+}
+
+export const putRefreshSessionToken = (sessionToken: string, endpoint: string = DEFAULT_ENDPOINT) => {
+  return doPut('/auth/v1/session', { sessionToken }, undefined, undefined, endpoint)
+}
+
+export const putSessionTokenAndTermsOfUse = (
```
jay-hodgson: Please remove the ability to set terms of use from this utility.  A portal developer may want to use it, and they should not set this to true unless the user agrees to the full Synapse ToU (so the portal would need to properly present it).

src/lib/utils/SynapseClient.tsx
leem42: whoops, that was an accident

PR #138, opened 2019-05-31
PORTALS-430

leem42:
Implemented mocks for new generic card rendering
- Includes GenericCard component
- Changed CardFooter drastically to render single column of key-values using a table with table rows, on mobile the key values render to individual rows so it all wraps
Fixed bug with loading where if index was building the initial load was happening then the async job status wouldn't show
Fixed bug so async job status only shows when index is building
Changed prop loadingBar -> loadingScreen to maintain consistency across all components
Fixed whitespace in a few places


src/lib/containers/CardContainer.tsx
```diff
@@ -117,10 +126,17 @@ export class CardContainer extends React.Component<CardContainerProps, CardConta
...
+      type,
+      isLoading,
+      loadingScreen,
+      secondaryLabelLimit
     } = this.props
     if (data === undefined || Object.keys(data).length === 0) {
-      return (<div/>)
+      return (
+        <div>
+         {(isLoading && loadingScreen !== undefined) && loadingScreen}
```
leem42: isLoading && loadingScreen?

src/lib/style/Cards.scss
```diff
@@ -234,4 +233,25 @@ $text-color-lighter: #898989;
 
 img.iconImg.SRC-datasetIcon {
   max-width: 60px !important;
+}
+
+@media only screen and (min-width: 600px) {
```
leem42: use two classes and set in cardfooter based on mobile size

PR #144, opened 2019-06-05
PORTALS-489

jay-hodgson:


package.json
```diff
@@ -61,6 +62,7 @@
     "react-test-renderer": "^16.4.2",
     "react-tooltip": "^3.9.2",
     "react-transition-group": "2.6.0",
+    "rss-parser": "^3.7.1",
```
leem42: We'll need to make sure the single file build works with this new package in one of two three -
1. It bundles itself with that package included 
2. We add the package to rollup and add a CDN to a UMD bundled version
3. Or we know that this won't ever go in SWC and can safely ignore it

src/lib/containers/RssFeed.tsx
```diff
@@ -0,0 +1,56 @@
...
+type RssState = {
+  rssFeed: rssParser.Output,
+  isLoadingError: boolean
+}
+
+export type RssFeedProps = {
+  url: string
+}
+
+export default class RssFeed extends React.Component<RssFeedProps, RssState> {
```
leem42: I believe this class needs to get exported in the main index.ts file, I don't see that anywhere in this PR?

PR #150, opened 2019-06-13
PORTALS-525: Add basic Synapse multi-file multipart file upload support to SRC

jay-hodgson:


rollup.config.js
```diff
@@ -76,8 +76,8 @@ export default {
 		]),
 		// Common js is used to handle the import of older javascript modules not using es6 
 		commonjs(),
-		// minify the bundle with uglify
-		uglify()
```
jay-hodgson: uglify was not happy with es6 classes defined in md5 library

src/lib/containers/Uploader.tsx
```diff
@@ -0,0 +1,156 @@
...
+      })
+    }).catch((error: any) => {
+      this.handleUploadError(error)
+    })
+  }
+
+  render() {
+    return (
+      <div>
+        <input
```
leem42: For full clarity I think this could have a comment explaining that the input field is used only functionally, the button acts as the real UI.

src/lib/containers/Uploader.tsx
```diff
@@ -0,0 +1,156 @@
...
+      {
+        isUploading: isStillUploading,
+        filesUploadedCount: incrementedFilesUploadedCount,
+        successfullyUploaded: !isStillUploading
+      })
+  }
+
+  handleUploadError = (error: any) => {
+    this.finishedProcessingOneFile()
+    this.setState({ error: { error } })
```
leem42: I believe this should be this.setState({error}) ? otherwise the state contains a key 'error' who's value is a dictionary with a key named 'error' and the actual error value

src/lib/containers/Uploader.tsx
```diff
@@ -0,0 +1,156 @@
...
+    Array.from(selectorFiles).forEach((file) => {
+      // check for existing filename in parent folder before upload (add new version if exists).
+      // note that the parent container (project/folder) is configurable.
+      const newFileEntity: FileEntity = {
+        parentId: this.props.parentContainerId,
+        name: file.name,
+        concreteType: 'org.sagebionetworks.repo.model.FileEntity',
+        dataFileHandleId: '',
+      }
+      lookupChildEntity(
```
leem42: the spacing is a little bit hard to read here, what if the dictionary passed to lookupChildEntity was set as a variable -

const lookup = {  entityName: file.name, parentId: this.props.parentContainerId } 
and then it would read as lookupChildEntity(lookup, this.props.token).then( .... )

src/lib/containers/Uploader.tsx
```diff
@@ -0,0 +1,156 @@
...
+          this.state.successfullyUploaded &&
+          <span style={{ marginLeft: '10px' }}>
+            Successfully uploaded {this.state.totalFilesToUploadCount} file(s) to
+            <a style={{ marginLeft: '2px' }} href={`https://www.synapse.org/#!Synapse:${this.props.parentContainerId}`} target="_blank">{this.props.parentContainerId}</a>
+          </span>
+        }
+        {
+          this.state.error &&
+          <p>
+            Upload error!
```
leem42: Could add the error msg as well - Upload error! {this.state.error}

src/lib/utils/SynapseClient.tsx
```diff
@@ -14,6 +23,7 @@ const DEFAULT_ENDPOINT = 'https://repo-prod.prod.sagebase.org/'
 const DEFAULT_SWC_ENDPOINT = 'https://www.synapse.org/'
 
 export const AUTH_PROVIDER = 'GOOGLE_OAUTH_2_0'
+export const SYNAPSE_STORAGE_LOCATION_ID = 1
```
leem42: Could write a comment here 'This hardcodes the storage location to S3'

src/lib/utils/jsonResponses/MultipartUploadStatus.tsx
```diff
@@ -0,0 +1,11 @@
+// https://docs.synapse.org/rest/org/sagebionetworks/repo/model/file/MultipartUploadStatus.html
+export type MultipartUploadStatus = {
+  uploadId: string,
+  startedBy: string,
+  startedOn: string,
+  updatedOn: string,
+  partsState: string,
+  state: 'UPLOADING' | 'COMPLETED' // The state of this file upload
+  resultFileHandleId: string,
+  clientSidePartsState?: boolean[]
```
leem42: Could write a comment explaining that clientSidePartsState this is a utility to read over the partsState which is represented by the string with 1s and 0s

src/lib/utils/SynapseClient.tsx
```diff
@@ -608,3 +638,245 @@ export const signOut = () => {
...
+  request: MultipartUploadRequest,
+  fileUploadResolve: (fileUpload: FileUploadComplete) => void,
+  fileUploadReject: (reason: any) => void,
+  endpoint: string = DEFAULT_ENDPOINT) => {
+  const url = 'file/v1/file/multipart'
+  doPost(url, request, sessionToken, undefined, endpoint).then(async (status: MultipartUploadStatus) => {
+    // started the upload
+    // keep track of the part state client-side
+    const clientSidePartsState: boolean[] = []
+    for (let i = 0; i < status.partsState.length; i = i + 1) {
```
leem42: this could get rewritten as 
const clientSidePartsState: boolean[] = status.partsState.split('').map(bit => bit === '1')

src/lib/utils/SynapseClient.tsx
```diff
@@ -608,3 +638,245 @@ export const signOut = () => {
...
+  fileUploadResolve: (fileUpload: FileUploadComplete) => void,
+  fileUploadReject: (reason: any) => void,
+  endpoint: string = DEFAULT_ENDPOINT
+) => {
+  if (multipartUploadStatus.clientSidePartsState![partNumber - 1]) {
+    // no-op. this part has already been processed!
+    return
+  }
+
+  const uploadID = multipartUploadStatus.uploadId
```
leem42: this could get renamed to upload**Id** and then in presignedUploadUrlRequest you could use the shorthand for uploadId, so 
{
contentType: 'application/octet-stream', // each part is binary
partNumbers: [partNumber],
uploadId
}

src/lib/utils/SynapseClient.tsx
```diff
@@ -608,3 +638,245 @@ export const signOut = () => {
...
+    uploadId: uploadID
+  }
+  const presignedUrlUrl = `/file/v1/file/multipart/${uploadID}/presigned/url/batch`
+  doPost(presignedUrlUrl, presignedUploadUrlRequest, sessionToken, undefined, endpoint).then(
+    async (presignedUrlResponse: BatchPresignedUploadUrlResponse) => {
+      const presignedUrl = presignedUrlResponse.partPresignedUrls[0].uploadPresignedUrl
+      // calculate the byte range
+      const startByte = (partNumber - 1) * request.partSizeBytes
+      let endByte = partNumber * request.partSizeBytes - 1
+      if (endByte >= request.fileSizeBytes) {
```
leem42: Is this a sanity check or could this happen? If its a sanity check then it might be better to at least console.error() the situation?

src/lib/utils/SynapseClient.tsx
```diff
@@ -608,3 +638,245 @@ export const signOut = () => {
...
+              checkUploadComplete(
+                multipartUploadStatus,
+                fileName,
+                sessionToken,
+                fileUploadResolve,
+                fileUploadReject,
+                endpoint)
+            } else {
+              // retry after a brief delay
+              delay(1000).then(() => {
```
leem42: Does this ever fail entirely? I'm sure there's lots of edge cases to handle, but this seems like one of the more fragile situations,  a TODO: Handle if filepart doesn't upload after x number of attempts could be good

src/lib/utils/SynapseClient.tsx
```diff
@@ -608,3 +638,245 @@ export const signOut = () => {
...
+      includePreSignedURLs: true,
+      includePreviewPreSignedURLs: false,
+      requestedFiles: fileHandleAssociationList
+    }
+    getFiles(request, sessionToken, endpoint).then(
+      (data: BatchFileResult) => {
+        const presignedUrl = data.requestedFiles[0].preSignedURL
+        const fileHandle = data.requestedFiles[0].fileHandle
+        // sanity check!  must be less than 5MB
+        if (fileHandle.contentSize < 5242880) {
```
leem42: 5242880 could be a constant

src/lib/utils/SynapseClient.tsx
jay-hodgson: Could have a max limit, but in general we just want to keep trying until it works.

src/lib/utils/SynapseClient.tsx
jay-hodgson: This is actually usually the case.  The last file part size will usually be less than the max part size.

PR #151, opened 2019-06-14
PORTALS-523

leem42:


src/lib/style/Cards.scss
```diff
@@ -270,3 +265,45 @@ img.iconImg.SRC-datasetIcon {
     white-space: nowrap;
   }
 }
+
+// TODO: keep this scss as the source of truth
+.SRC-portalCardHeader {
+  color: white;
+  font-family: 'Lato', sans-serif;
```
jay-hodgson: use variable?  does this need to be set at all (or is it default)?

PR #154, opened 2019-06-20
PORTALS-527: detect GROUP BY sql, gives ability to show underlying row data

jay-hodgson:
(only matching headers to column models are used to set the additional WHERE condition clauses, so aggregate function column values are ignored)


src/lib/containers/SynapseTable.tsx
```diff
@@ -248,6 +259,67 @@ export default class SynapseTable extends React.Component<QueryWrapperChildProps
...
+
+  public getSqlUnderlyingDataForRow(
+    selectedRow: Row,
+    originalSql: string,
+    headers: SelectColumn[],
+    columnModels: ColumnModel[]
+  ): { synId: string, newSql: string } {
+    const tokens: string[][] = lexer.tokenize(originalSql)
+    // remove all tokens after group (by truncating array)
+    tokens.length = tokens.findIndex(el => el[0] === 'GROUP')
```
leem42: instead of truncating with length, you could instead use the slice method - 
 
tokens = tokens.slice(0, tokens.findIndex(el => el[0] === 'GROUP'))

src/lib/containers/SynapseTable.tsx
```diff
@@ -113,6 +116,13 @@ export default class SynapseTable extends React.Component<QueryWrapperChildProps
     return facetAliases[facetName] || facetName
   }
 
+  public isAggregate(): boolean {
+    if (this.props.getLastQueryRequest && this.props.getLastQueryRequest().query) {
```
leem42: this whole function could instead be - 
return AGGREGATE_REGEX.test(this.props.getLastQueryRequest!().query.sql)

src/lib/containers/SynapseTable.tsx
jay-hodgson: Read somewhere that truncating using the length parameter was faster, but
1.  Using slice is more clear IMO
2.  For this particular function, we're calling it once, so even if it was faster we should not sacrifice readability.
Made the suggested change!

src/lib/containers/SynapseTable.tsx
jay-hodgson: Wasn't sure if the last query request would always be set when this component was initially rendered, but I'll take your word for it!

PR #157, opened 2019-06-21
PORTALS-527: now the count column(s) values are links

jay-hodgson:


src/lib/containers/SynapseTable.tsx
```diff
@@ -473,24 +502,25 @@ export default class SynapseTable extends React.Component<QueryWrapperChildProps
           // past the initial load -- when a user has started clicking items, then isColumnSelected is
           // not null and we verify that this column is part of the selection.
           const isColumnActivePastInitLoad = isColumnSelectedLen !== 0 && this.state.isColumnSelected[j]
+          const isCountColumn = countColumnIndexes.includes(j)
           if (isColumnActiveInitLoad || isColumnActivePastInitLoad) {
             return (
               <td className="SRC_noBorderTop" key={`(${i}${columnValue}${j})`}>
-                  <p className={`${index === -1 ? '' : 'SRC-boldText'}`}>{columnValue}</p>
```
leem42: just a note to myself - index should get renamed to something like 'isBeingSortedOn', otherwise it gives no indication of what it does...

PR #163, opened 2019-06-27
PORTALS-478

leem42:
Removed RowContainer -- was unnecessary
Added prop to cards for internal linking -- takes in a base URL and set of columnValues to generate the URL
Fixed bug in sqlFunctions, AND needed to get pushed to front of current token and operator was incorrect


src/lib/containers/GenericCard.tsx
```diff
@@ -70,8 +72,24 @@ export default class GenericCard extends React.Component<GenericCardProps, Gener
...
+      const urlParams = internalLinkConfiguration.columnValues.map(
+        (el, index) => {
+          if (!schema.hasOwnProperty(el)) {
+            console.error(`Could not find match for data: ${data} with columnName ${el}`)
+          }
+          const stringEnd = index < columnValuesLength - 1 ? '&' : ''
+          return `${el}=${data[schema[el]]}${stringEnd}`
+        }
+      ).join('')
+      // tested this link on the browser, there's no need to encode the URL, the browser picks up on that automatically
```
jay-hodgson: did you have a chance to verify on Chrome + FF + Safari?  should be ok, but just want to make sure!

PR #165, opened 2019-07-04
SWC-4905

jay-hodgson:
Adds dynamic form UI component, based on form schema (and ui schema) held in Synapse Files.  On successful save, uploads form data to new Synapse File (and executes callback containing SynID of that file)


src/lib/containers/EntityForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  token?: string, // user's session token
+  synIdCallback: (synId: string) => void, // callback.  Once the form output has been saved to a FileEntity, will send synID back
+}
+type EntityFormState = {
+  error?: any,
+  isLoading?: boolean,
+  successfullyUploaded: boolean,
+  containerId?: string,
+  userprofile?: UserProfile,
+  successMessage?: string,
```
jay-hodgson: needed?

PR #176, opened 2019-07-18
PORTALS-569, PORTALS-570, PORTALS-568, PORTALS-443

leem42:
PORTALS-569: Fix by adding title prop to cards allowing the card to hide the title if there is no data returned from the backend, also changed triangle under bar chart to a dot by using a different unicode character
PORTALS-570: Changed the padding from the outer container to be set by inner container of the menu, so the scollbar doesn't get pushed inside as well with the padding being set
PORTALS-568: Used max-width instead of width


src/lib/containers/SynapseTable.tsx
```diff
@@ -725,29 +725,25 @@ export default class SynapseTable extends React.Component<QueryWrapperChildProps
...
-                />
-                <span className="checkmark" />
-              </label>
-              <span ref={ref}>
-                {this.renderFacetSelection(facetColumnResult, ref, columnName)}
-              </span>
-            </ul>
-
-          </div>
+        <div className={`dropdown-menu dropdown-menu-override SRC-minDropdownWidth ${classList}`}>
```
leem42: removed unnecessary div

src/lib/containers/CardContainer.tsx
```diff
@@ -189,7 +197,8 @@ export class CardContainer extends React.Component<CardContainerProps, CardConta
 
     return (
       <div>
-        {(unitDescription && showBarChart) &&
+        {title &&  <h2 className="SRC-title">{title}</h2>}
+        {(!title && unitDescription && showBarChart) &&
           <TotalQueryResults
```
jay-hodgson: TotalQueryResults visibility depends on title (not being set)?

leem42: Yes, at least in the current set of mocks there's no view that has both a title and a card count

PR #188, opened 2019-08-17
PORTALS-584: small ui cleanup

jay-hodgson:


jay-hodgson:
Subscribe UI changed to this: Screen Shot 2019-08-16 at 5 30 28 PM

PR #200, opened 2019-09-03
feat: (CNT): drug nomination tool file grid

conundrumgirl:
First PR for Compound Nomination Tool
This is a new component that lists user's files within user's directory in  a given project.
If the directory doesn't exists it gets created.
User able to navigate to a new submission (TODO) by clicking 'new' button or edit a current one (TODO) populated with the data from the file if clicking on the file.
User is able to delete the file by clicking on the trash icon if they confirm that modal.
Some styles are TBD.
Some functionality is still in progress. 
This is not a final version
Comp:
image

Code Output
image


conundrumgirl:
@leem42 @jay-hodgson

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,286 @@
...
+      isLoading: true,
+      token: this.props.token,
+      fileList: []
+    }
+  }
+
+  componentDidMount(): Promise<any> {
+    // Note:  All portals should do this once on the initial app load.
+    // This looks for the session token cookie (HttpOnly, unable to directly access), and initialize the session if it does exists.
+    SynapseClient.detectSSOCode()
```
jay-hodgson: Should be removed.  As you note, SRC demo hands down session token in prop, portals do this during app initialization.

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,286 @@
...
+      return result;
+    }
+  }
+  onError = (args: any) => {
+    console.log(args)
+  }
+
+
+  // creates data folder for user
+  createDataFolder = async (userprofile: UserProfile, token: string): Promise<string | undefined> => {
```
jay-hodgson: TODO: create folder and create ACL (see https://sagebionetworks.jira.com/browse/PLFM-5692) allowing only the current user access (private folder containing their form data).

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const request = {
+      includeTypes: ['file'],
+      parentId: targetFolderId,
+      sortBy: 'NAME',
+      sortDirection: 'ASC'
+    }
+
+    try {
+      const data = await SynapseClient.getEntityChildren(request, token);
+      return data.page;
```
jay-hodgson: Looks like this is only getting the first page of results.  Maybe this is ok for initial release, but we should confirm (and likely add a "Show more" button to get the next page of data).  Are previously submitted items shown?  If not, then there are a few possible solutions.  You create a subfolder (Submissions) and move the entity into there after successful submission (then they would not be returned by your getChildren query).  Or we could utilize annotations.

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,286 @@
...
+        isLoading: false
+      })
+    }
+  }
+
+  /* ------------------------------------------   rendering fns  ------------------------------------------------*/
+  /*  login control if user is not logged in */
+  renderLogin = (stateToken: string = ''): JSX.Element => {
+    let token: string | undefined = ''
+    const loggedInState: JSX.Element = <div className='bg-success text-center' role='alert'>
```
jay-hodgson: Also to be removed.  Sign out functionality handled by other components (header nav bar dropdown in portals context).

jay-hodgson:
On Submit, still need to:
- Use SynapseClient.submitToEvaluation() to send the form data file to the evaluation queue (which means we need to pass an evaluation queue ID to the configuration of this component!).
- Indicate that the Synapse File has been submitted (move, or annotate?)

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: Yes. I will add it to the TODO: in the top of UserFileGrid.tsx

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: You are correct. This is still outstanding functionality.  Will try to get it into v1. This is somewhat of a WIP. Do you have a preference as to subfolder vs annotation?

conundrumgirl:
@jay-hodgson, Thank you for outlining the submit flow. I was going to ask you about it:)  There is another component that will handle loading the configuration data and will handle save and submit as well.  Probably will have the PR for it early next week

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+  constructor(props: UserFileGridProps) {
+    super(props)
+    this.state = {
+      isLoading: true,
+      token: this.props.token,
+      fileList: []
+    }
+  }
+
+  componentDidMount(): Promise<any> {
```
leem42: As was already mentioned, token should come in as a prop and then componentDidMount would call refresh like it already is except it would pass in this.props.token, and then componentDidUpdate would check if the token has changed from the previous props and call refresh if the token has changed. And once that gets changed, I don't believe  componentDid(Mount/Update) and refresh should return a value?

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+      })
+
+      this.refresh(this.state.token);
+  }
+
+
+  refresh = (token?: string): Promise<any> => {
+    this.setState({ token })
+    if (token) {
+      return this.getUserFileListing(token!, this.props.parentContainerId).catch((error) => {
```
leem42: the not null assert on token shouldn't be required since you've got the check on line 76 to see that token is defined

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+  private getUserFileListing = async (token: string, parentContainerId: string): Promise<UserData> => {
+    this.setState({ isLoading: true });
+    const result: UserData = {
+      fileList: [],
+      existingData: {},
+      dataFolderId: ''
+    }
+
+    // get user profile
+    const userprofile = await SynapseClient.getUserProfile(token)
```
leem42: camelCase on userProfile, and this could get wrapped in a try/catch -- Although it really shouldn't fail, I think it doesn't hurt to have more error catching.

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+      return <></>
+    }
+  }
+
+  renderFileTable = (fileList: Entity[], pathpart: string, dataFolderId?: string): JSX.Element => {
+    if (!dataFolderId) {
+      this.onError('Data Folder ID is undefined')
+      return <></>
+    }
+    if (_.isEmpty(fileList)) {
```
leem42: is there a reason why you can't check fileList.length === 0 for this?

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+          <th>Submitted On</th>
+          <th>Action</th>
+
+        </tr>
+      </thead>
+      <tbody>
+        {fileList.map((entity, i: number) => {
+          return (<tr className='test' key={i}>
+            <td><Link className='nav-green' to={`/${pathpart}/${dataFolderId}/${entity.id}/`}>{entity.name}</Link></td>
+            <td>{moment(entity.modifiedOn).calendar()}</td>
```
leem42: I believe this can use the browser built in Date library, instead of `moment(entity.modifiedOn).calendar()`, it would be `new Date(entity.modifiedOn).toLocaleDateString()`

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+          <th>Submission Name</th>
+          <th>Edited On</th>
+          <th>Submitted On</th>
+          <th>Action</th>
+
+        </tr>
+      </thead>
+      <tbody>
+        {fileList.map((entity, i: number) => {
+          return (<tr className='test' key={i}>
```
leem42: For key prop you could use `entity.id` to have a stable key

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+            <td><Link className='nav-green' to={`/${pathpart}/${dataFolderId}/${entity.id}/`}>{entity.name}</Link></td>
+            <td>{moment(entity.modifiedOn).calendar()}</td>
+            <td>?</td>
+            <td><button className='btn' aria-label="delete" onClick={() => this.setState({ modalContext: { action: this.deleteFile, arguments: [this.state.token!, entity.id!] } })}><FontAwesomeIcon icon={faTrash} aria-hidden="true" ></FontAwesomeIcon></button></td>
```
leem42: I see the argument to have an inline onClick handler here but I think it would be more readable to put that in a separate method

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,269 @@
...
+      {!(this.state.isLoading) &&
+        <div>
+          <h3>Your Submissions</h3>
+          <div className='panel panel-default padding-full'>
+            {this.renderFileTable(this.state.fileList, this.props.pathpart, this.state.dataFolderId)}
+            <div className='text-center'>
+              <a className='btn btn-success' href={`/${this.props.pathpart}/${this.state.dataFolderId}`}>Add new Compound</a></div>
+          </div></div>}
+      {this.state.modalContext &&
+        <WarningModal show={true} title={this.modalTitle} copy={this.modalCopy} callbackArgs={this.state.modalContext.arguments} onCancel={() => this.setState({ modalContext: undefined })} onOK={(x: string, y: string) => this.deleteFile(x, y)}></WarningModal>}
```
leem42: When we were using tslint there was a rule that when a component had a lot of props (~3+), to place them on their own line so its a bit more readable, I will look into eslint and see about enabling that back again.
```JSX
<WarningModal 
      show={true}
      title={this.modalTitle}
      copy={this.modalCopy}
      callbackArgs={this.state.modalContext.arguments} 
      onCancel={() => this.setState({ modalContext: undefined })} 
      onOK={(x: string, y: string) => this.deleteFile(x, y)}>
</WarningModal>
```

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: fewer keystrokes:) I will change it though

src/lib/utils/SynapseClient.ts
```diff
@@ -531,6 +554,13 @@ export const updateEntity = (
   return doPut(url, entity, sessionToken, undefined, endpoint)
 }
 
+export const deleteEntity: GetEntity = <T>(sessionToken: string | undefined = undefined,
```
leem42: The GetEntity Type could be better named, not sure what though..

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: The specific design request is along the lines of '3 days ago', '12 days ego' etc... ie more verbal. If you want to avoid additional dependency we can push back on that I don't think it's a major req.

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: out of curiosity, why does it matter?

src/lib/containers/drug_upload_tool/WarningModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  copy: string | JSX.Element
+  confirmCopy?: string
+  show: boolean
+  onOK: Function
+  onCancel: Function
+  callbackArgs: any
+}
+
+export default function WarningModal(props: WarningModalProps) {
+  const [show] = useState(props.show);
```
leem42: it doesn't seem like `show` is being set anywhere, so the modal component could set `show` from props directly ```JSX <Modal show={props.show}.../>```and then in UserFileGrid on line 264 you can set ```JSX <WarningModal show={this.state.modalContext}.../>``` and not do conditional rendering since `show` already defines whether WarningModal gets rendered or not

src/lib/utils/SynapseClient.ts
conundrumgirl: Agree. I used whatever was there for get. Will rename once we decide:)

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
leem42: If I understand this tool/interaction correctly, this _shouldn't_ really matter with what will happen, but we've had some pretty big bugs in the past with things re-rendering a lot because of setting the key to an index or random number (it was in a tutorial online, literally the worst thing you could set the key prop to be..) so we just always use a stable key so react doesn't re-render and drop the page/content because it thinks something new is on the screen.

src/lib/containers/drug_upload_tool/WarningModal.tsx
conundrumgirl: good point. was left over from old implementation. will remove

src/__tests__/lib/containers/drug_upload_tool/UserFileGrid.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+describe('basic tests', () => {
+ 
+  const props: UserFileGridProps = {
+    parentContainerId,
+    token,
+    pathpart,
+  }
+
+  beforeEach(() => {
```
leem42: nice tests and mocks for all the calls!

src/__tests__/lib/containers/drug_upload_tool/UserFileGrid.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const SynapseClient = require('../../../../lib/utils/SynapseClient')
+const targetFolderId = 'syn9988882982'
+const token: string = '123444'
+const pathpart = 'someTool'
+const parentContainerId: string = 'syn20355732'
+
+
+const createShallowComponent = async (props: UserFileGridProps, disableLifecycleMethods: boolean = false) => {
+  const wrapper = await shallow(
```
leem42: can do  `const wrapper = await shallow<UserFileGrid>(...) ` so you can remove the cast below on line 27

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: Returning the value is mostly for the sake of testing.
I think that way when we await for it we know that the promise it is calling has resolved?
I will just set it as async

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+  constructor(props: UserFileGridProps) {
+    super(props)
+    this.state = {
+      isLoading: true,
+      token: this.props.token,
+      fileList: []
+    }
+  }
+
+  async componentDidMount(){
```
leem42: componentDidMount doesn't need to be async

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  async componentDidUpdate(prevProps: any) {
+    const shouldUpdate = this.props.token !== prevProps.token
+    if (shouldUpdate) {
+      await this.refresh(this.props.token)
+    }
+  }
+
+  async refresh (token?: string) {
+    this.setState({ token })
```
leem42: I would remove token from state and only use `this.props.token`, copying props to state is an anti-pattern, (with a few exceptions)

src/lib/containers/drug_upload_tool/WarningModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+  show: boolean
+  onOK: Function
+  onCancel: Function
+  callbackArgs: any
+}
+
+export default function WarningModal(props: WarningModalProps) {
+ 
+  return (
+    <Modal show={true} animation={false}>
```
leem42: should accept the show prop instead of passing true-- so show={props.show} and then [this line](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/200/files#diff-cc39a698514630d6fd1a46f8127f8d0cR257) can read as `show={this.state.modalContext}`

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+      token: this.props.token,
+      fileList: []
+    }
+  }
+
+  async componentDidMount(){
+    await this.refresh(this.state.token)
+  }
+
+  async componentDidUpdate(prevProps: any) {
```
leem42: can add type --- `prevProps: UserFileGridProps`

src/__tests__/lib/containers/drug_upload_tool/WarningModal.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    title: 'Some Warning',
+    copy: 'Some Copy',
+    confirmCopy: 'Do it!',
+    show: true,
+    onOK: mock.okFn,
+    onCancel: _.noop,
+    callbackArgs: ['one', 'two']
+  }
+
+  beforeEach(() => {
```
leem42: this could get removed, I don't think its running any code?

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -0,0 +1,267 @@
...
+        this.onError(error)
+      })
+    }
+  }
+
+
+
+  private getUserFileListing = async (token: string, parentContainerId: string): Promise<UserData> => {
+    this.setState({ 
+      token: token,
```
leem42: same as above -- remove token from setState

PR #202, opened 2019-09-05
PORTALS-654, PORTALS-655

jay-hodgson:


src/lib/containers/RssFeed.tsx
```diff
@@ -78,6 +103,13 @@ export default class RssFeed extends React.Component<RssFeedProps, RssState> {
               }
             }
             let isItemVisible: boolean = index < this.props.defaultItemsToShow || this.state.isShowingMoreItems
+            
+            if (!this.state.itemId2MoreItem[item.guid]) {
```
leem42: Setting state in render is not a good practice, this should happen in componentDidMount, this works because this.state.itemId2MoreItem is a dictionary that's reference gets updated so its bypassing react's state tracking.

But -- I don't think either is necessary, in your method onToggleReadMore() you can do this check to see if item.guid has an object held in state on [line 44](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/202/files#diff-9f97b1490012385f47e3e5c9f9a7bfa4R44), if not then you know its yet to be clicked, so you can lazily update the state for each item.

jay-hodgson: updated render to use the values from the state if present, or defaults if missing.  in the toggle function (click event handler), it uses the default (if not yet in the state).  Fixed?

PR #204, opened 2019-09-06
feat(CNT): NavButtons, Header, and StepsSideNav components

conundrumgirl:
Adding 3 new subcomponents
image

### DIFF
--- a/src/__tests__/lib/containers/drug_upload_tool/NavButtons.test.tsx
+++ b/src/__tests__/lib/containers/drug_upload_tool/NavButtons.test.tsx
@@ -82,7 +82,7 @@ describe('NavButtons tests', () => {
       expect(wrapper.find('button.prev')).toHaveLength(0);
     });
 
-    it('if not wizard, only display of not the first step', () => {
+    it('if not wizard, only display if not the first step', () => {
       let { wrapper } = createShallowNavButtonsComponent(props);
       expect(wrapper.find('button.prev')).toHaveLength(1);
       const _props = { ...props, ...{ currentStep: stepsArray[0] } };
--- a/src/__tests__/lib/containers/drug_upload_tool/NavButtons.test.tsx
+++ b/src/__tests__/lib/containers/drug_upload_tool/NavButtons.test.tsx
@@ -1,0 +1,143 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import * as _ from 'lodash';
+import {
+  Step,
+  StepStateEnum,
+  NavActionEnum
+} from '../../../../lib/containers/drug_upload_tool/types';
+import {
+  NavButtons,
+  NavButtonsProps,
+  NextStepLink,
+  NextStepLinkProps
+} from '../../../../lib/containers/drug_upload_tool/NavButtons';
+
+const stepsArray: Step[] = [
+  {
+    id: 'toxicology _data',
+    order: 60,
+    title: 'Toxicology Data',
+    default: 'ld50',
+    static: true,
+    inProgress: true,
+    state: StepStateEnum.TODO,
+    rules: []
+  },
+  {
+    id: 'ld50',
+    order: 61,
+    title: 'LD50',
+    inProgress: false,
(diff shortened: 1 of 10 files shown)


conundrumgirl:
@leem42 @jay-hodgson 
If you have time early next week please

src/lib/containers/drug_upload_tool/Header.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+
+export type HeaderProps = {
+  copy?: string;
```
leem42: It looks like the variable `copy` is an html string, it seems like `html` would be a more fitting name?

src/lib/containers/drug_upload_tool/StepsSideNav.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const nonChildSteps = props.stepList.filter(step => step.child !== true);
+
+  const getIcon = (
+    state: StepStateEnum,
+    isExcluded: boolean | undefined,
+    isStatic: boolean | undefined
+  ): { iconDef: IconDefinition; flip: FlipProp | undefined } => {
+    let iconDef: IconDefinition = faCircle;
+    let flip: FlipProp | undefined = undefined;
+    if (isStatic) {
```
leem42: is the `isStatic` checked needed?

src/__tests__/lib/containers/drug_upload_tool/NavButtons.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    it('if wizard, only display if previousStepIds array is not empty', () => {
+      const _props = { ...props, ...{ isWizardMode: true } };
+      let { wrapper } = createShallowNavButtonsComponent(_props);
+      expect(wrapper.find('button.prev')).toHaveLength(1);
+      _props.previousStepIds = [];
+      wrapper = createShallowNavButtonsComponent(_props).wrapper;
+      expect(wrapper.find('button.prev')).toHaveLength(0);
+    });
+
+    it('if not wizard, only display of not the first step', () => {
```
leem42: small typo: only display _if_ not

src/lib/containers/drug_upload_tool/NavButtons.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <div className="buttonWrapper pull-right">
+        {previousButton} {nextButton} {saveButton}
+      </div>
+    </div>
+  );
+}
+
+export function NextStepLink(props: NextStepLinkProps): JSX.Element {
+  const nextStep = props.steps.find(step => step.id === props.nextStepId);
+  if (typeof nextStep === 'undefined') {
```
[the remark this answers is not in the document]
conundrumgirl: more explicit

src/lib/containers/drug_upload_tool/StepsSideNav.tsx
conundrumgirl: yes for icon rendering

src/lib/containers/drug_upload_tool/Header.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+
+export type HeaderProps = {
+  bodyText?: string;
```
jay-hodgson: should probably at least add a comment that this should not be user-defined html (that the bodyText should always be hard-coded) if using dangerouslySetInnerHTML with the value.

src/lib/containers/drug_upload_tool/NavButtons.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return <></>;
+  }
+  return (
+    <span className="nav-link">
+      <a onClick={e => props.onNavAction(nextStep)}>{nextStep.title}&nbsp;</a>{' '}
```
jay-hodgson: lot of odd spaces here (nbsp and a ' '?).  could you just set margin right instead of these solutions?

PR #218, opened 2019-09-19
PORTALS-720

leem42:
added search params prop to DrugUploadTool, changed demo app to read in search params


src/demo/containers/App.tsx
```diff
@@ -109,96 +109,65 @@ export default class App extends React.Component<{}, AppState> {
           <Route
             exact={true}
             path="/drugUploadTool"
-            render={props => (
+            render={props => {
+              let searchParamsProps: any = {}
+              // https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams -- needs polyfill for ie11
+              const searchParams: any = new URLSearchParams(props.location.search)
```
conundrumgirl: would Iterable<[string, string]> work as a type here?

src/demo/containers/App.tsx
```diff
@@ -109,96 +109,65 @@ export default class App extends React.Component<{}, AppState> {
...
+              const iter = searchParams.entries()
+              let result = iter.next()
+              while (!result.done) {
```
conundrumgirl: why not forEach here?

src/lib/containers/drug_upload_tool/DrugUploadTool.css
```diff
@@ -0,0 +1,164 @@
+/* MODAL OVERWRITE! */
```
conundrumgirl: wo we need to track .css files at all?

src/lib/containers/drug_upload_tool/DrugUploadTool.tsx
```diff
@@ -9,16 +9,19 @@ import Alert from 'react-bootstrap/Alert';
...
   // Provide the parent container (folder/project), that should contain a folder (named <user_id>) that this user can write to.
   parentContainerId: string; // ProjectId
   formSchemaEntityId: string; // Synapse file that contains the form schema.
   formUiSchemaEntityId: string; // Synapse file that contains the form ui schema.
   formNavSchemaEntityId: string; //Synapse file that consists screen nav schema
   token?: string; // user's session token
-  currentFileEntityId?: string; //entityId fo the file that contains form data
-  currentFileParentEntityId: string;
-
+  searchParams?: SynapseEntityContainer
```
conundrumgirl: nice!

src/lib/containers/drug_upload_tool/DrugUploadTool.css
leem42: we shouldn't have to, its been on the backlog to remove that, theres one thing in the build script that has to get changed so we can ignore those files. I think the change is to build the css and then the project but right now its the opposite.

src/demo/containers/App.tsx
leem42: Didn't see that method as an option, used it now, thanks!

PR #219, opened 2019-09-19
Portals 560

conundrumgirl:
This has several enhancements and updates. 
1. Added a toggle to only show help when it's toggled
image

2. Cleaned up the file gird and added another parameter to have the 'add new ...' button text configurable through property
image

3. fixed the formatting on the summary page
image

4. added some logic for the hidden 'included' field to accommodate for 'required' properties on conditional screens

### DIFF
--- a/src/lib/containers/drug_upload_tool/DrugUploadForm.tsx
+++ b/src/lib/containers/drug_upload_tool/DrugUploadForm.tsx
@@ -468,6 +468,33 @@ export default class DrugUploadForm extends React.Component<
     return <></>;
   };
 
+
+  renderHelpToggle =(currentStep: Step, showHelp: boolean, callbackFn: Function): JSX.Element => {
+    if (currentStep.static  || currentStep.final) {
+     return <></>
+    }
+     return (
+        <>
+          <label className="pull-right toggle-help-label">
+            <span>Hide help</span>
+            <Switch
+              checkedIcon={false}
+              uncheckedIcon={false}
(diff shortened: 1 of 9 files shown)


src/lib/containers/drug_upload_tool/DrugUploadForm.tsx
```diff
@@ -101,10 +102,23 @@ export default class DrugUploadForm extends React.Component<
...
+      doShowHelp: true
     };
   }
 
+  componentDidMount() {
+    if(_.isEmpty(this.state.formData)) {
+      const result = {};
+      const defs = this.props.schema.definitions;
+      Object.keys(defs).forEach((key: string) => {
+        if (defs[key].properties && Object.keys(defs[key].properties).indexOf('included') > -1) {
```
leem42: How come this is required, this seems like something that would fall on the data curation side?

src/lib/containers/drug_upload_tool/SummaryTable.tsx
```diff
@@ -104,34 +104,41 @@ export default function SummaryTable(props: SummaryTableProps): JSX.Element {
...
-              </td>
-              <td>{line.label}</td>
-              <td>{line.value}</td>
-            </tr>
-          );
-        })}
-      </tbody>
-    </table>
+    <>
+      <table className="table summary-table-header">
```
leem42: Why did the table get split into two?

PR #226, opened 2019-09-24
Portals 560

conundrumgirl:


src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -6,53 +6,62 @@ import './DrugUploadTool.scss'
 
 import * as React from 'react'
 
-import { Entity, } from '../../utils/jsonResponses/Entity'
+import { Entity } from '../../utils/jsonResponses/Entity'
```
conundrumgirl: This is 90% formatting. The only change should be 'unauthenticated' message

src/lib/containers/drug_upload_tool/DrugUploadTool.tsx
```diff
@@ -337,11 +337,24 @@ class DrugUploadTool extends React.Component<
     }
   };
 
+  renderUnauthenticatedView = (token: string | undefined) => {
+    if (token) {
+      return <></>
+    } else {
+      return (
```
leem42: I believe that this should change slightly from renderUnauthenticatedView -> renderUnauthorizedView, and the check would be if (state.status.UNAUTHORIZED) {} because the portal will block the user from reaching this component until they signed in and have a token

PR #228, opened 2019-09-24
PORTALS-672

jay-hodgson:
Add form service calls, and request/response object types.


src/demo/containers/App.tsx
```diff
@@ -14,8 +14,10 @@ import { SynapseClient } from '../../lib/utils/';
  */
 
 type AppState = {
-  token: string;
-};
+  token: string
+}
+export const TokenContext = React.createContext('')
```
jay-hodgson: added token context to App

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
```diff
@@ -0,0 +1,173 @@
+import * as React from 'react'
```
jay-hodgson: Added new demo area (note that you must change the endpoints to staging this week to see this in action).

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public componentDidUpdate() {
+    this.listMyFormData()
+    this.listFormAdminFormData()
+  }
+
+  render() {
+    return (
+      <TokenContext.Consumer>
```
leem42: I like using context for the token value! in the future we could even consider wrapping all our components in a consumer since its a universal prop

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
```diff
@@ -0,0 +1,173 @@
+import * as React from 'react'
+import { TokenContext } from '../App'
+import { FormGroup, FormData, ListRequest, ListResponse, StatusEnum } from 'lib/utils/jsonResponses/Forms'
+import { SynapseClient } from 'lib'
+import { FileUploadComplete } from 'lib/utils/jsonResponses/FileUploadComplete'
+
+type FormServicesIntegrationDemoState = {
+  token?: string
```
leem42: [here's a link](https://reactjs.org/docs/context.html#classcontexttype) to the react docs showing how to use the context value from the class itself, e.g. `this.context`, so you don't have to copy the context value over to state

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
```diff
@@ -0,0 +1,173 @@
...
+    this.onCreateFormGroup = this.onCreateFormGroup.bind(this)
+    this.onCreateFormData = this.onCreateFormData.bind(this)
+    this.onUpdateFormData = this.onUpdateFormData.bind(this)
+    this.onSubmitFormData = this.onSubmitFormData.bind(this)
+    this.listMyFormData = this.listMyFormData.bind(this)
+    this.listFormAdminFormData = this.listFormAdminFormData.bind(this)
+  }
+
+  public onCreateFormGroup(): void {
+    SynapseClient.createFormGroup('test form group 3', this.state.token!)
```
leem42: same as above with reading the token, should be able to read it directly

src/lib/utils/jsonResponses/Forms.ts
```diff
@@ -0,0 +1,50 @@
+//https://docs.synapse.org/rest/org/sagebionetworks/repo/model/form/FormGroup.html
+export type FormGroup = {
+  groupId?: string
```
leem42: is name the only guaranteed field for FormGroup?

src/lib/utils/jsonResponses/Forms.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// see https://docs.synapse.org/rest/org/sagebionetworks/repo/model/form/ListRequest.html
+export type ListRequest = {
+  filterByState?: StatusEnum[],
+  groupId: string,
+  nextPageToken?: string
+}
+
+// see https://docs.synapse.org/rest/org/sagebionetworks/repo/model/form/ListResponse.html
+export type ListResponse = {
```
leem42: this could be made generic and then typed in the actual usage since thats how the API is constructed

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
```diff
@@ -0,0 +1,173 @@
...
+    this.listFormAdminFormData()
+  }
+
+  render() {
+    return (
+      <TokenContext.Consumer>
+        {
+          (token: string) => {
+            if (token !== this.state.token) {
+              this.setState({
```
leem42: noooooo

src/lib/utils/jsonResponses/Forms.ts
jay-hodgson: That's a good point.  In other service calls, objects are used for both request and response objects (where fields like "id" would not be filled in for the request), but that's not applicable for this.  I'll mark other fields as required...

src/lib/utils/jsonResponses/Forms.ts
jay-hodgson: could you elaborate on this point?

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
jay-hodgson: could you elaborate on this point?

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
jay-hodgson: I'll check it out!

jay-hodgson: Tried setting
FormServicesIntegrationDemo.contextType = TokenContext
but this.context was not updated with the token (was always just {} after login or session init)

leem42: Okay, we can leave this as a TODO for now and come back to it later

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
leem42: Ideally setState wouldn't get called in render and instead in componentDidMount since setState is async and it kind of goes against the idea of render being pure.

src/lib/utils/jsonResponses/Forms.ts
leem42: Sorry for not being clear, I meant that this could be re-written as 

```
export type ListResponse<T> = {
   page: T [],
   nextPageToken?: string
}
```
and then on this [line](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/228/files#diff-870197c271aeaea05bb0cef7bc9a65e7R11) and the next you could write them as 

```
 myFormDataList<FormData>
 formAdminFormDataList<FormData>
```

jay-hodgson: But ListResponse is org.sagebionetworks.repo.model.**form**.ListResponse, where the page data object type must be FormData.  
If we made it configurable, the developer who uses this method (Alina in this case) would need to set the response type, and what if they specify a type other than FormData?  It would compile, but fail at runtime (because the response object would not be the expected type).

src/demo/containers/playground/FormServicesIntegrationDemo.tsx
jay-hodgson: This makes sense, thanks Michael!

src/lib/utils/jsonResponses/Forms.ts
leem42: whoops! I was wrong, earlier I though I was looking at a docs page that said page<Entity> with a type param, but clearly not... please ignore my comment.

PR #236, opened 2019-10-02
SWC-4926

jay-hodgson:


src/lib/containers/StatisticsPlot.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+class StatisticsPlot extends React.Component<StatisticsPlotProps, StatisticsPlotState> {
+
+  constructor(props: StatisticsPlotProps) {
+    super(props)
+    this.state = {
+      isLoaded: false,
+      plotData: {} as ProjectFilesStatisticsResponse
+    }
+    this.fetchPlotlyData = this.fetchPlotlyData.bind(this)
```
conundrumgirl: I think you can avoid having to do bind those functions by having them as class props e.f.
`fetchPlotlyData = () =>  {
...` instead of `  public fetchPlotlyData() { ...`

src/__tests__/lib/containers/StatisticsPlot.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    SynapseClient.getProjectStatistics = jest.fn(() =>
+      Promise.resolve(projectFilesStatsResponse),
+    )
+  })
+
+  it('displays plot', async () => {
+    const { wrapper, instance } = await createShallowComponent(props)
+    await instance.componentDidMount()
+    expect(wrapper).toBeDefined()
+    expect(wrapper.find(StatisticsPlot)).toBeDefined()
```
conundrumgirl: does the fact that it's defined guarantee that it displays a pot?
 I think that since [fetchPlotlyData()]( https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/236/files#diff-0a1def2335aa3a38f950b484938b2f36R48) does not return a promise componentDidMount() can return before it gets resolved.

PR #240, opened 2019-10-04
Updates

conundrumgirl:
I think this might update the current gh-pages site: https://leem42.github.io/Synapse-React-Client/


jay-hodgson:
Note that we'll need to merge master into gh-pages to fix the login issue.

PR #241, opened 2019-10-04
github pages support. if outside synapse.org subdomain, then store in…

jay-hodgson:
… local domain only cookie


src/lib/utils/SynapseClient.ts
```diff
@@ -1278,3 +1278,28 @@ export const getProjectStatistics = (
   endpoint: string = DEFAULT_ENDPOINT) : Promise<ProjectFilesStatisticsResponse>=> {
     return doPost(`/repo/v1/statistics`, request, sessionToken, undefined, endpoint)
 }
+
+const setCookie = (name: string, value: string | undefined, days: number) => {
+  if (value) {
```
conundrumgirl: I would early out here `if(!value) {return}`

PR #242, opened 2019-10-06
Switch to using standard spark-md5 (195k DL/week) library (with types) instead of browser-md5-file (295 DL/week)

jay-hodgson:
Also rebuild umd build


jay-hodgson:
> looks good to me. Unrelated question: do we need to keep the .css files in git? Those are generated, aren't they?

I don't know why all of the build output has been checked in, is that to make github pages work?

PR #243, opened 2019-10-07
Updates to drug nomination tool

conundrumgirl:
1. Enhanced error reporting.  If configuration JSON fails to load there is a notification displaying the error
2. On a submission grid made the submitted items clickable and removed '.json' extension from file names

image
3. Added a 'submitted' treatment to the form disabling actions and making it obvious to the user that form has been submitted
image

4 updated configuration files (just for version control)
5 added paging


src/lib/containers/drug_upload_tool/Header.tsx
```diff
@@ -1,18 +1,26 @@
...
-  bodyText?: string; //should never take user input only use cardcoded text from a parent component. Uses dangerouslySetInnerHTML
-  title: string;
-};
+  isSubmitted?: boolean
+  bodyText?: string //should never take user input only use cardcoded text from a parent component. Uses dangerouslySetInnerHTML
+  title: string
+}
 
 export default function Header(props: HeaderProps) {
+  const bodyText = props.isSubmitted ? (
```
jay-hodgson: IMO, the using the ternary operator twice inline makes this more challenging to parse.  If you agree, please rewrite to ease mental burden.

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -69,6 +67,14 @@ export default class UserFileGrid extends React.Component<
       <p>Are you sure you want to trash this submission?' </p>
     </>
   )
+  static requestFilter = {
```
jay-hodgson: If this is unchanging, could you make this a const (not just a single instance associated to the class)

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
```diff
@@ -159,6 +150,47 @@ export default class UserFileGrid extends React.Component<
...
+        inProgress: {
+          fileList: [...prevState.inProgress.fileList, ...result.fileList],
+          nextPageToken: result.nextPageToken,
+        },
+      }))
+    }
+  }
+
+  getUserFileListing = async (): Promise<void> => {
+    const inProgress = await this.getTypeFileListing(
```
jay-hodgson: These two calls could be changed so that they are run in parallel (no dependency on one another) to reduce the time it takes to see submitted forms.

src/lib/containers/drug_upload_tool/Header.tsx
conundrumgirl: Agree

src/lib/containers/drug_upload_tool/UserFileGrid.tsx
conundrumgirl: TS yells at me that it can't be static and constant but I did make it read only

PR #250, opened 2019-10-10
on deploy, publish to the gh-pages upstream branch as well

jay-hodgson:


jay-hodgson:
@conundrumgirl and I decided that we would rather the upstream gh-pages show a bleeding edge version of SRC than a stale one, so deploy now updates organization level gh-pages (found here: https://sage-bionetworks.github.io/Synapse-React-Client/)
(FYI @leem42)

PR #253, opened 2019-10-15
PORTALS-858: introduce mechanism to lead to cleaner queries (since an…

jay-hodgson:
…notation keys cannot have spaces) and improved functionality (since aliased columns lose the native column model type info)


leem42:
I consider this a high risk change because it impacts many different pieces. I believe the idea is sound and would simplify things overall, but am concerned that there's not a lot of time to test this out. 

That being said, I believe there are a few other places where the unCamelCase function needs to be used. In Search on lines [87](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/cf3a24d7193b1ed528cc211d3474656ec55635a0/src/lib/containers/Search.tsx#L87), [205](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/cf3a24d7193b1ed528cc211d3474656ec55635a0/src/lib/containers/Search.tsx#L205), and [229](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/cf3a24d7193b1ed528cc211d3474656ec55635a0/src/lib/containers/Search.tsx#L229) . And then also in GenericCard on lines [185](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/master/src/lib/containers/GenericCard.tsx#L185), [217](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/master/src/lib/containers/GenericCard.tsx#L217), [218](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/master/src/lib/containers/GenericCard.tsx#L218), [219](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/master/src/lib/containers/GenericCard.tsx#L219)

jay-hodgson:
We certainly don't want to introduce unnecessary risk to the release, so we can hold off on this PR.  But I did make the other suggested code changes (in Search and GenericCard).

PR #254, opened 2019-10-16
Portals 560

conundrumgirl:
Last feature requests and updates to Drug Upload Tool


src/lib/containers/drug_upload_tool/DrugUploadTool.tsx
```diff
@@ -78,16 +79,20 @@ class DrugUploadTool extends React.Component<
   getFileEntityData = async (
     token: string,
     entityId: string,
+    {doDereference = false} = {}
```
leem42: I believe this signature could be simplified by making the doReference boolean its own argument instead of it being a key in a dictionary.

conundrumgirl: Yes,  it's bad code agree. I will create  another fn for the dereferencing case

PR #263, opened 2019-10-23
PORTALS-660

leem42:
PORTALS-660 Updates to search component
PORTALS-591 Changed endpoint logic to look for key in window with overrides, otherwise points to production
Got rid of css files!


README.md
```diff
@@ -84,6 +84,19 @@ Note there are a number of CDNs required to finish this functionality-
   <script src="https://unpkg.com/markdown-it-br@1.0.0/dist/markdown-it-br.min.js"></script>
 ```
 
+## Setting Endpoints
+
+### Configuring endpoint destinations for repo and portal
+Specifying the following in the window object will override request endpoints:
+
+```js
+window.SRC.OVERIDE_ENDPOINT_CONFIG = {
```
jay-hodgson: (sp) Override

src/lib/utils/getEndpoint.ts
```diff
@@ -0,0 +1,33 @@
+export type BackendDestination = 'PORTAL_ENDPOINT' | 'SWC_ENDPOINT'
+// These variables indicate which endpoint should be called
+export const PORTAL_ENDPOINT: BackendDestination = 'PORTAL_ENDPOINT'
```
jay-hodgson: why do you have PORTAL_ENDPOINT and SWC_ENDPOINT, isn't that the same thing?  Expected PORTAL (or SWC) and REPO

src/lib/containers/Login.tsx
```diff
@@ -47,104 +45,111 @@ class Login extends React.Component<Props, State> {
...
-              })
-            })
-            .catch((err: any) => {
-              console.log('Handle login failed with err = ', err)
-              this.setState({
-                errorMessage: err.reason,
-                hasLoginInFailed: true,
-                isSignedIn: false
-              })
+    SynapseClient.login(this.state.username, this.state.password)
```
conundrumgirl: Just a general thought, I wonder if it makes sense to start refactoring  nested promises into  async/awaits when we are in those areas doing other changes.

src/lib/utils/getEndpoint.ts
```diff
@@ -0,0 +1,33 @@
...
+  PORTAL: string
+}
+
+export const PRODUCTION_ENDPOINT_CONFIG: EndpointObject = {
+  REPO: 'https://repo-prod.prod.sagebase.org/',
+  PORTAL: 'https://www.synapse.org/',
+}
+
+// Given an endpoint will return the specific stack object
+export const getEndpoint = (endpoint: BackendDestination) => {
```
conundrumgirl: please add return type

src/lib/utils/getEndpoint.ts
```diff
@@ -0,0 +1,33 @@
+export type BackendDestination = 'PORTAL_ENDPOINT' | 'SWC_ENDPOINT'
+// These variables indicate which endpoint should be called
+export const REPO_ENDPOINT: BackendDestination = 'PORTAL_ENDPOINT'
```
conundrumgirl: naming is a little confusing REPO_ENDPOINT = 'PORTAL_ENDPOINT' and PORTAL_ENDPOINT: BackendDestination = 'SWC_ENDPOINT' ?

src/lib/utils/getEndpoint.ts
```diff
@@ -0,0 +1,33 @@
...
+export const PORTAL_ENDPOINT: BackendDestination = 'SWC_ENDPOINT'
+
+// Encodes the portal and repo service destinations
+type EndpointObject = {
+  REPO: string
+  PORTAL: string
+}
+
+export const PRODUCTION_ENDPOINT_CONFIG: EndpointObject = {
```
conundrumgirl: I wonder if these objects can just be distilled to an ENUM keyed off the BackendDestination  (may be the answer is 'no':)

src/lib/containers/Login.tsx
leem42: I like that idea too! I can change this over to use async/await

src/lib/utils/getEndpoint.ts
leem42: whoops good catch, I will fix that..

PR #275, opened 2019-11-04
PORTALS-880

leem42:
Updated getUserProjectList function in SynapseClient to use new API, made an enum for the filter types


src/lib/utils/SynapseClient.ts
```diff
@@ -832,14 +836,14 @@ export const getUserFavorites = (sessionToken: string | undefined) => {
   ) as Promise<any>
 }
 /**
- *  http://docs.synapse.org/rest/GET/projects/type.html
- *  @param {String} projectDetails Can be "MY_PROJECTS", "MY_CREATED_PROJECTS" or "MY_PARTICIPATED_PROJECTS"
+ *  https://docs.synapse.org/rest/GET/projects.html
+ *  @param {String} projectDetails Can be "CREATED", "PARTICIPATED" or "TEAM"
  */
 export const getUserProjectList = (
```
jay-hodgson: good change, but recommend we just remove these calls since they are not being used today (service does not support offset anymore, for example, so this version certainly doesn't support pagination).

PR #277, opened 2019-11-05
Feat/miscl

leem42:
* There was a misnamed prop that had to be fixed. This was because two components had copies of the same set of props, I fixed this by importing the common props of one into the other so that won't happen again.
* Bump version


src/lib/containers/CardContainerLogic.tsx
```diff
@@ -51,12 +54,10 @@ export type CardContainerLogicProps = {
   sqlOperator?: SQLOperator
   searchParams?: KeyValue
   facet?: string
-  loadingScreen?: JSX.Element
-  genericCardSchema?: GenericCardSchema
   backgroundColor?: string
   isHeader?: boolean
   sql: string
-} & CommonCardProps
+} & CardConfiguration
```
conundrumgirl: Intersection types - fancy!

PR #279, opened 2019-11-06
PORTALS-871

leem42:
Fixes markdown component by:

1. Using DOMParser.parseFromString function to render the markdown html into an HTMLElement
2. Recursing on the resulting HTMLElement and rendering a 'react' version of it while replacing widgets with the correct components


src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -379,25 +364,88 @@ export default class MarkdownSynapse extends React.Component<
...
 
+  /**
+   * recursiveRender will render react tree from HTML tree
+   *
+   * @param {Node} element This will be either a text Node or an HTMLElement
+   * @param {string} markdown The original markdown, its kept as a special case for the table of contents widget
+   * @returns {*}
+   * @memberof MarkdownSynapse
+   */
+  public recursiveRender(element: Node, markdown: string): any {
```
jay-hodgson: ⭐️

PR #282, opened 2019-11-06
PORTALS-882

jay-hodgson:


src/lib/utils/DownloadSpeedTest.ts
```diff
@@ -0,0 +1,75 @@
...
+    getEntity(sessionToken, entityId).then((entity: Entity) => {
+      const fileEntity:FileEntity = entity as FileEntity
+      const fileHandleAssociationList = [
+        {
+          associateObjectId: fileEntity.id,
+          associateObjectType: 'FileEntity',
+          fileHandleId: fileEntity.dataFileHandleId,
+        },
+      ]
+      const request: any = {
```
leem42: Could use BatchFileRequest instead of any for the request

src/lib/utils/DownloadSpeedTest.ts
```diff
@@ -0,0 +1,75 @@
...
+      getFiles(request, sessionToken)
+        .then((data: BatchFileResult) => {
+          const presignedUrl: string = data.requestedFiles[0].preSignedURL
+          const fileHandle: FileHandle = data.requestedFiles[0].fileHandle
+          // start test!
+          const startMs = new Date().getTime()
+          return getFileHandleContent(fileHandle, presignedUrl).then(
+            () => {
+              // bytes/second
+              console.log(`Transferred ${fileHandle.contentSize/1000000} MB in ${((new Date().getTime() - startMs - ESTIMATED_CORS_TIME_MS)/1000)} seconds`)
```
leem42: Could remove this console.log

src/lib/utils/DownloadSpeedTest.ts
jay-hodgson: Done.

src/lib/utils/DownloadSpeedTest.ts
jay-hodgson: Useful calculation, especially if you don't trust the reported rate.  So I just commented out.

PR #289, opened 2019-11-20
added download details component

leem42:


src/lib/containers/download_list/DownloadDetails.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export default function DownloadDetails(props: DownloadDetailsProps) {
+  const [state, setState] = useState<State>({
+    isLoading: true,
+    downloadSpeed: 0,
+  })
+  const { isLoading, downloadSpeed } = state
+  const { token, numFiles, numBytes } = props
+
+  useEffect(() => {
```
conundrumgirl: If the intention to only run it on mount? 
If so,  may be this would be more explicit?
```useEffect(() => {
{
      if (token) {
        testDownloadSpeed(token).then(speed => {
          setState({
            isLoading: false,
            downloadSpeed: speed,
          })
        })
    }, []);
```
Or is this because there is a possibility that token will come through later?

src/lib/containers/download_list/DownloadDetails.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { useState, useEffect } from 'react'
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { faFile, faDatabase, faClock } from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import calculateFriendlyFileSize from '../calculateFriendlyFileSize'
```
conundrumgirl: Haha I did exactly the same thing. Shouldn't it live in _utils_ though?

src/lib/containers/download_list/DownloadDetails.tsx
leem42: whoops! on mount was my intent, that would be better.

src/lib/containers/download_list/DownloadDetails.tsx
leem42: thanks for pointing that out, yep calculateFriendlyFileSize should be in utils!

PR #290, opened 2019-11-22
PORTALS-881

leem42:
- Added download list table component
- Added create package component
- Pieced out functions  /utils/UtilityFns -> utils/functions/{function name}
- Changed rollup configuration to load SRC after the body has loaded onto the screen because of react-bootstrap umd build need access to dom on load


src/lib/containers/download_list/CreatePackage.tsx
```diff
@@ -0,0 +1,187 @@
...
+  updateDownloadList: Function
+}
+
+type Alert = {
+  message: string
+  variant: ReactBoostrap.AlertProps['variant']
+  className: string | undefined
+}
+
+export const templateDownload =
```
conundrumgirl: do we have a convention to CAPITALIZE consts?

src/lib/containers/download_list/CreatePackage.tsx
```diff
@@ -0,0 +1,187 @@
...
+    className: undefined,
+    variant: undefined,
+  })
+  const [bulkFileDownloadResponse, setBulkFileDownloadResponse] = useState<
+    BulkFileDownloadResponse | undefined
+  >(undefined)
+  const { token, children, updateDownloadList } = props
+
+  const createPackageHandler = async (event: React.SyntheticEvent) => {
+    event && event.preventDefault()
```
conundrumgirl: isn't event always defined ?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+    batchFileResult: undefined,
+    downloadList: undefined,
+  })
+  // https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables
+  let [isLoading, setIsLoading] = useState<LoadingState>(true)
+  const { token } = props
+  const { references, batchFileResult, downloadList } = data
+  const requestedFiles =
+    (batchFileResult && batchFileResult.requestedFiles) || []
+  const ownerIds: any = requestedFiles
```
conundrumgirl: what is the reason for 'any' isn't this string[]?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+    .map(el => el.fileHandle && el.fileHandle.createdBy)
+    .filter(el => el)
```
conundrumgirl: does this filter out falsey values?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+      // which has additional metadata - createdBy, numBytes, etc.
+      const batchFileResult = await getFiles(batchFileRequest, token)
+      setData({
+        references,
+        batchFileResult,
+        downloadList,
+      })
+      setIsLoading(false)
+    } catch (e) {
+      console.error('Error in DownloadList API call : ', e)
```
conundrumgirl: do we need a `finally` here to turn off loading? Also may be we should notify the user on the failure?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }
+  }
+
+  const clearDownloadList = (
+    _event: React.SyntheticEvent<HTMLButtonElement>,
+  ) => {
+    setIsLoading(true)
+    deleteDownloadList(token)
+      .then(() => {
```
conundrumgirl: any particular reason why not async/await here? Also stop loading on error?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+  ) => {
+    const list: FileHandleAssociation[] = [
+      {
+        fileHandleId,
+        associateObjectId,
+        associateObjectType: FileHandleAssociateType.FileEntity,
+      },
+    ]
+    setIsLoading(true)
+    deleteDownloadListFiles(list, token)
```
conundrumgirl: same quesiton about sync/await and also do we need  `setIsLoading(false)` somewhere?

src/lib/utils/SynapseClient.ts
```diff
@@ -1806,3 +1843,53 @@ export const getRestrictionInformation = (
...
+  return doPost(
+    '/file/v1/download/list/remove',
+    { list },
+    sessionToken,
+    undefined,
+    BackendDestinationEnum.REPO_ENDPOINT,
+  )
+}
+
+// https://rest-docs.synapse.org/rest/GET/download/list.html
```
conundrumgirl: `https://rest-docs.synapse.org/rest/DELETE/download/list.html` ?

src/lib/utils/functions/testDownloadSpeed.ts
```diff
@@ -0,0 +1,88 @@
+import { getEntity, getFileHandleContent, getFiles } from '../SynapseClient'
```
conundrumgirl: That's preexisting code, just renamed/moved?

src/lib/utils/jsonResponses/BulkFileDownloadRequest.ts
```diff
@@ -0,0 +1,16 @@
+import { FileHandleAssociation } from './FileHandleAssociation'
+
+export enum ZipFileFormat {
+  CommandLineCache = 'CommandLineCache', // This is the default format. Zip file format structure is: (%1000)//
```
conundrumgirl: Should keys be capitalized?

src/__tests__/lib/containers/download_list/CreatePackage.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+  })
+  it('Fails to creates a package without a filename', async () => {
+    const error = {
+      reason: 'Failed to download any files',
+    }
+    SynapseClient.getDownloadOrder = jest.fn().mockRejectedValue(error)
+    const wrapper = shallow(<CreatePackage {...props} />)
+    expect(wrapper).toBeDefined()
+    expect(wrapper.find(Alert)).toHaveLength(0)
+    await act(async () => {
```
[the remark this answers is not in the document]
leem42: fixed to check that alert component is present

src/lib/containers/download_list/CreatePackage.tsx
leem42: fixed

src/lib/containers/download_list/CreatePackage.tsx
leem42: In the testing environment its not, but perhaps there's a library that would help with that. For now I changed the tests to pass in a stub for preventDefault when triggering this method, which I think is preferable because then we don't have odd statements like the above, thanks for pointing this out.

src/lib/containers/download_list/DownloadListTable.tsx
leem42: fixed

src/lib/containers/download_list/DownloadListTable.tsx
leem42: Yes, but I realize now thats a bit clunky, I changed it to instead filter out the items that have createdBy set and then map them out to just that value

src/lib/utils/SynapseClient.ts
leem42: fixed

src/lib/containers/download_list/DownloadListTable.tsx
leem42: changed to use try/catch/finally with loading state set to false in finally block, same as below.

src/lib/utils/jsonResponses/BulkFileDownloadRequest.ts
leem42: I'm open to change on this, but my thinking here was to keep consistency with the enum defined by the backend

PR #299, opened 2019-11-27
chore(SRC): remove some circular deps, move shared consts

conundrumgirl:
1. created `lib/utils/UtilityFns` and moved `unCamelCase()` and `calculateFriendlyFileSize()` there
2. created `lib/containers/table/SynapseTableConstants` and moved shared constants from `SynapseTable` there
3. moved css imports into DownloadDetails and DownloadConfirmation instead of index


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -1,3 +1,4 @@
+import {  } from './SynapseTableConstants';
```
leem42: could remove

src/lib/containers/table/SynapseTableConstants.ts
```diff
@@ -0,0 +1,6 @@
+export const TOOLTIP_DELAY_SHOW = 500
```
leem42: awesome!

PR #302, opened 2019-12-03
PORTALS-924, and PORTALS-925

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -471,16 +471,20 @@ export default class SynapseTable extends React.Component<
         Next
       </button>
     )
-
+    let isShowingAccessColumn: boolean | undefined = showAccessColumn
+    if ( showAccessColumn && rows.length > 0 ) {
```
leem42: I think this could be refactored to: `if (isShowingAccessColumn && isGroupByInSql(....) && rows.length > 0 )`
because isShowingAccessColumn is equivalent to showAccessColumn and there also needs be a check for a group by in the sql since QueryWrapperMenu doesn't have the granularity to specify per facet if showAccessColumn is true and the amp-ad explore/data tab has a group by in the individuals facet query

PR #311, opened 2019-12-06
fix(SRC style): PORTALS-884 Proper spacing above the table

conundrumgirl:
https://sagebionetworks.jira.com/browse/PORTALS-884

before:
image

after:
image

### DIFF
--- a/src/lib/containers/QueryWrapper.tsx
+++ b/src/lib/containers/QueryWrapper.tsx
@@ -340,11 +340,6 @@ export default class QueryWrapper extends React.Component<
     )
 
     const loadingCusrorClass = isLoading ? 'SRC-logo-cursor' : ''
-    if (this.props.showMenu) {
-      // menu is to the left of the child components so we let that add its
-      // own html
-      return <div className={`${loadingCusrorClass}`}>{childrenWithProps}</div>
-    }
-    return <div className={`${loadingCusrorClass}`}>{childrenWithProps}</div>
+    return <div className={`SRC-wrapper ${loadingCusrorClass}`}>{childrenWithProps}</div>
   }
 }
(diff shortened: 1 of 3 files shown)


src/lib/containers/QueryWrapper.tsx
```diff
@@ -339,7 +339,7 @@ export default class QueryWrapper extends React.Component<
       },
     )
 
-    const loadingCusrorClass = isLoading ? 'SRC-logo-cursor' : ''
+    const loadingCusrorClass = isLoading ? 'SRC-logo-cursor' : 'SRC-wrapper'
```
leem42: I think the class 'SRC-wrapper' can be fixed to the top move div, it doesn't need to be conditional.

src/lib/style/Core.scss
```diff
@@ -967,3 +971,10 @@ a.dropdown-item {
 .plotly .slice {
   cursor: pointer;
 }
+
+hr ~ .SRC-wrapper {
```
leem42: I think this scss should be put in the GenerateComponentsFromRow's [scss](https://github.com/Sage-Bionetworks/portals/blob/master/app-template/src/portal-components/GenerateComponentsFromRow.scss) instead because the hr is only added by that component

conundrumgirl: So i hesitated with this one.
Part of me agrees with the above statement but on the other hand it doesn't know about the
.SRC-wrapper   .SRC-marginTopMinusTen  components and the margin gets applied to those components and not to hr

src/lib/containers/QueryWrapper.tsx
conundrumgirl: isn't this the topmost div?

PR #312, opened 2019-12-06
PORTALS-881 Updates

leem42:
HasAccess component:
* changed colors
* Use's file handle concreteType to determine if external link
* Accepts prop isRestricted to quickly tell component no access
* More special cases for when download type is cloud/external/too large

DownloadListTable component:
* Gets entity name instead of file name
* Passes down file handle for downloadable files to HasAccess
* Filters list of batch entity header calls to only those which are non downloadable files
from user download list
* Delete file from list now uses the returned download list to remove extraneous calls

Synapse Client:
* Simplify fetch to combine calls
* change statusCode -> status
* Added more types - Response, Response['credentials']

DownloadDetails test:
* Change test to use act() function call
* Changed component to add additional case for inactive state


src/lib/containers/HasAccess.tsx
```diff
@@ -55,65 +94,109 @@ export default class HasAccess extends React.Component<
     this.getRestrictionInformation()
   }
 
-  getRestrictionInformation() {
-    const { synapseId, token } = this.props
-    if (this.state.restrictionInformation || !synapseId || !token) {
+  getRestrictionInformation = () => {
+    const { synapseId, token, deniedAccess = false } = this.props
```
jay-hodgson: deniedAccess => forceIsRestricted ?

src/lib/containers/HasAccess.tsx
```diff
@@ -55,65 +94,109 @@ export default class HasAccess extends React.Component<
...
+
+  renderIconHelper = (iconProp: IconProp, classColor: string) => {
+    return (
+      <span className="fa-layers fa-fw" style={{ marginRight: 5 }}>
+        <FontAwesomeIcon icon={faCircle} className={classColor} size="lg" />
+        <FontAwesomeIcon
+          icon={iconProp}
+          className="SRC-half-opacity"
+          style={{ transform: 'translate(4%, -4%)' }}
+          size="xs"
```
jay-hodgson: bigger?

src/lib/containers/HasAccess.tsx
```diff
@@ -55,65 +94,109 @@ export default class HasAccess extends React.Component<
...
+        restrictionInformation!.hasUnmetAccessRequirement) ||
+      ''
+    const restrictionLevel =
+      (restrictionInformation && restrictionInformation!.restrictionLevel) || ''
     let viewARsLink: React.ReactElement = <></>
     if (
-      restrictionInformation &&
-      RestrictionLevel.OPEN !== restrictionInformation.restrictionLevel
+      RestrictionLevel.OPEN !== restrictionLevel &&
+      downloadType !== DownloadTypeEnum.TooLargeFile
```
jay-hodgson: Alternatively, you could still show the "Get Access"/"View Terms" link when file is too large.

src/lib/containers/HasAccess.tsx
```diff
@@ -126,9 +209,37 @@ export default class HasAccess extends React.Component<
...
+    let viewARsLink: React.ReactElement = this.renderARsLink(downloadType)
     return (
       <span style={{ whiteSpace: 'nowrap' }}>
-        {icon} {viewARsLink}
+        {tooltipText && (
+          <>
+            <span tabIndex={0} data-for={synapseId} data-tip={tooltipText}>
+              {icon}
+            </span>
+            <ReactTooltip
```
jay-hodgson: Is this still the same syntax (odd ReactTooltip in the jsx)

jay-hodgson: and max width

PR #315, opened 2019-12-12
PORTALS-941

leem42:
Simplified props for cards

There are now only two types:
```ts
type CardLink {
  baseURL: string 
  // the key that will go into the url
  URLColumnName: string
  // the column name who's value will be used for the markdown
  matchColumnName: string
  isMarkdown: false
}
```

```ts
type MarkdownLink {
  // the column name who's value will be used
  matchColumnName: string
  isMarkdown: false
}
```

Also removed tests which only made sense in the prior version


src/lib/containers/GenericCard.tsx
```diff
@@ -91,16 +90,15 @@ export default class GenericCard extends React.Component<
...
-      // tested this link on the browser, there's no need to encode the URL, the browser picks up on that automatically
-      linkDisplay = `#/${titleLink.baseURL}?${urlParams}`
+      const { matchColumnName, URLColumnName } = titleLink
+      const indexInData = schema[matchColumnName]
+      if (indexInData === undefined) {
+        console.error(
+          `Could not find match for data: ${data} with columnName ${matchColumnName}`,
+        )
+      }
+      const value = data[indexInData]
```
jay-hodgson: else (if indexInData is defined)

PR #316, opened 2019-12-13
fix login failure that occurs on staging (plain text response issue)

jay-hodgson:


src/lib/utils/SynapseClient.ts
```diff
@@ -112,14 +112,20 @@ const fetchWithExponentialTimeout = <T>(
...
+          try {
+            const json = JSON.parse(text)
+            return resp.ok ? Promise.resolve(json) : Promise.reject<T>(json)
+          } catch (error) {
+            // failed to parse json, return text
+            return resp.ok ? Promise.resolve(text) : Promise.reject<T>(text)
+          }
         })
         .catch((error: SynapseError) => {
           if (resp.ok) {
```
leem42: this block (if lines 127-130) can be removed because in the then above it will always handle when the response is okay

PR #319, opened 2019-12-13
feat(SRC-Table-Filter):  refactor of FacetColumnResult

conundrumgirl:
This PR addresses 3 issues with no functional changes
1. [QueryResultBundle.tsx ](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/319/files#diff-2a93e7bb3a9a44cb1b41ac44fdf7bcc7R8)  is changed to be a generic type which takes FacetColumnResultValues or FacetColumnResultRange
2. Vars.scss is added to make better use of variables
3 Added another mask to [QueryWrapperMenu](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/319/files#diff-5acad44bd51e5983e5e5700658d37ef2R380) to get column types


src/lib/utils/jsonResponses/Table/QueryResultBundle.tsx
```diff
@@ -5,14 +5,14 @@ import { SelectColumn } from './SelectColumn'
...
-export interface QueryResultBundle {
+export interface QueryResultBundle <T extends FacetColumnResult> {
   concreteType: 'org.sagebionetworks.repo.model.table.QueryResultBundle'
   queryResult: QueryResult // A page of query result.
   queryCount?: number // The total number of rows that match the query. Use mask = 0x2 to include in the bundle.
   selectColumns?: SelectColumn[] // The list of SelectColumns from the select clause. Use mask = 0x4 to include in the bundle.
   maxRowsPerPage?: number // The maximum number of rows that can be retrieved in a single call. This is a function of the columns that are selected in the query. Use mask = 0x8 to include in the bundle.
   columnModels?: ColumnModel[] // The list of ColumnModels for the table. Use mask = 0x10 to include in the bundle.
-  facets?: FacetColumnResult[] // The list of facets for the search results
+  facets?: T[] // The list of facets for the search results
```
leem42: So I have two qualms with this:

1. The backend does not have a type parameter for QueryResultBundle and I'd like to keep the frontend and backend types as close as possible to each other

2. With the new type parameter added to the facets field it suggests that the facets array will be a uniform type of exclusively FacetColumnResultValues or FacetColumnResultRange, but the array really could be a mix of the two, so I think the interface should remain. I do see elsewhere in this PR the union of FacetColumnResultValues|FacetColumnResultRange is passed in as the type parameter for the facets array but I think that union is essentially functioning as an interface because it saying that the facet-values could be one or other.

Could you expand on the argument for having the type parameter?

conundrumgirl: Here is what I am trying to correct: The current version of QueryResultBundle.tsx has facets as FacetColumnResult[] which is  defined in the code  as 
```concreteType: string
  columnName: string // The name of the faceted column
  facetType: 'enumeration' | 'range'
  facetValues: any // The list of QueryFace
```
This actually is incorrect.
From the documentation
```
 <INTERFACE>
org.sagebionetworks.repo.model.table.FacetColumnResult
Resulting information about a faceted column
Known Implementations:
FacetColumnResultValues,  FacetColumnResultRang

Field | Type | Description
-- | -- | --
concreteType | STRING |  
columnName | STRING | The name of the faceted column
facetType | FacetType

```
https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/319/files#diff-c59696e6b1deecde0e5c75d51a1972f0L27

https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/FacetColumnResult.html

There are no facetValues there. They only exist in the https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/FacetColumnResultValues.html they don't exists for the Range implementation

So the current implementation is actually not correct.

As far as the second concern, my understanding is that yes, it is the union type  but the array of a union type means that it's an array of items that belong to either of those types.
https://stackoverflow.com/questions/29794014/typing-an-array-with-a-union-type-in-typescript

conundrumgirl:
I have replaced the union type with the FacetColumnResult which they both extend. I think it's way cleaner now:) Thanks for bringing up the ugliness

PR #321, opened 2019-12-16
PORTALS-927

leem42:
### Part 1: PORTALS-927

Renamed folder jsonResponses -> synapseTypes
Put all types in barrel file from synapseTypes folder, coalesced all imports that imported from synapseTypes in more than one place
Fixed file extensions for all the types: .tsx -> .ts
Removed unused mock file
Made entityId required type across the project
Turned off semi-colon warning for when its added to unpacking objects

### **Bonus**: [PORTALS-922](https://sagebionetworks.jira.com/browse/PORTALS-922)
There was instances in the portals where it would start making too many requests, it would load all the queries for querywrappermenu at once instead of when they were selected. Found out that there was a componentdidupdate issue where the token wasn't being examined correctly so in certain cases it would load them all.
_Additionaly_ , there was another issue with the QueryCount component making extraneous calls because its componentDidUpdate wouldn't see that it had already made a call.


leem42:
I think thats a great idea, to make that more useful I think that we should also look into configuring absolute imports https://create-react-app.dev/docs/importing-a-component/#absolute-imports which I had tried at the very beginning of this project but couldn't quite figure out.

PR #325, opened 2019-12-23
fix(SRC): get entity ID from query and not from props for SynapseTable

conundrumgirl:


conundrumgirl:
Should we get rid of `entityId` in QueryWrapperChildProps?

PR #326, opened 2020-01-02
Checkbox, RadioGroup, and Range Controls

conundrumgirl:
This PR adds Checkbox, RadioGroup and Range (for numbers and dates) controls into the /widgets
image

### DIFF
--- a/src/__tests__/lib/containers/widgets/RadioGroup.test.tsx
+++ b/src/__tests__/lib/containers/widgets/RadioGroup.test.tsx
@@ -1,0 +1,125 @@
+import * as React from 'react'
+import { mount, ReactWrapper } from 'enzyme'
+import {
+  RadioGroup,
+  RadioGroupProps,
+} from '../../../../lib/containers/widgets/RadioGroup'
+
+const mockCallback = jest.fn()
+
+function createTestProps(overrides?: RadioGroupProps): RadioGroupProps {
+  return {
+    options: [
+      { label: 'Label0', value: 'value0' },
+      { label: 'Label1', value: 'value1' },
+      { label: 'Label2', value: 'value2' },
+    ],
+    id: 'radioGroup1',
+    className: 'radioGroupClass',
+    value: 'value1',
+    onChange: mockCallback,
+
+    ...overrides,
+  }
+}
+
+let wrapper: ReactWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>
+let props: RadioGroupProps
+
+function init(overrides?: RadioGroupProps) {
+  props = createTestProps(overrides)
+  wrapper = mount(<RadioGroup {...props} />)
+}
+
+beforeEach(() => init())
+
+describe('basic function', () => {
+  it('should render with correct properties', async () => {
+    expect(wrapper.find('.radio')).toHaveLength(props.options.length)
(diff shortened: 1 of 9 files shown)


src/lib/containers/widgets/Range.tsx
```diff
@@ -0,0 +1,96 @@
...
+type ControlType = 'number' | 'date'
+
+export type RangeValues = {
+  min: string
+  max: string
+}
+
+export type RangeProps = {
+  type: ControlType
+  initialValues: RangeValues
```
leem42: make optional prop

src/__tests__/lib/containers/widgets/RadioGroup.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    ).toBe(true)
+    expect(
+      wrapper
+        .find('input')
+        .at(2)
+        .props().checked,
+    ).toBe(false)
+  })
+
+  it('should have correct item changed', async () => {
```
leem42: can remove async

src/__tests__/lib/containers/widgets/Range.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    ).toBe(true)
+  })
+
+  it('should call callbackFn correctly', async () => {
+    updateValue(wrapper, 2, 4.9)
+    expect(mockCallback).toHaveBeenCalledWith({ min: 2, max: 4.9 })
+  })
+  it('should not trigger callback for min > max and should show error', async () => {
+    updateValue(wrapper, 4.9, 2)
+    expect(mockCallback).not.toHaveBeenCalledWith()
```
leem42: expect(mockCallback).not.toHaveBeenCalledWith() -> expect(mockCallback).not.toHaveBeenCalled()

src/__tests__/lib/containers/widgets/Range.test.tsx
```diff
@@ -58,13 +62,34 @@ describe('number range ', () => {
...
+    expect(
+      wrapper
+        .find('input')
+        .at(1)
+        .props().value,
+    ).toBeUndefined()
+  })
+
+  it('should call callbackFn correctly', () => {
+    mockCallback.mockClear()
```
leem42: this call could be put in the beforeEach or inside the init function

src/lib/containers/widgets/Range.tsx
```diff
@@ -37,13 +40,18 @@ export const Range: React.FunctionComponent<RangeProps> = (
 
     type: ControlType = 'number',
   ) => {
+    if (!values.min || !values.max) {
```
leem42: I think this needs to explicitly check whether min or max is undefined, otherwise if min or max is zero this will incorrectly show an error message

src/lib/containers/widgets/Range.tsx
```diff
@@ -37,13 +40,18 @@ export const Range: React.FunctionComponent<RangeProps> = (
...
+    }
+
     if (type === 'number') {
       if (Number(min) > Number(max)) {
         setError(true)
         return false
       }
     }
-    if (Date.parse(min) > Date.parse(max)) {
+    if (min && max && Date.parse(min) > Date.parse(max)) {
```
leem42: same thing as above, these need to check !== undefined

PR #328, opened 2020-01-07
feat(rangeSlider): add rangeSliderWidget

conundrumgirl:
image
More info about the lib can be found here: https://www.npmjs.com/package/react-compound-slider/v/0.17.0

### DIFF
--- a/src/lib/containers/widgets/RangeSlider.tsx
+++ b/src/lib/containers/widgets/RangeSlider.tsx
@@ -1,0 +1,225 @@
+import * as React from 'react'
+import './RangeSlider.scss'
+
+import { Slider, Rail, Handles, Tracks, Ticks } from 'react-compound-slider'
+import {
+  SliderItem,
+  GetHandleProps,
+  GetTrackProps,
+} from 'react-compound-slider'
+import { RangeValues } from './Range'
+import { useState } from 'react'
+
+
+/**************** RANGE SLIDER SUBCOMPONENTS *********************/
+
+/***  handles ***/
+
+interface IHandleProps {
+  domain: number[]
+  handle: SliderItem
+  getHandleProps: GetHandleProps
+}
+
+export const Handle: React.FunctionComponent<IHandleProps> = ({
+  domain: [min, max],
+  handle: { id, value, percent },
+  getHandleProps,
+}) => (
+  <div
+    role="slider"
+    className="rangeSlider__handle"
+    aria-valuemin={min}
+    aria-valuemax={max}
+    aria-valuenow={value}
+    style={{
+      left: `${percent}%`,
+    }}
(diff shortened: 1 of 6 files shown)


src/lib/containers/widgets/RangeSlider.tsx
```diff
@@ -0,0 +1,225 @@
...
+  onChange: Function
+}
+
+export type RangeSliderState = {
+  values: readonly number[]
+}
+
+function getInitialValues(initialValues: RangeValues, domain: string[]) {
+  const result = [
+    initialValues.min ? +initialValues.min : +domain[0],
```
leem42: I've never seen this syntax before, does this add one to the values?

src/lib/containers/widgets/RangeSlider.tsx
```diff
@@ -0,0 +1,225 @@
...
+    <div
+      className="rangeSlider__tick"
+      style={{
+        left: `${tick.percent}%`,
+      }}
+    />
+    <div
+      className="rangeSlider__tickInner"
+      style={{
+        marginLeft: `${-(100 / count) / 2}%`,
```
leem42: Could this margin be positive and go from the right instead? e.g. marginRight: `${(100 / count) / 2}%`

conundrumgirl: I think we want to align the lefts vs the rights though

src/lib/containers/widgets/RangeSlider.tsx
conundrumgirl: It casts string to a number.

leem42: I see, that's neat, I think it would be easier to read the code if those were cast using Number(val)

src/lib/containers/widgets/RangeSlider.tsx
leem42: I would think it would act the same perhaps? but its not a big deal either way, ill resolve this,

conundrumgirl: Intuitively I agree. I think the problem is the apply button is a part of the general width.  I could probably refactor the layout and use left margin but since it's not a big deal I won't:)
image

src/lib/containers/widgets/RangeSlider.tsx
```diff
@@ -150,6 +150,7 @@ export const RangeSlider: React.FunctionComponent<RangeSliderProps> = ({doUpdate
 
   return (
     <div className="rangeSlider">
+      <div className="rangeSlider__values">{values[0]} - {values[1]}</div>
```
conundrumgirl: I added this so that the user can immediately see what the values they are setting before hitting apply.
image

PR #336, opened 2020-01-10
feat(filedDownload): display modal if user not logged in

conundrumgirl:
[PORTALS-970](https://sagebionetworks.jira.com/browse/PORTALS-970)

When downloading files from table, if you are not logged in the menu option is not enabled. User has no idea why. We should handle that case by showing information dialog explaining why its disable and what they can do to remedy that (log in).

image

### DIFF
--- a/src/__tests__/lib/containers/DownloadLoginModal.test.tsx
+++ b/src/__tests__/lib/containers/DownloadLoginModal.test.tsx
@@ -28,7 +28,7 @@ function init(overrides?: DownloadLoginModalProps) {
 beforeEach(() => init())
 
 describe('basic function', () => {
-  it('should render with explanqtory links and without login', () => {
+  it('should render with explanatory links and without login', () => {
     expect(wrapper.find('ModalHeader')).toHaveLength(1)
     expect(wrapper.find('ModalFooter').find('button')).toHaveLength(2)
     expect(
@@ -65,7 +65,6 @@ describe('basic function', () => {
     wrapper = wrapper.update()
     expect(wrapper.find('ModalFooter').find('button')).toHaveLength(1)
     expect(wrapper.find('Login')).toHaveLength(1)
-    const but = wrapper.find('Login').find('button')
     expect(
       wrapper
         .find('Login')
--- a/src/__tests__/lib/containers/DownloadLoginModal.test.tsx
+++ b/src/__tests__/lib/containers/DownloadLoginModal.test.tsx
@@ -1,0 +1,82 @@
+import { ReactWrapper, mount } from 'enzyme'
+import * as React from 'react'
+import {
+  DownloadLoginModal,
+  DownloadLoginModalProps,
+} from '../../../lib/containers/table/table-top/DownloadLoginModal'
+
+const mockCallback = jest.fn()
+
+function createTestProps(
+  overrides?: DownloadLoginModalProps,
+): DownloadLoginModalProps {
+  return {
+    showModal: true,
+    onHide: mockCallback,
+    ...overrides,
+  }
+}
+
+let wrapper: ReactWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>
+let props: DownloadLoginModalProps
+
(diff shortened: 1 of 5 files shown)


src/lib/containers/table/table-top/DownloadLoginModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (!isShowLogin) {
+      return (
+        <>
+          <p>
+            Anyone can browse public content on the Synapse we site, but in
+            order todownload and create content you will need to register for an
+            account using an email address
+          </p>
+          <p>
+            To find out more see, see{' '}
```
leem42: typo - "To find out more, see"

src/lib/containers/table/table-top/DownloadLoginModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+export type DownloadLoginModalProps = {
+  showModal?: boolean
+  onHide: Function
+}
+
+export const DownloadLoginModal: React.FunctionComponent<DownloadLoginModalProps> = props => {
+  const [showLogin, setShowLogin] = useState(false)
+
+  const getModalContent = (isShowLogin: boolean): JSX.Element => {
+    if (!isShowLogin) {
```
leem42: I find it a bit easier to read if the boolean is reversed in this specific case, so if (isShowLogin) return login else show content

src/__tests__/lib/containers/DownloadLoginModal.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+function init(overrides?: DownloadLoginModalProps) {
+  props = createTestProps(overrides)
+  wrapper = mount(<DownloadLoginModal {...props} />)
+}
+
+beforeEach(() => init())
+
+describe('basic function', () => {
+  it('should render with explanqtory links and without login', () => {
```
leem42: explanqtory -> explanatory

src/lib/containers/table/table-top/DownloadLoginModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return (
+        <>
+          <p>
+            Anyone can browse public content on the Synapse we site, but in
```
leem42: Some small typos in this text, here's the fixed version:

Anyone can browse public content on the Synapse website, but in order to download and create content you will need to register for an account using an email address.

src/lib/containers/table/table-top/DownloadLoginModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+            >
+              User Accounts
+            </a>
+            <span> and </span>
+            <a
+              href="https://docs.synapse.org/articles/governance.html"
+              target="_blank"
+            >
+              Governance Overview
+            </a>
```
leem42: from the design it looks like there should be a period after this anchor tag, or inside it I can't tell exactly.

src/__tests__/lib/containers/DownloadLoginModal.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    wrapper
+      .find('ModalFooter')
+      .find('button')
+      .at(1)
+      .simulate('click')
+    wrapper = wrapper.update()
+    expect(wrapper.find('ModalFooter').find('button')).toHaveLength(1)
+    expect(wrapper.find('Login')).toHaveLength(1)
+    const but = wrapper.find('Login').find('button')
```
leem42: it looks like this variable isn't used and can get removed

src/lib/containers/table/table-top/DownloadLoginModal.tsx
conundrumgirl: :flushed:

src/lib/containers/table/table-top/DownloadLoginModal.tsx
conundrumgirl: :woman_facepalming:

PR #337, opened 2020-01-13
PORTALS-966

leem42:
Fixed two bugs in this issue - 

1. CardContainerLogic's componentDidUpdate didn't account for the search params updating, if there are any suggestions for doing a diff between two objects I'd like to know!

2. There was another bug with getting the concreteType in synapse table where there was an endless loop of calls to synapse, I added a stop condition to componentDidUpdate.


src/lib/containers/CardContainerLogic.tsx
```diff
@@ -124,6 +124,25 @@ export default class CardContainerLogic extends React.Component<
     if (prevProps.sql !== this.props.sql) {
       this.executeInitialQueryRequest()
     }
+    const { searchParams: prevSearchParams = {} } = prevProps
+    const { searchParams: currentSearchParams = {} } = this.props
+    // check if search parameters have changed
+    if (
```
conundrumgirl: Why not use 
[_isEqual()](https://lodash.com/docs/4.17.15#isEqual) 

May be something like this?
```
  if ((prevProps.sql !== this.props.sql) || (!_.isEqual(prevProps.searchParams,  this.props.searchParams))
      this.executeInitialQueryRequest()
    }
```
I think this is all that needs to be done. 
Another alternative would be to make searchParams immutable on the parent level that way a new object gets created every time it changes and you can just test the object equality. 

Also seems like it might be a good candidate for a fn component with a hook? (that was purely a question)

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -167,21 +167,32 @@ export default class SynapseTable extends React.Component<
...
     this.getEntityHeadersInData()
+    this.getTableConcreteType(prevProps)
+  }
+
+  public async getTableConcreteType(
+    prevProps: QueryWrapperChildProps & SynapseTableProps,
+  ) {
+    const { data, token } = this.props
+    const previousTableId =
+      prevProps.data && prevProps.data.queryResult.queryResults!.tableId
```
conundrumgirl: Invitation for discussion: _'using _.get vs checking for undefined'_
we do a lot of this :
`const previousTableId = prevProps.data && prevProps.data.queryResult.queryResults!.tableId` throughout the codebase which can be instead accomplished with 
`const previousTableId = _.get(prevProps, 'data.queryResult.queryResults.tableId')`

I think that latter approach makes the code cleaner but you are loosing the type safety of TS.

Here is an interesting alternative:
https://codewithstyle.info/Deep-property-access-in-TypeScript/
custom util function?

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -167,21 +167,32 @@ export default class SynapseTable extends React.Component<
...
+  componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps) {
     this.getEntityHeadersInData()
+    this.getTableConcreteType(prevProps)
+  }
+
+  public async getTableConcreteType(
+    prevProps: QueryWrapperChildProps & SynapseTableProps,
+  ) {
+    const { data, token } = this.props
+    const previousTableId =
```
conundrumgirl: can we early out of there is no data?
```
 if (!data) {
  return
}
```
and then you don't need the `data&&` check below

src/lib/containers/CardContainerLogic.tsx
leem42: I agree with making this a hook, I'd like to do that in another PR.

PR #338, opened 2020-01-14
feat( PORTALS-975): Image button with tooltip control for top nav in tables

conundrumgirl:
https://sagebionetworks.jira.com/browse/PORTALS-975
- created the new control
- replaced the current icon/tooltip, image/tooltip, dropdown-toggle/tooltip in the table
- installed and using `testing-library/react` for testing

### DIFF
--- a/src/lib/containers/widgets/ImageButtonWithTooltip.tsx
+++ b/src/lib/containers/widgets/ImageButtonWithTooltip.tsx
@@ -12,11 +12,10 @@ type CustomImageProps = {
 
 type ImageButtonWithTooltipProps = {
   image: IconDefinition | CustomImageProps
-  callbackFn: Function
+  callbackFn?: Function
   tooltipText: string
   cssClass?: string
   idForToolTip: string
-  isDropdownToggle?: boolean
 }
 
 export const ImageButtonWithTooltip: FunctionComponent<ImageButtonWithTooltipProps> = ({
@@ -25,7 +24,6 @@ export const ImageButtonWithTooltip: FunctionComponent<ImageButtonWithTooltipPro
   callbackFn,
   tooltipText,
   cssClass,
-  isDropdownToggle,
 }) => {
   var buttonImage = <></>
   if ('svgImg' in image) {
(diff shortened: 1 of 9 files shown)


src/lib/containers/widgets/ImageButtonWithTooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const ImageButtonWithTooltip: FunctionComponent<ImageButtonWithTooltipProps> = ({
+  image,
+  idForToolTip,
+  callbackFn,
+  tooltipText,
+  cssClass,
+  isDropdownToggle,
+}) => {
+  var buttonImage = <></>
+  if ('svgImg' in image) {
```
leem42: I think we've generally used the syntax `if(image.svgImage) { }`, is there a reason to prefer using the `in` keyword?

src/lib/containers/widgets/ImageButtonWithTooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Dropdown } from 'react-bootstrap'
+
+type CustomImageProps = {
+  svgImg: string
+  altText: string
+}
+
+type ImageButtonWithTooltipProps = {
+  image: IconDefinition | CustomImageProps
+  callbackFn: Function
```
leem42: I see in other parts of the PR that `callbackFn` is passed an empty function, e.g. () => null, these empty fns could instead be removed by making callbackFn as an optional prop and then if this PR pull's in the latest develop you could change the syntax below `onClick={() => callbackFn()}` to use optional chaining onClick={() => callbackFn?()}

conundrumgirl: Thank you for pointing this out!
The logic is that it always needs a callbackFn unless it's a dropdown toggle.
So I'm getting rid of the toggle flag, making the callbackFn optional and use the logic of 
'if it doesn't have a callBack -- I assume it's a dropdown toggle'. Simplifies things a lot

src/lib/containers/widgets/ImageButtonWithTooltip.tsx
conundrumgirl: Yes, there is a subtle difference if how those work
`image` type can be either `CustomImageProps` or `IconDefinition` in which case it will not have the `svgImg` prop so `image.svgImage` would result in TS screaming about that.
the `in` operator tests if the  property is in the object (or its' prototype chain) 
for example in 
`const _someobj = { prop1: undefined, prop2: 'something else'}`
`('prop1' in _someobj)` will evaluate to true even if `_someobj.prop1` is falsey
So by using `in` operator I'm checking the type of the object rather than the value of the property

leem42: gotcha, this is good to know!

PR #347, opened 2020-01-20
PORTALS-934

leem42:
Changed EnumFacetFilter to use two react hooks - useGetProfiles and useGetEntityHeaders (new)
Changed EnumFacetFilter to use @testing-library/react for testing
Replaced set difference with lodash `without` function
Made a generic function getAllOfPaginatedService to get all the data for a paginated service, this was added to the entity header and access requirement calls
Fixed type issue with getUserProfileWithProfilePicAttached


src/lib/utils/hooks/useGetEntityHeaders.ts
```diff
@@ -0,0 +1,45 @@
...
+          .map(el => el.targetId),
+      )
+      const setDifference = difference(incomingListSet, curListSet)
+      if (setDifference.size > 0) {
+        try {
+          const newReferences = Array.from<string>(setDifference).map(el => {
+            return { targetId: el }
+          })
+          const data = await getEntityHeader(newReferences, token)
+          setData(data)
```
leem42: combine with old data

src/lib/utils/hooks/useGetEntityHeaders.ts
```diff
@@ -0,0 +1,45 @@
...
+      // look at current list of data, see if incoming ids has new data,
+      // if so grab those ids
+      const curList = data?.results.map(el => el.id)
+      const curListSet = new Set(curList)
+      const incomingListSet = new Set(
+        references
+          .filter(el => el.targetId !== SynapseConstants.VALUE_NOT_SET)
+          .map(el => el.targetId),
+      )
+      const setDifference = difference(incomingListSet, curListSet)
```
leem42: _.without

src/lib/utils/hooks/useGetEntityHeaders.ts
```diff
@@ -0,0 +1,45 @@
...
+import { getEntityHeader } from '../SynapseClient'
+import { difference } from '../functions/difference'
+import { SynapseConstants } from '..'
+
+export type UseGetProfilesProps = {
+  references: ReferenceList
+  token?: string
+}
+
+// React hook to get user profiles
```
leem42: TODO: Handle more than 50

src/lib/utils/hooks/useGetEntityHeaders.ts
leem42: make another function - getAllEntityHeaderArray -- would handle pagination and only return results

PR #348, opened 2020-01-20
fix(PORTALS-800): hide table if there are 0 rows

conundrumgirl:
https://sagebionetworks.jira.com/browse/PORTALS-800
Also have fixed a bug with the facet filter and adjusted some styles

No records returned as a result of user filtering:
image
No records returned 
image

### DIFF
--- a/src/lib/containers/table/SynapseTable.tsx
+++ b/src/lib/containers/table/SynapseTable.tsx
@@ -60,6 +60,7 @@ import {
 } from './table-top/'
 import FacetFilter from './table-top/FacetFilter'
 import { QueryFilter } from '../widgets/query-filter/QueryFilter'
+import NoData from '../../assets/icons/file-dotted.svg'
 
 const EMPTY_HEADER: EntityHeader = {
   id: '',
@@ -111,6 +112,7 @@ export type SynapseTableState = {
   mapUserIdToHeader: Dictionary<Partial<UserGroupHeader & UserProfile>>
   showColumnSelection: boolean
   isShowLeftFilter?: boolean
+  isUserModifiedQuery?: boolean
 }
 export type SynapseTableProps = {
   visibleColumnCount?: number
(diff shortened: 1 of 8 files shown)


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -358,28 +371,31 @@ export default class SynapseTable extends React.Component<
...
-              this.state.isShowLeftFilter
-                ? 'col-xs-12 col-sm-9 col-lg-9'
-                : 'col-xs-12'
-            }`}
-          >
-            {this.renderTable(headers, facets, rows)}
+            {this.state.isShowLeftFilter && (
+              <div
+                className="col-xs-12 col-sm-3 col-lg-3"
+                style={{ paddingRight: '0px' }}
```
conundrumgirl: The only change here is the style - removing padding for the left filter

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -1202,6 +1218,7 @@ export default class SynapseTable extends React.Component<
   public applyChangesFromQueryFilter = (facets: FacetColumnRequest[]) => {
     const queryRequest: QueryBundleRequest = this.props.getLastQueryRequest!()
     queryRequest.query.selectedFacets = facets
+    this.setState({ isUserModifiedQuery: true })
```
leem42: I understand what the variable name isUserModifiedQuery is getting at but I find it a bit confusing because the user could  modify the query in a number of ways, like with the old version of faceting or having sorted the columns of the data, but this variable doesn't apply to that.

 What do you think about renaming this variable to 'isQueryModifiedFromQueryFilter' or 'hasUserModifiedQueryWithQueryFilter'? I'm open to more concise suggestion as well

conundrumgirl: @jay-hodgson has mentioned that we are going to have another way to modify the query along with the new filters.
May be it would make sense to base the naming on 'if the user can change the criteria'. Basically what we are getting at is if we want to allow them to change the selection criteria or not.  So may be 'canUserChangeCriteria' does that make more sense?

leem42: I see more of what you mean now, it's a tricky name.. on second thought I think it's enough to just put a comment next to where the variable is declared and keep isUserModifiedQuery.

PR #352, opened 2020-01-23
refactor(PORTALS-652): reorganize sass files

conundrumgirl:
Reorganizing sass files using a modified [`7-1 pattern`](https://sass-guidelin.es/#architecture)
I have kept the original architecture for themed styles. We can revisit it later.
New file layout:
image
with expanded components:
image

Before:
image

I've changed the file names to use lowercase with the exception of Index.scss in template_styles to minimize the effect on portals. I have renamed the main style index.scss into main.scss since just chaning the case does not get picked up sometimes.
(style/Index.scss -> style->main.scss)
I have joined the styles from Core.scss and SWC.scss into base/core.scss This eventually should be refactored into the appropriate stylesheets. For right now I have just moved the most obvious (markdown) into it's own component stylesheet.

Once everything is agreed upon and merged I will update CONTRIBUTING.MD


src/lib/style/components/_query-filter.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../../style/Vars.scss';
+//ALINA
```
jay-hodgson: //ALINA ?

src/lib/style/vendors/_normalize.scss
```diff
@@ -0,0 +1,386 @@
+/*! normalize.css v5.0.0 | MIT License | github.com/necolas/normalize.css */
```
jay-hodgson: Is this all new, or was it moved?

conundrumgirl: It is new.  I was a little hesitant about it. There was a bunch of stuff in it that I have removed. On once hand it's generally a good idea to have a reset/normalize so that there is a common baseline, but it's a little risky with preexisting codebase. I left things that looked 'safe' to me. But I am not adverse to removing it entirely

src/lib/template_style/_theme.scss
```diff
@@ -3,6 +3,9 @@
 // params: $primary-action-color                            //
 //////////////////////////////////////////////////////////////
 
+
+$primary-action-color: #c94281 !default;
```
leem42: remove default

src/lib/template_style/_form.scss
```diff
@@ -1,6 +1,9 @@
 //////////////////////////////////////////////////////////////
 // params: $primary-action-color                            //
 //////////////////////////////////////////////////////////////
+
+$primary-action-color: #c94281 !default;
```
leem42: remove default

src/lib/style/components/synapse_form_wrapper/_steps-side-nav.scss
```diff
@@ -1,4 +1,6 @@
-@import './Vars.scss';
+@import '../../abstracts/variables';
```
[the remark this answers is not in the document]
leem42: can remove

PR #353, opened 2020-01-23
remove page reload calls and add callbacks

leem42:
Added callback functions to synapse client login/sign out functions
Used Universal-Cookie library to replace cookie functions in synapse client
Made it so the portal-side login modal shows up from synapsetable when clicking download files, this is the only way I could have the session callback for the portals get passed in without using a global variable.

### DIFF
--- a/src/demo/containers/App.tsx
+++ b/src/demo/containers/App.tsx
@@ -44,7 +44,7 @@ export default class App extends React.Component<{}, AppState> {
       })
   }
 
-  establishSession = () => {
+  getSession = () => {
     SynapseClient.detectSSOCode()
     SynapseClient.getSessionTokenFromCookie()
       .then(sessionToken => this.handleChange({ token: sessionToken }))
@@ -59,7 +59,7 @@ export default class App extends React.Component<{}, AppState> {
         You are logged in.&nbsp;
         <button
           onClick={() => {
-            SynapseClient.signOut(this.establishSession)
+            SynapseClient.signOut(this.getSession)
           }}
         >
           <span aria-hidden="true">Sign out</span>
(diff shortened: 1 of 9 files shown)


jay-hodgson:
Prefer to surface encoded query request bundle as an url query param

src/demo/containers/App.tsx
```diff
@@ -65,6 +74,7 @@ export default class App extends React.Component<{}, AppState> {
           token={
             SynapseClient.IS_OUTSIDE_SYNAPSE_ORG ? token : this.state.token
```
leem42: document on yarn link react in portals

src/lib/utils/SynapseClient.ts
```diff
@@ -955,35 +958,52 @@ export const getWikiAttachmentsFromEvaluation = (
...
+    if (!token) {
+      cookies.remove(SESSION_TOKEN_COOKIE_KEY)
+    } else {
+      // set's cookie in session storage
+      cookies.set(SESSION_TOKEN_COOKIE_KEY, token, {
+        // expires in a day
+        maxAge: 60 * 60 * 24,
+      })
+    }
+    sessionCallback?.()
```
leem42: make mandatory

src/lib/utils/SynapseClient.ts
```diff
@@ -955,35 +958,52 @@ export const getWikiAttachmentsFromEvaluation = (
...
+    return
```
leem42: return void

src/lib/utils/SynapseClient.ts
```diff
@@ -1049,20 +1069,10 @@ export const detectSSOCode = () => {
...
-        console.error('err when clearing the session cookie ', err)
-      })
-  } else {
-    // is not on a synapse.org subdomain
-    removeCookie(SESSION_TOKEN_COOKIE_KEY)
-    window.location.reload()
-  }
+export const signOut = (sessionCallback?: Function) => {
+  setSessionTokenCookie(undefined, sessionCallback).catch(err => {
+    console.error('err when clearing the session cookie ', err)
```
leem42: nothing to catch, remove

leem42: make callback mandatory

PR #354, opened 2020-01-23
Update CONTRIBUTING.md

conundrumgirl:
Proposed format for commit messages 
I wonder if we can leverage this: https://www.npmjs.com/package/auto-changelog and if it makes sense

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@
 
 1. Fork this repository
 2. Create your feature branch, if applicable name it after the specific JIRA issue being fixed.
-3. Try to keep commits organized by `squashing` closely related commits. If it makes sense format your commit in the following way: `<type>(<JIRA>): <subject><BLANK LINE><body> or link to the JIRA` [^1] .
+3. To keep commits organized consider `squashing` closely related commits (optional). When it makes sense format your commit in the following way: `(<JIRA>): <subject><BLANK LINE><body or link to the JIRA if applicable (optional)>`.
 4. Create a pull request against `develop` from your forked branch.
 5. Ensure someone else review's the pull request and that person should be the one to merge the PR.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -13,8 +13,9 @@
 
 1. Fork this repository
 2. Create your feature branch, if applicable name it after the specific JIRA issue being fixed.
-3. Create a pull request against `develop` from your forked branch.
-4. Ensure someone else review's the pull request and that person should be the one to merge the PR.
+3. Try to keep commits organized by `squashing` closely related commits. If it makes sense format your commit in the following way: `<type>(<JIRA>): <subject><BLANK LINE><body> or link to the JIRA` [^1] .
```
jay-hodgson: What does the additional overhead (of defining commit type and squashing for a clean commit history) buy us?
Having the Jira issue ID makes sense to me, and I would think that the Jira issue would provide a full description of the change type.

conundrumgirl: squashing makes the history a little cleaner/shorter and easier to read. Same with commit type. e.g. if it's just a style and you are looking for a code change you will skip it.
But I agree that sometimes there is not a clear distinction.
So then` (<JIRA>): <subject><BLANK LINE><body if needed>` ? works for me

leem42: I think using that commit message format would be a great addition, it makes commit messages much more valuable. And it makes writing the description of the pull request easier because you can read the commit messages. I'm hesitant with commit squashing though, I can see the argument for it, but I would make the case that it should be a contributor preference.

conundrumgirl: @leem42 re:squashing yes, I should have communicated more clearly. It's definitely  should be more of a 'be mindful of ..' situation rather than a rule

conundrumgirl:
Opening https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/356/files instead

PR #355, opened 2020-01-28
(PORTALS-746): long value handling in tables

conundrumgirl:
If the column value length > 20 chars trim it and show the full value on hover

image

### DIFF
--- a/src/lib/containers/table/SynapseTable.tsx
+++ b/src/lib/containers/table/SynapseTable.tsx
@@ -1,4 +1,4 @@
-import { ImageButtonWithTooltip } from '../widgets/ImageButtonWithTooltip'
+import { ElementWithTooltip } from '../widgets/ElementWithTooltip'
 import { IconProp, library } from '@fortawesome/fontawesome-svg-core'
 import {
   faCheck,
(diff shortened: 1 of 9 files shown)


src/lib/containers/widgets/ImageButtonWithTooltip.tsx
```diff
@@ -11,40 +11,57 @@ type CustomImageProps = {
 }
 
 type ImageButtonWithTooltipProps = {
```
conundrumgirl: The change in this component allows for a string value to be a trigger - not just an image.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -995,7 +1009,22 @@ export default class SynapseTable extends React.Component<
...
-      return <p className={isBold}> {columnValue} </p>
+      const [displayString, isShortened] = getShortString(columnValue)
+      if (!isShortened) {
+        return <p className={isBold}> {columnValue}</p>
+      } else {
+        return (
+          <p className={isBold}>
+            <ImageButtonWithTooltip
+              tooltipText={columnValue}
+              callbackFn={_.noop}
```
leem42: if callbackFn is an optional prop can't it be left out instead of passing _.noop?

conundrumgirl: No because the absence of calllbackFn will make the component render the dropdown since that's how it decides if it's a button or a dropdown trigger. For a button we do expect some kind of callback since most the the time there is some kind of action to perform. If this seems hacky to you I can add another prop to component to explicitly flag that it is not a dropdown trigger. I don't have strong feelings either way

leem42: Ah I see, I would push towards making ImageButtonWithTooltip have a third option to not render as a button since clicking doesn't have an action for this case, the component could be renamed to something like ElementWithTooltip.

Granted, there are other parts of SRC that aren't perfect in this regard (anchors that don't have hrefs etc.).

PR #357, opened 2020-01-29
PORTALS-987

jay-hodgson:


leem42:
@jay-hodgson did you validate that this works on portals? I believe this will work, but double checking doesn't hurt. 

I looked at the portals code again and it actually gives you full reign of the subpath and all search parameters passed down to it, you can see here - https://github.com/Sage-Bionetworks/portals/blob/a92d0362d8a9fbf76face50f1b7b832fb80507e0/src/RouteResolver.tsx#L108 all it does when search parameters are passed in is see if a programmatic route was specified.

jay-hodgson:
Verified that it works in portals: https://youtu.be/ii7kK-qr7MU

PR #362, opened 2020-01-31
Remove Node.js require statement

jay-hodgson:
and fix potential memory leak in rss feed (do not set state if component is no longer mounted)


jay-hodgson:
This does not fix the umd build.  Looks like a recently introduced dependency (universal-cookie) is to blame this time.

src/lib/containers/RssFeed.tsx
```diff
@@ -34,16 +36,28 @@ export default class RssFeed extends React.Component<RssFeedProps, RssState> {
       itemId2MoreItem: {},
     }
   }
-
+  
   componentDidMount() {
+    this._isMounted = true
```
leem42: I would suggest removing the _isMounted variable, if it's set to true at the top of this method then in theory it should always be true when being checked on lines 47 and 52.

PR #363, opened 2020-02-03
feat(PORTALS-982): 

leem42:
- Added outline of access list component, added necessary api calls
- Added type parameter to doPost synapse client function. Added type params to all functions calling doPost. 
- Added type to create entity, returns entity based on entity type passed in
- Updated contributing.md to include section on adding a new component


src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Returns rendering for the access requirement.
+   *
+   * Only supports SelfSignAccessRequirement and TermsOfUseAccessRequirement
+   *
+   * @param {AccessRequirement} accessRequirement accessRequirement being rendered
+   */
+  const renderAccessRequirement = (accessRequirement: AccessRequirement) => {
+    switch (accessRequirement.concreteType) {
+      case 'org.sagebionetworks.repo.model.SelfSignAccessRequirement':
```
conundrumgirl: declare those as constants?

src/lib/containers/access_requirement_list/SelfSignAccessRequirement.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import { AccessRequirement } from 'lib/utils/synapseTypes/AccessRequirement/AccessRequirement'
+
+type Props = {
```
conundrumgirl: is the generic name 'Props' intentional?

src/lib/utils/SynapseClient.ts
```diff
@@ -1797,6 +1803,78 @@ export const getRestrictionInformation = (
...
+  id: string,
+): Promise<Array<AccessRequirement>> => {
+  let isMoreData = true
+  const accessRequirementResults = [] as AccessRequirement[]
+  const limit = 50
+  let offset = 0
+  while (isMoreData) {
+    const data = await getAccessRequirement(sessionToken, id, limit, offset)
+    accessRequirementResults.concat(data.results)
+    if (data.totalNumberOfResults < 50) {
```
conundrumgirl: I don't think  `accessRequirementResults.concat(data.results)` modifies accessRequirementResults. I think it should.

leem42: Whoops! Good call, that would've been a fun bug

PR #367, opened 2020-02-10
PORTALS-775: moving away from hashrouter

jay-hodgson:


jay-hodgson:
Need to release a new SRC version since portals-side set of changes depend on these.

PR #370, opened 2020-02-11
SWC-4925: add renderer support for new multi-value list column types

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -963,14 +978,37 @@ export default class SynapseTable extends React.Component<
         />
       )
     }
+    if (dateListColumnIndicies.includes(colIndex)) {
+      const jsonData:number[] = JSON.parse(columnValue)
+      return jsonData.map(
+        (val: number, index: number) => {
+          return <span className={isBold}>
+            {new Date(Number(val)).toLocaleString()}{index !== jsonData.length-1 ? ', ' : ''}
```
leem42: If val is being cast to a Number, I would think that mean's val is not actually a number but instead a string? Which makes me think that jsonData should be a string array, e.g. jsonData: string [] to be more accurate

jay-hodgson: Copy/paste error.  columnValue is a string for a single Date, but it really is a number in this case (no need to construct a Number object!).  Good catch.

PR #372, opened 2020-02-17
if a HeaderCard is being shown, then update the page title and descri…

jay-hodgson:
…ption using that information

### DIFF
--- a/src/lib/containers/HeaderCard.tsx
+++ b/src/lib/containers/HeaderCard.tsx
@@ -1,5 +1,5 @@
-import * as React from 'react'
 import { CardFooter, Icon } from './row_renderers/utils'
+import React, { useState, useEffect } from 'react'
 
 export type IconOptions = {
   [index: string]: string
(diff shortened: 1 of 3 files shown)


src/lib/containers/HeaderCard.tsx
```diff
@@ -29,6 +29,15 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
   secondaryLabelLimit,
   isAlignToLeftNav
 }) => {
+  // update page title and description based on header card values
+  if (title && document.title !== title) {
+    document.title = title
+  }
+  const descriptionElement:Element|null = document.querySelector('meta[name="description"]')
```
leem42: Are we covering all cases of setting the document title in the portals? My concern is that a page will have the wrong document title and metadata with this sort of modification existing on a per component basis, as opposed to in the portal code.

src/lib/containers/HeaderCard.tsx
```diff
@@ -29,6 +29,15 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
...
+  if (descriptionElement && (description || subTitle)) {
+    descriptionElement.setAttribute('content', description ? description : subTitle)
+  }
```
leem42: On componentWillUnmount can reset document title and content, need to save original on mount

src/lib/containers/HeaderCard.tsx
```diff
@@ -29,6 +29,27 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
...
 }) => {
+  // store old document title and description so that we can restore when this component is removed
+  const descriptionElement:Element|null = document.querySelector('meta[name="description"]')
+  const [docTitle] = useState<string>(document.title)
+  const [docDescription] = useState<string>(descriptionElement ? descriptionElement.getAttribute('content')! : '')
+  useEffect(() => {
+    return function cleanup() {
+      document.title = docTitle
+      if (descriptionElement) {
+        descriptionElement.setAttribute('content', docDescription)
```
leem42: for future reference, you can use the new optional chaining syntax when a variable might not be defined, e.g. 
 descriptionElement?.setAttribute('content', docDescription)

PR #374, opened 2020-02-18
SWC-5052 continued

jay-hodgson:
- Show Access column if not logged in (may be open data).  If unable to get the FileHandle and not logged in, then indicate user needs to first sign in.
- Distinguish HasAccess being used in a DownloadList vs other contexts.  Only show warnings about being part of a DownloadOrder package if item is in the DownloadList.  Add tests for HasAccess outside of DownloadList.

### DIFF
--- a/src/lib/utils/SynapseClient.ts
+++ b/src/lib/utils/SynapseClient.ts
@@ -1438,8 +1438,12 @@ export const getFileEntityFileHandle = (
       }
       getFiles(request, sessionToken)
         .then((data: BatchFileResult) => {
-          const fileHandle: FileHandle = data.requestedFiles[0].fileHandle!
-          resolve(fileHandle)
+          if (data.requestedFiles.length > 0 && data.requestedFiles[0].fileHandle) {
+            resolve(data.requestedFiles[0].fileHandle)  
+          } else {
+            // not found, or not allowed to access
+            reject(undefined)
+          }
         })
         .catch(err => {
           reject(err)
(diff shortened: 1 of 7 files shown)


src/lib/utils/SynapseClient.ts
```diff
@@ -1409,6 +1409,49 @@ export const getFileEntityContent = (
...
+    getEntity(sessionToken, fileEntityId, fileEntityVersionNumber).then((entity:Entity) => {
+      const fileEntity:FileEntity = entity as FileEntity
+      const fileHandleAssociationList = [
+        {
+          associateObjectId: fileEntity.id,
+          associateObjectType: 'FileEntity',
+          fileHandleId: fileEntity.dataFileHandleId,
+        },
+      ]
+      const request: any = {
```
leem42: can remove the :any here, the type will be enforced by getFiles

src/lib/utils/SynapseClient.ts
```diff
@@ -1409,6 +1409,49 @@ export const getFileEntityContent = (
...
+ * @param fileEntityId: ID of the FileEntity
+ * @param endpoint
+ */
+export const getFileEntityFileHandle = (
+  fileEntityId: string,
+  fileEntityVersionNumber?: string,
+  sessionToken?: string,
+): Promise<FileHandle> => {
+  return new Promise((resolve, reject) => {
+    getEntity(sessionToken, fileEntityId, fileEntityVersionNumber).then((entity:Entity) => {
```
leem42: can add type to getEntity call - `getEntity<FileEntity>` and then remove the cast below

PR #379, opened 2020-02-21
add markdown rendering for description

leem42:
Render markdown in the description card field


src/lib/containers/GenericCard.tsx
```diff
@@ -324,4 +328,61 @@ export default class GenericCard extends React.Component<
...
+    if (descriptionLinkConfig) {
+      return <></>
+    }
+    return (
+      <div className={hasClickedShowMore ? 'SRC-hidden' : ''}>
+        <span
+          data-search-handle={descriptionSubTitle}
+          className={`SRC-font-size-base ${CARD_SHORT_DESCRIPTION_CSS} SRC-short-description`}
+        >
+          {getCutoff(description).previewText}
```
jay-hodgson: No markdown rendering in the short description?

src/lib/containers/GenericCard.tsx
```diff
@@ -53,17 +58,33 @@ export const DOI_REGEX = /^10.\d{4,9}\/[-._;()/:a-z0-9]+$/
...
+  let previewText = ''
+  const summarySplit = summary!.split(' ')
+  // find num words to join such that its >= char_count_cutoff
+  let i = 0
+  while (previewText.length < CHAR_COUNT_CUTOFF && i < summarySplit.length) {
+    previewText += `${summarySplit[i]} `
+    i += 1
+  }
+  previewText = previewText.trim()
+  return { previewText }
```
jay-hodgson: Is there a reason why you're returning an object with previewText instead of just previewText?  Certainly would make sense if there were multiple return values...

PR #385, opened 2020-02-28
added support for facet in url, removed menu index param

leem42:


src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -191,9 +191,10 @@ export default class MarkdownSynapse extends React.Component<
         ol: ['class'],
         span: ['*'],
         table: ['class'],
-        th: ['class'],
```
leem42: go back in!

PR #397, opened 2020-03-05
SWC-5125: check to see if user already consented to request.

jay-hodgson:
Note that this call will be available on next week's staging Synapse stack (unavailable today)


src/lib/utils/SynapseClient.ts
```diff
@@ -1570,6 +1571,23 @@ export const getOAuth2RequestDescription = (
   )
 }
 
+/**
+ * Check whether user has already granted consent for the given OAuth client, scope, and claims.
+ * Consent persists for one year.
+ */
+export const hasUserAuthorizedOAuthClient = (
+  oidcAuthRequest: OIDCAuthorizationRequest,
+  sessionToken: string | undefined,
```
jay-hodgson: @conundrumgirl noticed that the token is allowed to be undefined here.  But this case should not be allowed (the token might be invalid, but it should be set if calling this function!).

PR #398, opened 2020-03-05
PORTALS-1082

leem42:
Handle rendering multi value secondary labels in card, revert to using selected facets in query wrapper menu

Can see example multi value rendering here in theme label -

image

### DIFF
--- a/src/lib/containers/QueryWrapperMenu.tsx
+++ b/src/lib/containers/QueryWrapperMenu.tsx
@@ -194,7 +194,10 @@ export default class QueryWrapperMenu extends React.Component<
     }
   }
 
-  public getCurrentSqlFromActiveTab = () => {
+  public getCurrentSqlFromActiveTab = (): {
+    sql: string
+    selectedFacets: FacetColumnValuesRequest[] | undefined
+  } => {
     const { searchParams, menuConfig } = this.props
     const { activeMenuIndices } = this.state
     let facetValue = ''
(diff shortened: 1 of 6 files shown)


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -979,36 +988,44 @@ export default class SynapseTable extends React.Component<
       )
     }
     if (dateListColumnIndicies.includes(colIndex)) {
-      const jsonData:number[] = JSON.parse(columnValue)
-      return jsonData.map(
-        (val: number, index: number) => {
-          return <span className={isBold}>
-            {new Date(val).toLocaleString()}{index !== jsonData.length-1 ? ', ' : ''}
+      const jsonData: number[] = JSON.parse(columnValue)
```
conundrumgirl: It seems that we can abstract this into a function that would take column value and type
```
function outputValues<T>(columnValue: string, type: string) {
 return JSON.parse(columnValue).map((val: string, index: number) => {
        return (
          <span className={isBold}>
           // different handling here depending on the type passed  {val ? 'true' : 'false'}
            {index !== jsonData.length - 1 ? ', ' : ''}
          </span>
        )
      })
}```

File: src/lib/containers/QueryWrapperMenu.tsx [hunk 2]
```diff
@@ -235,8 +240,9 @@ export default class QueryWrapperMenu extends React.Component<
     } = this.props
 
     let sql = ''
+    let selectedFacets = undefined
     if (menuConfig) {
-      sql = this.getCurrentSqlFromActiveTab()
+      ;({ sql, selectedFacets } = this.getCurrentSqlFromActiveTab())
```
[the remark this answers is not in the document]
leem42: It's a bit awkward that eslint puts the semi-colon on the left hand side instead of the right hand side, but its there to allow for the sql and selected facets to be assigned from the destructured statement, mozilla has a better explanation here under 'Assignment without declaration'
  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

src/lib/containers/table/SynapseTable.tsx
leem42: I will punt this to another PR, but that would be a great rewrite

PR #403, opened 2020-03-06
fix TotalQueryResults calculation

leem42:


src/lib/containers/TotalQueryResults.tsx
```diff
@@ -36,66 +33,35 @@ export default class TotalQueryResults extends React.Component<
...
   componentDidUpdate(prevProps: TotalQueryResultsProps) {
-    // check that its done loading
-    if (!this.props.isLoading && prevProps.isLoading) {
+    // Check that it has entered a loading state, meaning that there is new data
+    if (this.props.isLoading && !prevProps.isLoading) {
       this.calculateTotal()
     }
   }
 
   calculateTotal() {
```
conundrumgirl: I think this should return a promise

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -36,66 +33,35 @@ export default class TotalQueryResults extends React.Component<
   }
 
   componentDidUpdate(prevProps: TotalQueryResultsProps) {
```
conundrumgirl: May be having it return a promise would solve testing issue

PR #415, opened 2020-03-09
fix(PORTALS-1086): add 'research themes' label on the left

conundrumgirl:
image

### DIFF
--- a/src/lib/containers/widgets/themes-plot/ThemesPlot.tsx
+++ b/src/lib/containers/widgets/themes-plot/ThemesPlot.tsx
@@ -29,6 +29,7 @@ export type ThemesPlotProps = {
   topBarPlot: PlotProps
   sideBarPlot: PlotProps
   tooltipProps?: TooltipVisualProps
+  dotPlotYAxisLabel?: string
 }
 
 type TotalsGroup = { y: string; count: number }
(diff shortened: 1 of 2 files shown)


src/lib/containers/widgets/themes-plot/ThemesPlot.tsx
```diff
@@ -287,97 +293,100 @@ const ThemesPlot: FunctionComponent<ThemesPlotProps> = ({
...
-                      <br />
-                      <BarPlot
-                        style={{ width: '100%' }}
-                        layoutConfig={barLayoutConfig}
-                        optionsConfig={optionsConfig}
-                        plotData={sideBarPlotData}
-                        isTop={false}
-                        xMax={xMaxForSideBarPlot}
+          <div style={{ display: 'flex', position: 'relative'}}>
+            <div className="ThemesPlot__dotPlotYLabel">{dotPlotYAxisLabel}</div>
```
conundrumgirl: This is the only major change.. the resulting diff is mostly reshuffling of HTML since the extra div was added

jay-hodgson: I'd argue that your __dotPlotYLabel class is doing most of the work!

PR #418, opened 2020-03-12
fix(PORTALS-1101): fix the infinite loop. Better fix to account for a…

conundrumgirl:
…ccess restriction
When we output the label in the Facet Enum we can instead of a name have the same treatment as synapse with a !. But it's a purely UI  concern


src/lib/utils/hooks/useGetEntityHeaders.ts
```diff
@@ -30,6 +41,17 @@ export default function useGetEntityHeaders(props: UseGetProfilesProps) {
...
+            )
+            const notFoundPlaceholders: EntityHeader[] = notFound.map(item => ({
+              ...entityHeaderTemplate,
+              id: item.targetId,
+              name: `Unknown/Unable to Access (${item.targetId})`
+            }))
+            setData(oldData =>
+              oldData.concat(...newData, ...notFoundPlaceholders),
+            )
             setData(oldData => oldData.concat(...newData))
```
leem42: this line can get removed with the addition of the line above

conundrumgirl: Good catch!
Thank you. Had it stashed.. and didn't quite resolve the conflicts correctly. Fixed

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -0,0 +1,125 @@
...
+        try {
+          const newIds = Array.from<string>(newValues)
+          const newReferences: LookupRequestType[] =
+            type === 'USER_PROFILE'
+              ? newIds
+              : newIds.map(el => ({ targetId: el }))
+          const newReferencesChunks = chunk(newReferences, 45)
+
+          for (const newReferences of newReferencesChunks) {
+            const newData =
```
leem42: Just thinking out loud, if we have more types that this hook supports in the future then we'll want to make this a switch statement and possibly put these `get...Items` functions in another file, but for now this works, same with idProp from above

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -0,0 +1,125 @@
...
+const getEntityHeaderItems = async (
+  lookupList: ReferenceList,
+  token: string | undefined,
+): Promise<EntityHeader[]> => {
+  const newData = await getEntityHeader(lookupList as ReferenceList, token)
+  const notFound = lookupList.filter(
+    item =>
+      newData.results.map(item => item.id).indexOf(item.targetId) == -1,
+  )
+  const notFoundPlaceholders = notFound.map(item => ({
```
leem42: One downside of this notFoundPlaceholders approach is that if the user logs in then any stored data won't update with the new token, if we could know that the token specifically changed in the useEffect then we could call this function all over again, but I don't believe that information is surfaced from useEffect. This problem isn't unique to this hook (or the previous implementation for that matter), any of our hooks which store data over multiple updates will have stale information from the token getting updated, I think this issue will have to be put on the back burner.

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -0,0 +1,125 @@
...
+    ownerId: item,
+    name: `Unknown User (${item})`,
+  }))
+
+  return [...newData.list, ...notFoundPlaceholders]
+}
+
+// React hook to get user profiles or entities
+//******************************************************************************************* */
+export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
```
leem42: Typed hook - shweet!

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -0,0 +1,125 @@
...
+  modifiedOn: 'null', //	The date this entity was last modified.
+  createdBy: 'null', //	The ID of the user that created this entity.
+  modifiedBy: 'null', //	The ID of the user that last modified this entity.
+}
+
+const getEntityHeaderItems = async (
+  lookupList: ReferenceList,
+  token: string | undefined,
+): Promise<EntityHeader[]> => {
+  const newData = await getEntityHeader(lookupList as ReferenceList, token)
```
leem42: is this cast `lookupList as ReferenceList` necessary?

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  return [...newData.results, ...notFoundPlaceholders]
+}
+
+const getUserProfileItems = async (
+  lookupList: string[],
+  token: string | undefined,
+): Promise<UserProfile[]> => {
+  const newData = await getUserProfileWithProfilePicAttached(
+    lookupList as string[],
```
leem42: is this cast `lookupList as string[]` needed?

src/lib/utils/hooks/useGetInfoFromIds.ts
conundrumgirl: my understanding that hook will rerun because the token is in the
[dependency array](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/418/files/e2e08bb1a90550ef36763493be4c21238b6df0ef#diff-7b41d0b46439218454ac9ff0541f5830R123)
You don't think so?

src/lib/utils/hooks/useGetInfoFromIds.ts
conundrumgirl: Good catch! I was moving the code around and forgot to remove. Thanks!

src/lib/utils/hooks/useGetInfoFromIds.ts
leem42: I agree that it will re-run with the token change and any new data coming in will be fine, but the existing data won't update. Since dummy objects are stored for all the synids that don't have a return item -- the ones the user doesn't have access too -- when the user log's in those won't update.

e.g.
1. If useGetInfoFromIds is called without a token (so logged out) and a list of entity header ids, let's say `[syn123, syn456, syn789]` the first two syn ids get a return mapped to them `[{name: 'zzzz', id: 'syn123'}, {name: 'xxxx', id: 'syn789'}]` and the last one `syn789` the user didn't have permission to view when signed out, so it has a faux entry stored `[{name: 'No Header', id: 'syn789'}]`

2. Now the user sign's in with their updated token (which should give them access to `syn789`) - the useEffect only call's get entity header if the list of incoming synIds isn't contained in the stored list of entity headers. In the case of an updated token, the incoming synIds are no different, only the token changed, so now `syn789` is stored as this faux entry and the data is stale.

conundrumgirl: I see what you are saying! I wonder how easy it would be to set up the test case.  I am thinking (a hunch;) that in practical terms changing the token would cause the rerender of the parent component which would it turn cause the rerender of the hook. May be I can test that.

PR #426, opened 2020-03-18
PORTALS-1170: Updated AccessRequirement Demo

OCEO-YJ:


src/demo/containers/playground/AccessRequirementDemo.tsx
```diff
@@ -1,23 +1,57 @@
...
+
   return (
-    <AccessRequirementList
-      onHide={() => {
-        window.location.href = 'http://localhost:3000/Playground'
-      }}
-      entityId={'syn3850484'}
-      token={token}
-    />
+    <div>
```
leem42: I think this can be simplified to only be the second part of the ternary statement - 

e.g.
 <>
          <p>SynId for Self Sign Access: syn3850484</p>
          <p>SynId for Terms Of Use: syn2426398</p>
          <p>SynId for Managed ACT Access: syn3191087</p>
          <p>SynId for ACT Access: syn12634840</p>

          <input
            id="AccessRequirementInput"
            type="text"
            placeholder="Synapse Id"
            onChange={event => {
              setInputValue(event.target.value)
            }}
          ></input>
          <button
            className="get-access-requirement-button"
            onClick={getAccessRequirement}
          >
            Get Access Requirement
          </button>
        </>

src/demo/containers/playground/AccessRequirementDemo.tsx
```diff
@@ -1,23 +1,57 @@
...
+      {displayAccessRequirement ? (
+        <AccessRequirementList
+          onHide={() => {
+            window.location.href = 'http://localhost:3000/Playground'
```
leem42: this should be something like this setDisplayAccessRequirement('') to reset the page

PR #435, opened 2020-03-24
Add DownloadList to Portals

leem42:
# Issues found and fixed

### Two async calls in HasAccess were causing 4x the number of calls
scenario:
- `componentDidMount()` is called with two functions inside - `asyncCall_1` and `asyncCall_2`
- `asyncCall_1` finishes before `asyncCall_2` and call's `setState`
- `componentDidUpdate` is triggered because of the `setState` call
- `asyncCall_1` and `asyncCall_2` are both called by `componentDidUpdate`
- `asyncCall_2` from the first `componentDidMount` finishes and call's `setState`
- `componentDidUpdate` is called (note - the first `componentDidUpdate` hasn't finished yet..)
-  ... until finally this all stop's because both stop conditions for `asyncCall_1` and `asyncCall_1` are met

### DownloadListTable renders as a modal
- The download list renders as a modal now to support the ShowDownload component,  removed invokeDownloadListUpdatedEvent by calling listUpdatedCallback() directly using typescript conditional properties

### DownloadConfirmation makes too many API calls on updates to props
One of the props to DownloadConfirmation is the queryBundleRequest object, this object break's the react memoization process because that only uses direct equality comparison. I fixed this by using the `use-deep-compare-effect` which handles this exact issue by doing a deep diff on the props coming in.

### Added ShowDownload component
(Need a better name for this) but this implements the download preview icon in the top right of the screen. The component listens for changes to the download list by a global event and updates itself accordingly.

### DIFF
--- a/src/lib/containers/download_list/DownloadListTable.tsx
+++ b/src/lib/containers/download_list/DownloadListTable.tsx
@@ -2,7 +2,7 @@ import { library } from '@fortawesome/fontawesome-svg-core'
 import { faTrash } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import moment from 'moment'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useState, useCallback } from 'react'
 import * as ReactBootstrap from 'react-bootstrap'
 import { calculateFriendlyFileSize } from '../../utils/functions/calculateFriendlyFileSize'
 import useGetInfoFromIds from '../../utils/hooks/useGetInfoFromIds'
@@ -62,9 +62,13 @@ export default function DownloadListTable(props: DownloadListTableProps) {
   // https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables
   const [isLoading, setIsLoading] = useState<LoadingState>(true)
   const [fileBeingDeleted, setFileBeingDeleted] = useState<string>('')
-  const [showAsModalState, setShowAsModalState] = useState(true)
   const { token } = props
-  const { forceSamePage = false, renderAsModal = false, onHide } = props
+  const {
+    forceSamePage = false,
+    renderAsModal = false,
+    onHide,
+    listUpdatedCallback,
+  } = props
   const { references, batchFileResult, downloadList } = data
   const requestedFiles =
     (batchFileResult && batchFileResult.requestedFiles) || []
@@ -80,11 +84,7 @@ export default function DownloadListTable(props: DownloadListTableProps) {
     type: 'USER_PROFILE',
   })
 
-  useEffect(() => {
-    fetchData(token)
-  }, [token])
-
-  const fetchData = async (token: string | undefined) => {
+  const fetchData = useCallback(async () => {
     if (!token) {
       setIsLoading(false)
       // doesn't make sense with anonymous user!
@@ -98,7 +98,6 @@ export default function DownloadListTable(props: DownloadListTableProps) {
         setData({
           downloadList,
         })
-        invokeDownloadListUpdatedEvent()
         return
       }
       const batchFileRequest: BatchFileRequest = {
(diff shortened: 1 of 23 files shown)


src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -80,46 +76,50 @@ export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
...
+                  token,
+                )
+          setData(oldData => oldData.concat(...(newData as T[])))
         }
+      } catch (error) {
+        console.error('Error on data retrieval', error)
+      } finally {
+        setIsLoading(false)
       }
     }
```
leem42: there's an issue in all of our hooks components which store data over time - they'll hold stale data if the user ever log's in. If we use another useEffect that only takes in token as a dev dependency then we can capture if the token changes and retrieve the right data.

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -80,46 +76,50 @@ export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
...
   const [data, setData] = useState<Array<T>>([])
+  const [isLoading, setIsLoading] = useState(false)
 
   const idProp = (type: HookType) =>
     type === 'USER_PROFILE' ? 'ownerId' : 'id'
 
-  useEffect(() => {
+  const getData = useCallback(async () => {
+    if (isLoading) {
+      // We use isLoading to ensure that we don't load extra data while its process the data in batches
```
leem42: found this bug while working on downloadlist

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -188,25 +187,25 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+  const results = references?.results ?? []
   let numBytes = 0
   let numFiles = 0
-  return (
-    <div>
-      <div className="SRC-split download-list-table-top">
-        <span className="create-package-text SRC-centerContentInline">
+  const content = (
+    <div className="DownloadListTable">
+      <div className="SRC-split download-list-table-top SRC-primary-background-color SRC-border-bottom-only">
```
conundrumgirl: I am thinking that here and in a similar situation it might make sense to use BEMs  instead of having a long list of clases. e.g. here we can have <div className = DownloadListTable__top that would have all the properties that we need . The in .scss file we'd have 
`DownloadListTable 
  $__top {
         all the props we need
`

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -188,25 +187,25 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+        <span className="create-package-text">
           Download List &nbsp;&nbsp; {isLoading && <span className="spinner" />}
         </span>
         <button
-          className="SRC-primary-text-color SRC-underline-on-hover"
+          className="SRC-underline-on-hover uppercase-text"
```
conundrumgirl: same comment as above about classes

src/lib/style/components/_download-link.scss
```diff
@@ -0,0 +1,57 @@
+.Download-Link {
+  position: relative;
+  display: inline-block;
+  .icon-container {
```
conundrumgirl: if we are to use bems - this class would be `Download-Link__icon container ` and then in the scss we'd have
```
.Download-Link {
  position: relative;
  display: inline-block;
 &__icon-container {
...
```

src/lib/containers/download_list/DownloadListTable.tsx
leem42: I agree with that! I have this issue https://sagebionetworks.jira.com/browse/PORTALS-1199 to track that

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -80,46 +76,50 @@ export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
 ) {
   const { token, ids, type } = props
   const [data, setData] = useState<Array<T>>([])
+  const [isLoading, setIsLoading] = useState(false)
 
   const idProp = (type: HookType) =>
     type === 'USER_PROFILE' ? 'ownerId' : 'id'
 
-  useEffect(() => {
+  const getData = useCallback(async () => {
```
leem42: try without useCallback!

PR #438, opened 2020-03-25
fix(PORTALS-1095): persist query parameters to URL in portals

conundrumgirl:
https://sagebionetworks.jira.com/browse/PORTALS-1095

I think a good way to test it would be to select a facet on the left and do some filtering. Wait of data to load. Copy the url, paste it into a different tab and see what happens

### DIFF
--- a/src/lib/containers/QueryWrapperMenu.tsx
+++ b/src/lib/containers/QueryWrapperMenu.tsx
@@ -15,11 +15,15 @@ import {
   isGroupByInSql,
   insertConditionsFromSearchParams,
 } from '../utils/functions/sqlFunctions'
-import { FacetColumnValuesRequest } from '../utils/synapseTypes/'
+import {
+  FacetColumnValuesRequest,
+  QueryBundleRequest,
+} from '../utils/synapseTypes/'
 import { library } from '@fortawesome/fontawesome-svg-core'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faPlus, faSearch } from '@fortawesome/free-solid-svg-icons'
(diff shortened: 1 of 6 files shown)


src/lib/containers/QueryWrapperMenu.tsx
```diff
@@ -15,11 +15,15 @@ import {
   isGroupByInSql,
```
leem42: I noticed that this component doesn't update when navigating with the back and forward buttons on the browser, I think this might have to get wrapped in withRouter to trigger componentDidUpdate on URL changes?

conundrumgirl: @leem42 - I have created[https://sagebionetworks.jira.com/browse/PORTALS-1188]( https://sagebionetworks.jira.com/browse/PORTALS-1188) to track this

src/lib/containers/QueryWrapperMenu.tsx
```diff
@@ -182,14 +183,21 @@ export default class QueryWrapperMenu extends React.Component<
...
       })
-      const facetName = this.props.menuConfig?.[menuIndexIn].facet
-      DeepLinkingUtils.updateUrlWithNewSearchParam(
-        'menuIndex',
-        undefined,
-        accordionIndexIn.toString(),
-      )
-      if (facetName) {
-        DeepLinkingUtils.updateUrlWithNewSearchParam('facet', undefined, facetName)
+
```
leem42, line 186: I found one small bug - `shouldDeepLink` need's to get forwarded to QueryWrapper on line 375, then this can get merged!

PR #441, opened 2020-04-01
PORTALS-1072

jay-hodgson:
Adds a SingleNewsFeedPost component - all url query params are passed through to the rss feed url.

### DIFF
--- a/src/lib/containers/SingleNewsFeedPost.tsx
+++ b/src/lib/containers/SingleNewsFeedPost.tsx
@@ -1,0 +1,32 @@
+import React, { useState } from 'react' // importing FunctionComponent
+import RssFeed from './RssFeed'
+
+export type SingleNewsFeedPostProps = {
+  feedUrl: string 
+}
+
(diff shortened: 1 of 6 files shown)


src/lib/containers/SingleNewsFeedPost.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const [modifiedFeedUrl, setModifiedFeedUrl] = useState()
+  const {feedUrl} = props
+  let updatedModifiedFeedUrl = feedUrl
+  const urlParams = new URLSearchParams(window.location.search)
+  if (urlParams) {
+    urlParams.forEach((value, key) => {
+      updatedModifiedFeedUrl = `${updatedModifiedFeedUrl}&${key}=${value}`
+    })
+  }
+  if (updatedModifiedFeedUrl && modifiedFeedUrl !== updatedModifiedFeedUrl)
```
leem42, line 18: I think state can be avoided all together here, should be able to use updatedModifiedFeedURL below and remove the modifiedFeedURL variable

PR #459, opened 2020-04-07
PORTALS 1100: Removed Cog UI and Updated DropDown Menu

OCEO-YJ:


src/lib/containers/table/table-top/EllipsisDropdown.tsx
```diff
@@ -39,6 +44,7 @@ export const EllipsisDropdown: React.FunctionComponent<EllipsisDropdownProps> =
         <Dropdown.Menu
           className="SRC-primary-color-hover-dropdown"
           alignRight={true}
+          style={{ width: '100%', right: '-100px' }}
```
OCEO-YJ, line 47: This is removed

PR #463, opened 2020-04-13
(PORTALS-1200) - Implement visualization based data exploration w/ facets

conundrumgirl:
facetNav

https://sagebionetworks.jira.com/browse/PORTALS-1200

It is still WIP


src/lib/containers/synapse_table_functions/renderTableCell.tsx
```diff
@@ -143,15 +143,15 @@ export const renderTableCell = ({
       return <p className={isBold}> {columnValue}</p>
     } else {
       return (
-        <p className={isBold}>
```
conundrumgirl, line 146: This is unrelated change. Just fixing 'can't nest div in p' error

src/lib/containers/widgets/query-filter/QueryFilter.tsx
```diff
@@ -75,7 +75,7 @@ const patchRequestFacets = (
   return selections
 }
 
-const applyChangesToValuesColumn = (
+export const applyChangesToValuesColumn = (
```
conundrumgirl, line 78: This function is also used  by facet nav

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const hideFacetInGrid = (columnName: string) => {
+    setUiPropertyForFacet(columnName, 'isHidden', true)
+  }
+
+  const setUiPropertyForFacet = (
+    columnName: string,
+    propName: 'isHidden' | 'isExpanded',
+    value: boolean,
+  ) => {
+    const cloneArray = _.cloneDeep(facetUiStateArray)
```
leem42, line 163: What do lines 163-167 do? It looks like they could be removed

src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -0,0 +1,292 @@
...
+import { SizeMe } from 'react-sizeme'
+import {
+  faFilter,
+  faExpandAlt,
+  faCompressAlt,
+  faTimes,
+} from '@fortawesome/free-solid-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+import { QueryWrapperChildProps } from 'lib/containers/QueryWrapper'
```
leem42, line 18: all these lib imports will have to change to absolute ones at some point

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -0,0 +1,259 @@
...
+  FacetColumnRequest,
+  FacetColumnResult,
+} from 'lib/utils/synapseTypes'
+import { useState, useEffect } from 'react'
+import * as _ from 'lodash'
+
+export type FacetNavOwnProps = {
+  applyChanges: Function
+  loadingScreen?: React.FunctionComponent | JSX.Element
+  facetsToPlot?: string[] | undefined
```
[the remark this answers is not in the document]
conundrumgirl: This is optional and if it is not defined the full enum facets array is used. At one point Ljubo has mentioned that we could add this in synapse as one of the table config properties. So this is there to account for this potential functionality.

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -0,0 +1,259 @@
...
+      facetUiStateArray.map(item =>
+        item.name === columnName ? { ...item, [propName]: value } : item,
+      ),
+    )
+  }
+
+  if (isLoadingNewData || !data) {
+    return (
+      <div className="SRC-loadingContainer SRC-centerContentColumn">
+        {!!loadingScreen && loadingScreen}
```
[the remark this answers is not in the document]
conundrumgirl: Yeah I wondered about it as well.. I will change

src/lib/containers/widgets/facet-nav/FacetNav.tsx
conundrumgirl: Nice catch! Totally I reworked this logic and forgot to remove old implementation :see_no_evil:

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -0,0 +1,259 @@
...
+    let result = data?.facets?.filter(
+      item =>
+        item.facetType === 'enumeration' &&
+        (!facetsToPlot?.length || facetsToPlot.indexOf(item.columnName) > -1),
+    )
+    if (!result) {
+      return
+    }
+    if (isFirstTime) {
+      setFacetUiStateArray(
```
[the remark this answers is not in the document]
conundrumgirl: The problem is that initially there might not be data since it's an async call so while we do need to do it once we do need to do it after the data is present

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -0,0 +1,259 @@
...
+        result.map((item, index) => ({
+          name: item.columnName,
+          isHidden: index < 4 ? false : true,
+          isExpanded: false,
+        })),
+      )
+      setIsFirstTime(false)
+    }
+    setFacets(result)
```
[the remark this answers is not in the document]
conundrumgirl: You are correct. I will put a TODO here. The reason why I'd rather not refactor this right now is that this area is likely to change depending on how we are going to handle show/hide. For example if we'd put the hidden facets in the back of the queue  rather than just hiding then in their current position we would need to maintain a queue of facets and in that case it would make sense to keep them in the state. I will add a note to make sure it does not fall through the cracks

PR #464, opened 2020-04-15
PORTALS-986

leem42:
### HasAccess
- Tooltip for unauthorized ACL restriction has been corrected
- Situation for a blocked ACL and blocked AR now shows the blocked AR state
- Bubble's up AR to DownloadList if set_arPropsFromHasAccess has been set
- Removed logic in constructor that made state harder to reason about
- Consolidated failed SynapseClient.getFileEntityFileHandle logic to the outer most  SynapseClient.getEntity call
- Exposed Access Restriction Modal [PORTALS-986](https://sagebionetworks.jira.com/browse/PORTALS-986)

### DownloadList 
- Fixed issue with download items not showing created by or created on, filled in the data by using the headers
- Fixed double modal situation with HasAccess restriction by bubbling up the AccessRequirementList to the download list modal
- Fixed created by not working [PORTALS-1243](https://sagebionetworks.jira.com/browse/PORTALS-1243), there was a race condition in useGetInfoFromIds where the current loading state would block the component from recognizing the updated props

### AccessRequirementList
- Fixed status request being made twice for ARs, there was a batch call in the top level list component that could be passed down to the children
- CSS font tweaks and alignment, added file icon

### AccessRequirement* Components
- All share the same prop interface for simplicity

### MarkdownSynapse
- Fixed tiny bug in key prop not being set

### ShowDownload
[PORTALS-1246](https://sagebionetworks.jira.com/browse/PORTALS-1246) - added tooltip
[PORTALS-1242](https://sagebionetworks.jira.com/browse/PORTALS-1242) - avoid closing download list on clear list button click

### Misclaneous
Added useCompare.ts, which will help with open issue of hooks not updating from token change

### PORTALS-1249
isLoadingNewData is set to false when there's an error in the network request, fixes the bar chart from forever spinning


src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -84,31 +109,43 @@ export default function AccessRequirementList({
...
           return parseInt(status.accessRequirementId)
         })
 
-      requirements = requirements.sort((requirementA, requirementB) => {
-        return (
-          sortOrder.indexOf(requirementA.id) -
-          sortOrder.indexOf(requirementB.id)
-        )
-      })
+      const requirementsAndStatuses = requirements
```
conundrumgirl, line 116: just an FYI lodash has a [sortBy](https://lodash.com/docs/4.17.15#sortBy) and [orderBy](https://lodash.com/docs/4.17.15#orderBy) which might make things like that easier

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -84,31 +109,43 @@ export default function AccessRequirementList({
...
+          return
+        }
         if (!accessRequirementFromProps) {
-          getAllAccessRequirements(token, entityId).then(async requirements => {
-            const sortedAccessRequirements = await sortAccessRequirementByCompletion(
-              requirements,
-            )
-            setAccessRequirements(sortedAccessRequirements)
-          })
+          getAllAccessRequirements(token, entityId).then(
```
conundrumgirl, line 141: may be use async await for both calls?

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -193,128 +210,159 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+              let contentSize = undefined
+              const synId = item.associateObjectId
+              const fileHandleId = item.fileHandleId
+              const isCurrentlyBeingDeletedClass =
+                fileBeingDeleted === fileHandleId ? 'SRC-inactive-bg' : ''
+              // See if batch file results has this fileHandleId
+              const fileResult = requestedFiles.find(
+                (fileRes) => fileRes.fileHandleId === fileHandleId,
+              )
+              const fileHandle = fileResult ? fileResult.fileHandle : undefined
```
conundrumgirl, line 253: would ` const fileHandle = fileResult?.fileHandle` work? nbd just wondering

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -218,18 +218,18 @@ export default class SynapseTable extends React.Component<
     // Make call to resolve entity ids
     if (distinctEntityIds.size > 0) {
       const referenceList: ReferenceList = Array.from(distinctEntityIds).map(
-        id => {
+        (id) => {
```
conundrumgirl, line 221: do you have a configuration that adds () ? I think it might be new

src/lib/style/components/_access-requirement-list.scss
```diff
@@ -1,93 +1,123 @@
...
+  }
 
-.self-sign-access-title {
-  margin-top: 20px;
-}
+  .view-terms-button {
+    color: rgb(77, 85, 144);
+  }
+  .view-terms {
+    &.hide {
```
conundrumgirl, line 58: There is a '.hidden' bootstrap class that does
`{
    display: none!important;
}`

src/lib/style/components/_access-requirement-list.scss
```diff
@@ -1,93 +1,123 @@
...
+    text-align: center;
+    color: rgb(77, 85, 144);
+  }
+
+  .self-sign-access-title {
+    margin-top: 20px;
+  }
+
+  .self-sign-access-certified-success-text {
+    &.hide {
```
conundrumgirl, line 89: per comment above I think if you'll add .hidden rather than .hide you will not need extra css definition

src/lib/utils/hooks/usePrevious.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from 'react'
+
```
conundrumgirl, line 2: What does it do?

src/lib/containers/table/SynapseTable.tsx
leem42: I'm not sure why this has been getting added lately, my eslint doesn't show this as a preference, Ill look into it more so we avoid these unnecessary diffs, thanks for calling this out.

src/lib/utils/hooks/usePrevious.ts
leem42: I added this because the `useCompare` hook requires it. And I added `useCompare` to track whether a value has changed, most notably for when the user's token has changed. I saw this from a stack overflow post - https://stackoverflow.com/a/56817365.

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -31,34 +47,40 @@ export enum SUPPORTED_ACCESS_REQUIREMENTS {
...
-  let hasUnSupported = false
   for (let i = 0; i < accessRequirements.length; i++) {
-    if (
-      accessRequirements[i].concreteType !==
-        SUPPORTED_ACCESS_REQUIREMENTS.TermsOfUseAccessRequirement &&
-      accessRequirements[i].concreteType !==
-        SUPPORTED_ACCESS_REQUIREMENTS.SelfSignAccessRequirement
-    ) {
-      hasUnSupported = hasUnSupported || true
+    if (isARUnsupported(accessRequirements[i].concreteType)) {
```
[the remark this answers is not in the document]
leem42: ah this is so much better, will add this, thank you!

PR #465, opened 2020-04-15
PORTALS-1022 - cont.

conundrumgirl:
- refactor  TotalQueryResults to use hooks
- add filters to the Facet visualizations
lates demo is on [gh-pages](https://conundrumgirl.github.io/Synapse-React-Client/Playground/QueryWrapperPlotNavDemo)

### DIFF
--- a/src/lib/containers/TotalQueryResults.tsx
+++ b/src/lib/containers/TotalQueryResults.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react'
 import { QueryBundleRequest } from '../utils/synapseTypes'
 import { SynapseClient, SynapseConstants } from '../'
+import useDeepCompareEffect from 'use-deep-compare-effect'
 
 export type TotalQueryResultsProps = {
   isLoading: boolean
   style?: React.CSSProperties
-  getLastQueryRequest: (() => QueryBundleRequest) | undefined
+  //getLastQueryRequest: (() => QueryBundleRequest) | undefined
+  lastQueryRequest: QueryBundleRequest 
   token: string | undefined
   unitDescription: string
   frontText: string
@@ -18,23 +20,24 @@ const TotalQueryResults: React.FunctionComponent<TotalQueryResultsProps> = ({
   style,
   unitDescription,
   frontText,
-  getLastQueryRequest,
+  //getLastQueryRequest,
+  lastQueryRequest,
   token,
   isLoading: parentLoading,
 }) => {
   const [total, setTotal] = React.useState(0)
   const [isLoading, setIsLoading] = React.useState(false)
 
-  React.useEffect(() => {
+  useDeepCompareEffect(() => {
     const calculateTotal = async () => {
-      const queryRequest = getLastQueryRequest!()
-      queryRequest.partMask =
+     // const lastQueryRequest = getLastQueryRequest!()
+      lastQueryRequest.partMask =
         SynapseConstants.BUNDLE_MASK_QUERY_COUNT |
         SynapseConstants.BUNDLE_MASK_QUERY_FACETS
 
-      if (parentLoading) {
+      if (!parentLoading) {
         setIsLoading(true)
-        SynapseClient.getQueryTableResults(queryRequest, token)
+        SynapseClient.getQueryTableResults(lastQueryRequest, token)
           .then(data => {
             setTotal(data.queryCount!)
           })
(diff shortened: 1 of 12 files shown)


src/lib/containers/TotalQueryResults.tsx
```diff
@@ -11,72 +13,52 @@ export type TotalQueryResultsProps = {
...
-        this.setState({
-          isLoading: false,
+        .finally(() => {
+          setIsLoading(false)
         })
-      })
-  }
+    }
+    calculateTotal()
+  }, [parentLoading, token, getLastQueryRequest])
```
leem42, line 49: I think this useEffect would calculate the facet results twice for every time a query is run, once when the query is kicked off (parenLoading = true) and then once more after its finished (parentLoading = false), but I believe that it should only calculate the query once (parentLoading = true). I think the logic to check parentLoading has changed to true need's to be tracked and checked, I have this logic in my PR [here](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/464/files#diff-838d52273514f635c18c31b169632988R92) but for token change.

I believe each time this component gets new props getLastQueryRequest will be a new function thats passed in, although it would (or could) contain the same query request from before. I think one way around this could be to pass in the last query request directly and use useDeepCompareEffect on that object, [like this](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/305e181dc8c57ee9563d0dd899dd5144c875c3e9/src/lib/containers/download_list/DownloadConfirmation.tsx#L144).

src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -217,44 +266,70 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = ({
...
-           
-            {onExpand && (
-              <button onClick={() => onExpand(index)}>
+            <FacetFilter
+              lastFacetSelection={{
+                columnName: '',
+                facetValue: '',
+                selector: '',
+              }}
+              isLoading={false}
```
leem42, line 294: should this be hard coded to false?

conundrumgirl: probably not:) Thank you!

src/lib/containers/TotalQueryResults.tsx
conundrumgirl: My intuition would be to kick it off once parentLoading == false (when things settle). The performance is indeed better when it gets kicked off while parent is loading ( parentLoading == true). And it does make sense since the query they are executing is the same. I did have to modify the test for it to pass. But I just want to bring it up to make sure that kicking it off while parent is still loading does not produce any unwanted effects

leem42: I agree with that assessment! I guess there's a fairly small possibility of a race condition introduced here - if the user makes two query updates quickly then its possible that the second result is overriden by the first result if the two come back out of sync.

e.g.

1. Make facet selection (parentLoading === true), total query results kicks off a query, call it A
2. Make facet selection, (the following still holds parentLoading === true) so total query results kicks off a second query, call it B
3. B resolves and setTotal is called
4. A resolves and setTotal is called
5. TotalQueryResults now has an incorrect total from A instead of B

I dunno how much this is worth worrying about, it _seems_ pretty unlikely

leem42: I still have concern with getLastQueryRequest being passed as a dependency, I'm mostly concerned that it _could_ be a problem, its an unknown.

conundrumgirl: May be I'm missing something but in practical terms when would the props change? We have :
```
 style,
  unitDescription,
  frontText,
  getLastQueryRequest,
  token,
  isLoading: parentLoading,
```
seems like only last 3 are the ones that would get changed during the life on the component and we do need to watch for those any way.  On a related note, why do we not pass lastQueryRequest directly to the QueryWraper child components and pass the function instead?

leem42: That's a good point about passing lastQueryRequest directly, I think you're right we should do that! then useDeepCompareEffect can get be used to do the deep diff on the object. And I agree that only the last 3 should change change, seems right to me too.

conundrumgirl: I have observed that when we are fetching when `parentLoading === true` - the first time the table comes up after clicking on left nav the total is 0. This might be the case where the the case of race condition or something else so I have changed it `parentLoading === false` check

PR #466, opened 2020-04-17
Query Results Count w' filter pills

conundrumgirl:
[Demo Here ](https://conundrumgirl.github.io/Synapse-React-Client/Playground/QueryWrapperPlotNavDemo)
1. Added facet names to count bar. 
- Clicking 'x' on the pill removes the facet selection
- Clicking on the user or entity facet should display it correctly
2. Added bar graph
 - If you expand the facet plot there is another icon to toggle to the bar graph


src/__tests__/lib/containers/widgets/facet-nav/FacetNavPanel.test.tsx
```diff
@@ -84,10 +84,11 @@ describe('initialization', () => {
     expect(panel).toHaveLength(1)
 
     const buttons =  container.querySelectorAll<HTMLElement>('button > svg')
-    expect(buttons.length).toBe(3)
-    expect((buttons[0].getAttribute('data-icon'))).toBe('filter')
-    expect((buttons[1].getAttribute('data-icon'))).toBe('compress-alt')
-    expect((buttons[2].getAttribute('data-icon'))).toBe('times')
+    expect(buttons.length).toBe(4)
+    expect((buttons[0].getAttribute('data-icon'))).toBe('chart-bar')
```
conundrumgirl, line 88: Added Bar Graph

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -82,6 +82,9 @@ export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
   const idProp = (type: HookType) =>
     type === 'USER_PROFILE' ? 'ownerId' : 'id'
```
conundrumgirl, line 83: The problem with hooks is that there are restrictions e..g they can't be used conditionally or inside of the callback.
Long term I think we need to store entity/user mapping in context or just local storage. But that's a refactor for another day. For now -- this is somewhat hacky but if we know that there is another component that's using the same entities as the component that uses the hook this seems to work. (for now)

src/lib/utils/functions/getDataFromFromStorage.ts
```diff
@@ -0,0 +1,24 @@
+import { EntityHeader, UserProfile } from '../synapseTypes'
+import { SynapseConstants } from '..'
+
+export const getStoredEntityHeaders = (): EntityHeader[] => {
+    try {
+      const lookUpEntityHeaders: EntityHeader[] = JSON.parse(
+        sessionStorage.getItem(SynapseConstants.ENTITY_HEADER_STORAGE_KEY) || '',
```
leem42, line 7: Can change at some point from `SynapseConstants.ENTITY_HEADER_STORAGE_KEY) || ''` to `SynapseConstants.ENTITY_HEADER_STORAGE_KEY) ?? ''`

src/lib/utils/hooks/useGetInfoFromIds.ts
leem42: I have some amount of concern with session storage, I think it mostly comes from having not used it before. But like you said we'll eventually get around to having a react context/store like way of saving these entities.

PR #473, opened 2020-04-21
taking queryfilter out

conundrumgirl:
This is purely an FYI -- pls do not merge.
This change to the query filter allows to take it out of the table and just add it separately.
In this configuration you will see it both as a part of the table as well as outside of it.
For illustration purposes only


leem42:
closing since this issue has been resolved

PR #474, opened 2020-04-21
PORTALS-1256: added early exit if search params are QueryWrapper definitions

jay-hodgson:
or there are no search params.  Also, parser returns DBLSTRINGs from tokenizer, and then fails when attempting to reconstruct the same sql using it's own parser!  Replace all DBLSTRINGs with LITERAL, but I don't think we'll ever get into this situation today because we don't mix the study page conditional with the encoded Query search param.

### DIFF
--- a/src/lib/utils/functions/sqlFunctions.ts
+++ b/src/lib/utils/functions/sqlFunctions.ts
@@ -51,6 +51,12 @@ export const insertConditionsFromSearchParams = (
   sql: string,
   operator: SQLOperator = 'LIKE',
 ) => {
+  // if there are no search params, or if all search params are QueryWrapper queries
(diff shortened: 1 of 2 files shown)


src/lib/utils/functions/sqlFunctions.ts
```diff
@@ -81,6 +87,12 @@ export const insertConditionsFromSearchParams = (
 export const formatSQLFromParser = (tokens: string[][]) => {
   // remove backtick from output sql (for table name): `syn1234` becomes syn1234
   const synId = tokens[tokens.findIndex(el => el[0] === 'FROM') + 1][1]
+  // replace all DBLSTRINGs (escaped strings) with LITERALs
+  tokens.forEach((value) => {
```
leem42, line 91: nice! this has been an issue for almost a year now, gold star for you! 🌟

PR #492, opened 2020-05-11
PORTALS-1021: Implemented Sorting Functionality for Download List

OCEO-YJ:


src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,28 +218,99 @@ export default function DownloadListTable(props: DownloadListTableProps) {
     }
   }
 
+  const sortColumn = async (column: string) => {
+    try {
+      setIsLoading(true)
+
```
leem42, line 224: Instead of `setColumns` you would have this

```tsx
// store isAscending to be passed into sort function below
const isAscending = column === sortedColumn.column ? !sortedColumn.isAscending: true

setSortedColumn({
   column,
   isAscending
})
```

The `column` can be set to the column no matter what the value is given, doesn't matter if it's setting the same value or if its a new one, but for is`Descending` we need extra steps. If the column coming in was the last column set, then swap the sort direction (e.g. if they click the sort button twice for one column) .

Otherwise, we sort ascendingly for the new column that is being sorted.

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,28 +218,99 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+        file: column === 'file' ? !columns[column] : false,
+        createdBy: column === 'createdBy' ? !columns[column] : false,
+        createdOn: column === 'createdOn' ? !columns[column] : false,
+        size: column === 'size' ? !columns[column] : false,
+      })
+
+      const filesToDownload = downloadList?.filesToDownload ?? []
+
+      filesToDownload.sort((itemA, itemB) => {
+        return sortDownLoadList(itemA, itemB, column)
```
leem42, line 235:
```tsx
// pass in isAscending to the sort function
return sortDownLoadList(itemA, itemB, column, isAscending)
```

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,28 +218,99 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
     )
-
     const fileHandle = fileResult ? fileResult.fileHandle : undefined
 
+    // fileHandle is defined, this file is downloadable, show its metadata
     if (fileHandle && itemA) {
-      fileNameA = fileHandle.fileName
+      fileName_A = fileHandle.fileName
+      createdBy_A = fileHandle.createdBy
+      createdOn_A = fileHandle.createdOn
```
leem42: I think there's a bug here, right now this code only look's at the detailed information for itemA, but not itemB. I would refactor to pull out the information for both using a helper function -

```tsx
const getFileHandleInfo = (
    item: FileHandleAssociation,
) => {
    const fileResult = requestedFiles.find(
      fileRes => fileRes.fileHandleId === item.fileHandleId,
    )
    let fileName = ''
    let createdBy = ''
    let createdOn_A = ''
    let contentSize = ''
    if (fileHandle && item) {
        fileName = fileHandle.fileName
        createdBy = fileHandle.createdBy
        createdOn = fileHandle.createdOn
        contentSize =fileHandle.contentSize
    } else {
       // file is not downloadable, only show its name from entity header info
      const requestedFile = results.find(req => req.id === item.associateObjectId)
      if (requestedFile) {
        fileName = requestedFile?.name
        createdBy = requestedFile?.createdBy
        createdOn = requestedFile?.createdOn
     }
    }
   return { fileName, createdBy,  createdOn,  contentSize }
}

const { fileName: fileName_A, createdBy: createdBy_A,  createdOn: createdOn_A,  contentSize: contentSize_A } = getFileHandleInfo(itemA)
const { fileName: fileName_B, createdBy: createdBy_B,  createdOn: createdOn_B,  contentSize: contentSize_B } = getFileHandleInfo(itemB)

```

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,28 +218,99 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+      console.error(err)
+    } finally {
+      setIsLoading(false)
+    }
+  }
+
   const sortDownLoadList = (
     itemA: FileHandleAssociation,
     itemB: FileHandleAssociation,
+    column: string,
```
leem42, line 252: also need to pass in `isAscending` to the function so the sort can switch directions

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,28 +218,99 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+        createdBy_A = requestedFile?.createdBy
+        createdOn_A = requestedFile?.createdOn
+      } else if (itemB) {
+        fileName_B = requestedFile?.name
+        createdBy_B = requestedFile?.createdBy
+        createdOn_B = requestedFile?.createdOn
+      }
     }
 
-    return fileNameA.localeCompare(fileNameB)
```
leem42: Here's where you'll want to use the `isAscending` parameter - 
```tsx
const direction = isAscending ? 1: -1
switch(column) {
    case 'file':
       return fileName_B?.localeCompare(fileName_A!)! * direction
     ...
}
```

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -227,17 +345,105 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+                  className={`sort SRC-primary-background-color-hover ${
+                    sortedColumn.column === 'createdBy'
+                      ? 'SRC-primary-background-color'
+                      : ''
+                  }`}
+                  onClick={() => {
+                    sortColumn('createdBy')
+                  }}
+                >
+                  <FontAwesomeIcon
```
leem42, line 384: we need to set the color of the icon to white if the column has been clicked, the logic for the icon state can be simplified by getting rid of the nested ternary statement
e.g.
```jsx
  <FontAwesomeIcon
      icon={
        sortedColumn.column === 'size'
          && sortedColumn.isDescending === true ?
           : faSortAmountUp
           : faSortAmountDown
      }
      color={sortedColumn.column === 'size' ? 'white' : ''}
      />
```

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -204,6 +221,107 @@ export default function DownloadListTable(props: DownloadListTableProps) {
     }
   }
 
+  const sortColumn = async (column: string) => {
+    try {
+      setIsLoading(true)
+
+      const isDescending =
+        column === sortedColumn.column ? !sortedColumn.isDescending : true
```
leem42, line 229: lets change this so that by default the column is ascending -

```jsx
  const isDescending =
        column === sortedColumn.column ? !sortedColumn.isDescending : false
```

PR #504, opened 2020-05-20
264: add a way to show checkboxes for every visible row, and a way to add "custom controls" (main CTA buttons) for the Table

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -917,6 +947,14 @@ export default class SynapseTable extends React.Component<
             <span style={{ whiteSpace: 'nowrap' }}>Access</span>
           </div>
         </th>,
+      )      
+    }
+    if (isRowSelectionVisible) {
+      tableColumnHeaderElements.unshift(<th key="rowSelectionColumn">
+          <div className="SRC-centerContent">
+            {/* checkboxes column */}
```
leem42, line 955: Should this have column header text 'WorkflowState'?

src/lib/utils/synapseTypes/Table/TableUpdate.ts
```diff
@@ -0,0 +1,36 @@
+// http://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/TableUpdateTransactionRequest.html
+// Update a table
+export type TableUpdateTransactionRequest = {
+  concreteType: 'org.sagebionetworks.repo.model.table.TableUpdateTransactionRequest'
+  entityId: string
+  changes: AppendableRowSetRequest[] //List of changes that describes schema and/or row changes to a table.
+  // createSnapshot?: boolean
```
leem42, line 7: would having the 'createSnapshot' options set serve any sort of record keeping utility?

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: No, this is the checkbox column.  The WorkflowState column is another one (that my custom command happens to be targeting).

src/lib/utils/synapseTypes/Table/TableUpdate.ts
jay-hodgson: Yes, it's a way to automatically create a version of the Table/View.  If we add this functionality, I'd rather make this an explicit call (rather than combining with a Table/View Update).

PR #514, opened 2020-05-27
PORTALS-1295: Update plus/minus images from design

emmanmills:


src/lib/containers/widgets/query-filter/FacetFilterHeader.tsx
```diff
@@ -1,5 +1,8 @@
 import * as React from 'react'
 import { unCamelCase } from '../../../utils/functions/unCamelCase'
+import iconPlus from '../../../assets/icons/icon_plus.svg'
+import iconMinus from '../../../assets/icons/icon_minus.svg'
+import '../../../style/components/query_filter/_facet-filter-header.scss'
```
leem42, line 5: Instead of importing the scss file here, if you could import it [here](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/develop/src/lib/style/components/_all.scss)

src/lib/style/components/query_filter/_facet-filter-header.scss
```diff
@@ -0,0 +1,12 @@
+
```
leem42, line 1: There are a few cases where we have two components style sheets in the same scss file, we have facet-filter-header in [_query-filter.scss](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/5db3a0093262cfebc788ad21dbed922c234e735e/src/lib/style/components/_query-filter.scss#L80), but since you now have it in its own file (which is preferable) then I think we should take its scss from _query-filter.scss and put it in this file.

src/lib/containers/widgets/query-filter/FacetFilterHeader.tsx
emmanmills: If the stylesheet is shared with other components, I agree it should go to a higher level css file like _all.scss. In this case, these styles are not being shared and I think we can import the scss file here. The reason is that the styles are only imported when the component is rendered and I believe this is preferable. However, if the facet filter component is pretty much on every page, I will import the file in _all.scss instead. Please let me know.

src/lib/style/components/query_filter/_facet-filter-header.scss
emmanmills: Yes, I created the new /query-filter/_facet-filter-header.scss file to match the existing typescript folder structure so it's easier to find the component css code. Like you said, it's preferable each component has its own css file unless the styles are shared among different components. I will move the _facet-filter-header specific styles to its own file.

PR #516, opened 2020-05-28
PORTALS-1304: Add search bar animation

emmanmills:


src/lib/containers/SearchV2.tsx
```diff
@@ -135,93 +136,101 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
...
+          </div>
+        </CSSTransition>
+        <div className="SearchV2__dropdown_pos">
+          <CSSTransition
+          in={show}
+          classNames="SearchV2__animate_dropdown"
+          timeout={{ enter: 0, exit: 300 }}
+          >
+
+              <form ref={this.radioFormRef} className="SearchV2__column-select SearchV2__animate_height">
```
leem42, line 200: @emmanmills if you could add `.vscode/settings.json` to your workspace with the following contents `settings.json` then you should get auto formatting.
```
{
  "editor.formatOnSave": true,
  "eslint.alwaysShowStatus": true
}
```

leem42:
@emmanmills there is a merge conflict that need's to get resolved

src/lib/containers/SearchV2.tsx
emmanmills: Cool! Done.

PR #526, opened 2020-06-02
Portals 1326: add tooltips to plot control icons

emmanmills:


src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -380,7 +381,14 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = ({
   const renderChartSelectionToggle = (): JSX.Element => (
     <Dropdown>
       <Dropdown.Toggle variant="light" id="plot-selector">
-        <FontAwesomeIcon icon={faChartBar} title="expand" />
+        <ElementWithTooltip
+                idForToolTip="toggleChart"
```
leem42, line 385: It seems like the formatting is not quite right, have you been able to get that working? I only ask because we'll end up with artificial merge differences if we have different formatting, if I check in this file and save it then we could have spaces in the PR when no code was changed.

leem42:
shoot, I completely missed that the tests failed in this PR, I have since fixed them in this one https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/527

PR #549, opened 2020-06-23
Explore 2.0 Bug Fixes

leem42:
PORTALS-1309 - Fix tooltips
PORTALS-1338 - Show More Button Alignment
PORTALS-1417 - Range Facet filter doesn't collapse
PORTALS-1301 - Add show less button to facet dropdown

Removed sorting on tables for entity, filehandle id, and user id


src/lib/containers/widgets/query-filter/RangeFacetFilter.tsx
```diff
@@ -80,51 +80,52 @@ export const RangeFacetFilter: React.FunctionComponent<RangeFacetFilterProps> =
         onClick={(isCollapsed: boolean) => setIsCollapsed(isCollapsed)}
         facetAliases={facetAliases}
       ></FacetFilterHeader>
+      <div style={{ display: isCollapsed ? 'none' : 'block' }}>
```
emmanmills, line 83: Did you make any changes in this file or just updating the spacing?

leem42: this was the only line that I actually changed, everything else was spacing.

PR #550, opened 2020-06-23
PORTALS-1459

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -158,17 +169,30 @@ export default class SynapseTable extends React.Component<
...
       isFetchingEntityVersion: false,
+      tableElementId: uniqueId()
     }
     this.getEntityHeadersInData = this.getEntityHeadersInData.bind(this)
   }
+  
+  // instance variables
+  resizer:any
+  
+  componentWillUpdate() {
```
leem42, line 180: it looks like this is a deprecated method, do you know what would happen if this method was removed?

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -393,6 +417,31 @@ export default class SynapseTable extends React.Component<
     )
   }
 
+
+  enableResize() {
+    if (!this.resizer) {
+      if (ReactDOM) {
+        const domNode: Text | Element | null = ReactDOM!.findDOMNode(this)
```
leem42, line 424: when I pulled down the PR eslint gave a warning 'dont use findDOMNode' it looks like there is a more preferable method using [refs instead](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-find-dom-node.md)

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -158,17 +169,30 @@ export default class SynapseTable extends React.Component<
       mapUserIdToHeader: {},
       isFetchingEntityHeaders: false,
       isFetchingEntityVersion: false,
+      tableElementId: uniqueId()
     }
     this.getEntityHeadersInData = this.getEntityHeadersInData.bind(this)
   }
+  
+  // instance variables
+  resizer:any
```
leem42, line 178: there are times when we have two tables on the same page, does have the variable declared like this pose any problem? I think it would be fine right?

src/lib/containers/table/SynapseTable.tsx
leem42: if we do need this method then it looks like we have to use [UNSAFE_componentWillUpdate](https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate)

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: should be ok (if the variable really is associated to the instance!).

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: I'll try it out, thanks!

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: I'll take a look.

jay-hodgson:
I steered clear of trying to use ref because of this post: https://stackoverflow.com/questions/51565904/how-to-make-column-resizer-work-in-react
But it works just fine, which means I get to clean up the code.  Thanks @leem42 !

PR #560, opened 2020-06-26
Wait to merge - PORTALS-1293

leem42:
Uses new conditions API

**BLOCKED**: The backend needs to respect this new parameter in this call https://docs.synapse.org/rest/POST/table/sql/transform.html otherwise the ProgrammaticOptions modal will show broken sql.


jay-hodgson:
@leem42 : Could you resolve conflicts and review with @emmanmills ?  
I think we should take the known Programmatic Download Modal bug for now in order to get the multi-value search (since there's a clear path forward for fixing the bug).  The temporary consequence is that people may be downloading more files than they would expect if they use the ProgrammaticOptions instructions, until the backend call is enhanced and we feed in the QueryFilters).

PR #573, opened 2020-07-07
PORTALS-1459: now functional on Safari, with slight updates to style based on design.

jay-hodgson:
https://www.figma.com/proto/DTsx5JIon6awus9Ers9jTuvL/Tables-2.0?node-id=470%3A3900&scaling=min-zoom


jay-hodgson:
Apparently that "componentWillUpdate" was an important step for it to work in Safari (see react example in https://www.npmjs.com/package/column-resizer).  Adding shouldComponentUpdate() with the disableResize made is so I could no longer repro the problem on Safari.

PR #577, opened 2020-07-08
fix filters in graphs

leem42:
🤞 
https://sagebionetworks.jira.com/browse/PORTALS-1493
this pr fixes broken facet filtering from the dropdown
also a small ux tweak which is that the user should not be able to turn the all checkbox off since
that is a no-op

### DIFF
--- a/src/lib/containers/widgets/Checkbox.tsx
+++ b/src/lib/containers/widgets/Checkbox.tsx
@@ -6,12 +6,13 @@ export type CheckboxProps = {
   checked?: boolean
   className?: string
   onChange: Function
+  isSelectAll?: boolean
 }
 
 export const Checkbox: React.FunctionComponent<CheckboxProps> = (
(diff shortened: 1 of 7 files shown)


src/lib/containers/widgets/Checkbox.tsx
```diff
@@ -6,21 +6,30 @@ export type CheckboxProps = {
...
   const handleCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setChecked(event.target.checked)
     props.onChange(event.target.checked)
+    if (isSelectAll && event.target.checked === false) {
+      /* 
+         You can click the all checkbox from off -> on
+         but clicking it off is a no-op
+      */
+      setChecked(true)
+      return
```
jay-hodgson, line 30: Consider using an 'else' block in favor of the early return

PR #606, opened 2020-07-29
PORTALS-1478: update to the latest plotly.js-basic-dist version

jay-hodgson:
from 1.53.0 to 1.54.7


jay-hodgson:
Nevermind, finally ran into the issue with the latest version.

PR #607, opened 2020-07-29
PORTALS-1478 : recreate Plot on resize

jay-hodgson:


src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -471,11 +476,11 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = ({
             {({ size }) => (
               <div className={getClassNameForPlotDiv(isExpanded, plotType)}>
                 <Plot
+                  key={`${facetToPlot.columnName}-${size.width}`}
                   layout={layout}
                   data={plotData?.data ?? []}
-                  style={getPlotStyle(size.width, plotType)}
-                  config={{ displayModeBar: false, responsive: true }}
```
leem42, line 477: Did removing `responsive: true` have any particular effect?

jay-hodgson: No, but I want to reduce any extra work that they might be doing within their component!

PR #608, opened 2020-07-29
PORTALS-1526

leem42:


src/lib/containers/ExternalFileHandleLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+        console.error('Error on getting external file handle = ', e)
+      }
+    }
+    getEntity()
+  }, [synId, token])
+
+  return file ? (
+    <a href={file.externalURL} className={className} target="_blank">
+      <span>
+        {file?.fileName}
```
jay-hodgson, line 57: Really want the FileHandle fileName in this case, or the FileEntity name?

leem42: I didn't realize there was a distinction between the two, it sounds like using the FileEntity name would make more sense?

jay-hodgson: Yes please (the file name might be gibberish, but the user can set a meaningful entity name).

PR #614, opened 2020-07-31
Follow up on PORTALS-1526

leem42:
added error component to markdownsynapse

### DIFF
--- a/src/lib/containers/Error.tsx
+++ b/src/lib/containers/Error.tsx
@@ -31,9 +31,7 @@ export const Error = (props: ErrorProps) => {
               Please <SignInButton /> to view this resource.
             </>
           )}
-          {accessDenied && (
-            <>You are not authorized to access the page requested.</>
-          )}
+          {accessDenied && <>You are not authorized to access this resource.</>}
           {
             // if we don't have a friendly error message then show the error
             !accessDenied && !loginError && <>{error.reason}</>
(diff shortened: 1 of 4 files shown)


src/lib/containers/Error.tsx
```diff
@@ -21,12 +25,20 @@ export const Error = (props: ErrorProps) => {
...
-        )}
-        <p>{error.reason}</p>
+        <p>
+          {loginError && (
+            <>
+              Please <SignInButton /> to view this resource.
+            </>
+          )}
+          {accessDenied && (
+            <>You are not authorized to access the page requested.</>
```
jay-hodgson, line 35: Is it always going to be a "page"?

src/lib/containers/Error.tsx
```diff
@@ -21,12 +25,20 @@ export const Error = (props: ErrorProps) => {
...
-        {error.status === 403 && (
-          <p>
-            Please <SignInButton /> to view this resource.
-          </p>
-        )}
-        <p>{error.reason}</p>
+        <p>
+          {loginError && (
+            <>
+              Please <SignInButton /> to view this resource.
```
jay-hodgson, line 31: Will `<SignInButton />` cause re-mounting problems?

leem42: It shouldn't, and if it did re-mount a bunch of extra times it would be less of an issue since it doesn't make any API calls, but thank you for pointing out!

src/lib/containers/Error.tsx
leem42: It would not.. maybe instead - "You are not authorized to access this resource"?

leem42: updated the text

PR #615, opened 2020-08-10
PORTALS-1522

jay-hodgson:
Initial integration of [Upset.js](https://medium.com/@sgratzl/upset-js-the-upset-js-ecosystem-ab6692d2f34a).
Styled using portal color palette, currently with the ability to set the labels and max column count via props (we may want to expose others).  It assumes that the given sql results in 2 columns that drive the viz.

Screen Shot 2020-08-12 at 11 51 30 AM


jay-hodgson:
On hover I'm changing the selection (and changed the selection style opacity so you can see it reflected in the other bars).

Screen Shot 2020-08-10 at 4 39 58 PM

jay-hodgson:
It takes time to pull down all of the pages of the table, so this definitely needs to be below the fold.

src/lib/containers/UpsetPlot.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return () => {
+      isCancelled = true
+    }
+  }, [entityId, sql, token])
+
+  return (
+    <>
+      {isLoading && loadingScreen}
+      {!isLoading && sets && combinations &&
+        <SizeMe>
```
conundrumgirl, line 109: I wonder if the same trick you used to fix plotly resizing would work here with listening to the resizing event and changing the key and thus forcing redraw.

src/lib/containers/UpsetPlot.tsx
```diff
@@ -0,0 +1,132 @@
...
+  height = 700,
+  loadingScreen,
+  token,
+}) => {
+  const [isLoading, setIsLoading] = useState<boolean>()
+  const [sets, setSets] = useState<any>()
+  const [combinations, setCombinations] = useState<any>()
+  const [selection, setSelection] = React.useState(null as ISetLike<any> | null)
+
+  const { colorPalette } = getColorPallette(0, 1)
```
leem42, line 38: I think we'll need a prop for this, unless we can always use this color which would be great!

src/lib/containers/UpsetPlot.tsx
```diff
@@ -0,0 +1,132 @@
...
+        token
+      )
+      // transform query data into plot data, and store.
+      // collect all values for each key
+      const keyValuesMap = {}
+      // keyValuesMap looks like { 'A': {name: 'A', sets: ['S1', 'S2'] }, 'B': { name: 'B', sets: ['S1'] }, ... }
+      // It's a little redudant, but makes the next step much easier.
+      for (const row of queryResult.queryResults.rows) {
+        for (let j = 1; j < row.values.length; j += 1) {
+          const rowValues: any = row.values
```
conundrumgirl, line 70: I'm questioning the explicitly setting the 'any' type. Would it make sense to strong type it?

src/lib/containers/UpsetPlot.tsx
```diff
@@ -0,0 +1,132 @@
...
+      // elems looks like [{ name: 'A', sets: ['S1', 'S2'] }, { name: 'B', sets: ['S1'] }, ...]
+      const sets = extractSets(elems);
+      const combinations = generateCombinations(sets, {
+        type: 'intersection',
+        min: 1,
+        limit: maxBarCount,
+        order: 'cardinality:desc',
+      })
+      if (!isCancelled) {
+        setSets(sets)
```
leem42, line 94: For future reference, since the `sets` and `combinations` variables are set at the same time they could also be put into a single state variable, like so
`const [data, setData] = useState<{sets, combinations }: {sets: null as ISetLike<any> | null, combinations: any} >()`

The benefit of the approach above ^ would be that it would reduce the number of render calls (one versus two) and amount of child component updates. And just to be clear, the only reason I'm saying those two variable could be put into one is because they both get updated at the same time.

src/demo/containers/playground/UpsetPlotDemo.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import UpsetPlot from 'lib/containers/UpsetPlot'
+
+export const UpsetPlotDemo = () => {
+  return (
+    <UpsetPlot
+      entityId='syn20821313'
```
leem42, line 7: I'm trying to revert course on the decision to use the entityId everywhere, if we could change this to grab the entityId from the sql instead.

src/lib/containers/UpsetPlot.tsx
jay-hodgson: Dynamic resizing started to work, not sure what changed!

src/lib/containers/UpsetPlot.tsx
jay-hodgson: done

src/lib/containers/UpsetPlot.tsx
jay-hodgson: updated to real type (string[])

src/lib/containers/UpsetPlot.tsx
jay-hodgson: updated to a single state variable (of new UpsetPlotData type)

src/demo/containers/playground/UpsetPlotDemo.tsx
jay-hodgson: parse entity id

PR #620, opened 2020-08-13
PORTALS-1522: improve error handling, and check for null annotation value

jay-hodgson:
(note that this particular error only shows up on staging psychencode if you are logged in as a user who has access to all files in the project.

### DIFF
--- a/src/lib/containers/UpsetPlot.tsx
+++ b/src/lib/containers/UpsetPlot.tsx
@@ -5,6 +5,7 @@ import { SynapseConstants, SynapseClient } from '../utils'
 import { SizeMe } from 'react-sizeme'
 import getColorPallette from './ColorGradient'
 import { parseEntityIdFromSqlStatement } from '../utils/functions/sqlFunctions'
+import { Alert } from 'react-bootstrap'
 
 export type UpsetPlotProps = {
   sql: string, // first column should contain values, second column should contain a single set value.  ie. SELECT distinct individualID, assay FROM syn20821313
@@ -37,6 +38,7 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
 }) => {
   const [isLoading, setIsLoading] = useState<boolean>()
   const [data, setData] = useState<UpsetPlotData>()
+  const [error, setError] = useState<string>()
   const [selection, setSelection] = useState(null as ISetLike<any> | null)
 
   const { colorPalette } = getColorPallette(rgbIndex, 2)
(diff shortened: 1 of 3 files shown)


src/lib/containers/UpsetPlot.tsx
```diff
@@ -48,65 +50,72 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
...
-          keyValuesMap[key].sets = keyValuesMap[key].sets || []
-          const found = keyValuesMap[key].sets.find((v: any) => v == newValue)
-          if (!found) {
-            if (caseInsensitiveSetNames[newValue.toUpperCase()]) {
-              newValue = caseInsensitiveSetNames[newValue.toUpperCase()]
-            } else {
-              caseInsensitiveSetNames[newValue.toUpperCase()] = newValue
+            keyValuesMap[key].sets = keyValuesMap[key].sets || []
+            const found = keyValuesMap[key].sets.find((v: any) => v == newValue)
+            if (!found && newValue) {
```
jay-hodgson, line 89: This is the bug fix (to verify newValue is defined before pushing it into the set)

src/lib/containers/UpsetPlot.tsx
```diff
@@ -48,65 +50,72 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
     let isCancelled: boolean = false
     const getPlotData = async () => {
       setIsLoading(true)
-      const entityId = parseEntityIdFromSqlStatement(sql)
```
jay-hodgson, line 51: This code was wrapped with a try/catch.

src/lib/containers/UpsetPlot.tsx
```diff
@@ -37,6 +38,7 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
 }) => {
   const [isLoading, setIsLoading] = useState<boolean>()
   const [data, setData] = useState<UpsetPlotData>()
+  const [error, setError] = useState<string>()
```
jay-hodgson, line 41: Show a "danger" bootstrap alert if any error occurs when we get the plot data and transform.

src/lib/containers/UpsetPlot.tsx
```diff
@@ -142,6 +151,15 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
           )}
         </SizeMe>
       }
+      {error &&
```
leem42, line 154: you can use our [Error](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/ed2978c0e238c8e6d17e203bf3dc39ad66813b8e/src/lib/containers/Error.tsx) component instead of the boostrap alert!

jay-hodgson: Saw that, but it looks like that one is currently specific to SynapseClientError.  If it was modified to be more generic (in the current case I run into a TypeError), then that's a great option!

leem42: ah right, in the future I can add an `errorString` prop and `backendError` or something like that to distinguish the two

jay-hodgson: Alternatively, you could pass in a single error object, and the component examines it to figure out what information it has (then in my case I could pass in the err, and it might be a SynapseClientError or a TypeError!).

PR #621, opened 2020-08-13
PORTALS-1552

leem42:
- bump react-scripts to fix dependabot alerts
- extended the Error component to detect the error coming in, put in the UpSet plot
- Added goals component

preview -
image

zoomed in preview -
image

### DIFF
--- a/src/lib/containers/Error.tsx
+++ b/src/lib/containers/Error.tsx
@@ -4,24 +4,20 @@ import SignInButton from './SignInButton'
 import { Alert } from 'react-bootstrap'
 type ErrorProps = {
   token?: string
-  error?: string | TypeError | SynapseClientError
+  error?: string | Error | SynapseClientError
 }
 
 function isSynapseClientError(
-  error: string | TypeError | SynapseClientError,
+  error: string | Error | SynapseClientError,
 ): error is SynapseClientError {
   return (error as any).status !== undefined
 }
 
-function isTypeError(
-  error: string | TypeError | SynapseClientError,
-): error is TypeError {
+function isError(error: string | Error | SynapseClientError): error is Error {
   return (error as any).message !== undefined
 }
 
-function isString(
-  error: string | TypeError | SynapseClientError,
-): error is string {
+function isString(error: string | Error | SynapseClientError): error is string {
   return typeof error === 'string'
 }
 
@@ -57,12 +53,12 @@ export const Error = (props: ErrorProps) => {
   }
 
   let synapseClientError: SynapseClientError | undefined = undefined
-  let typeError: TypeError | undefined = undefined
+  let jsError: Error | undefined = undefined
   let stringError: string | undefined = undefined
   if (isSynapseClientError(error)) {
     synapseClientError = error
-  } else if (isTypeError(error)) {
-    typeError = error
+  } else if (isError(error)) {
+    jsError = error
   } else if (isString(error)) {
     stringError = error
   }
@@ -78,7 +74,7 @@ export const Error = (props: ErrorProps) => {
           {synapseClientError && (
             <ClientError error={synapseClientError} token={token} />
           )}
-          {typeError && typeError.message}
+          {jsError && jsError.message}
           {stringError && stringError}
         </p>
       </Alert>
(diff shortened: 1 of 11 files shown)


src/lib/containers/Goals.tsx
```diff
@@ -0,0 +1,163 @@
...
+  useEffect(() => {
+    const getData = async () => {
+      const request: QueryBundleRequest = {
+        concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
+        entityId,
+        partMask:
+          SynapseConstants.BUNDLE_MASK_QUERY_COLUMN_MODELS |
+          SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
+        query: {
+          sql: `SELECT * FROM ${entityId}`,
```
jay-hodgson, line 66: I think these need an explicit order (controlled by the portal stakeholder).

src/lib/containers/Goals.tsx
```diff
@@ -0,0 +1,163 @@
...
+          fileId => {
+            return {
+              associateObjectId: entityId!,
+              associateObjectType: FileHandleAssociateType.FileEntity,
+              fileHandleId: fileId,
+            }
+          },
+        )
+        const batchFileRequest: BatchFileRequest = {
+          includeFileHandles: true,
```
jay-hodgson, line 94: Do you need the actual FileHandles?

leem42: nope! good catch

src/lib/containers/Goals.tsx
leem42: You can control the order of the rows right? I created a table and the rows appear to be in insert order?

leem42: the alternative is we take in sql with an orderBy clause

src/lib/containers/Error.tsx
```diff
@@ -14,9 +56,16 @@ export const Error = (props: ErrorProps) => {
     return <></>
   }
 
-  const loginError = error.status === 403 && !token
-  const accessDenied = error.status === 403 && token
-
+  let synapseClientError: SynapseClientError | undefined = undefined
+  let typeError: TypeError | undefined = undefined
```
jay-hodgson, line 60: Could this TypeError be expanded to the Error type, and in any case grab the associated message?  It would be a shame to need special cases for all of the standard error types (URIError, SyntaxError, InternalError, EvalError, ...)

src/lib/containers/Error.tsx
```diff
@@ -3,8 +3,46 @@ import { SynapseClientError } from '../utils/SynapseClient'
...
+function isError(error: string | Error | SynapseClientError): error is Error {
+  return (error as any).message !== undefined
+}
+
+function isString(error: string | Error | SynapseClientError): error is string {
+  return typeof error === 'string'
+}
+
+export const ClientError = (props: {
+  error: SynapseClientError
```
jay-hodgson, line 25: Does this type not need to match the prop?
string | Error | SynapseClientError

jay-hodgson: My misunderstanding.  Resolving.

src/lib/containers/Goals.tsx
```diff
@@ -0,0 +1,163 @@
...
+  BatchFileRequest,
+} from '../utils/synapseTypes'
+import { SynapseClient, SynapseConstants } from '../utils'
+import { SynapseClientError, getFiles } from '../utils/SynapseClient'
+import { Error } from '../containers/Error'
+import QueryCount from './QueryCount'
+import { parseEntityIdFromSqlStatement } from '../utils/functions/sqlFunctions'
+
+export type GoalsProps = {
+  sql: string
```
jay-hodgson, line 16: Another approach would be to just take the tableID in this component, and define the columns explicitly in the (sql baked into this component).  Then you wouldn't need to do any client-side validation, since the query execution would fail if any of the expected columns were missing.  `select TableId, Title, Summary, Link, Asset from ${entityId} order by ItemOrder`

leem42: ah, I like it when synapse does the work! I will change it to that

PR #646, opened 2020-08-25
Wait to merge: PORTALS-1577

leem42:
# BLOCKED 
Wait to merge once psychencode is released, this PR has a lot of changes which poses a risk to release.

This is the first step in creating components for mobile, contains mobile Resources Component view.

## Demo Page
avoid rendering the demo page until the token is loaded, this avoids an extra render that happens for all the components when the token gets updated, it helps record keeping with API calls. (this is also how its done in the portals)

## Put `Goals.tsx` And `Resources.tsx` in a new holder `home_page`
Proposed naming for components with mobile and desktop views could be `Component.Desktop.tsx` and `Component.Mobile.tsx`

## New hook `useGetQueryResultBundle`
- created new hook to get data from synapse: it exposes queryResult, isLoading, and error
- used in both Goals and Resources components

## New hook `useShowDesktop`
- lets component track whether it should render mobile or laptop view

## New component `ExpandableContent`
- used for mobile rendering home page modules
image

## Mobile Version of Goals Preview
image

## Mobile Version of Resources Preview
image


jay-hodgson:
I think we should move forward with these changes.  @emmanmills , could you review these changes with @leem42 ?

If absolutely necessary, we can go back to a commit before this and create a branch to patch psychencode.

src/lib/containers/home_page/goals/Goals.Desktop.tsx
```diff
@@ -0,0 +1,41 @@
...
+              name=""
+            />
+          </span>
+        </p>
+      </div>
+      <div className="Goals__Card__summary">
+        <p> {summary} </p>
+        <p>
+          <a className="Goals__Card__summary__link" href={link}>
+            EXPLORE
```
leem42, line 35: add prop for name

PR #650, opened 2020-08-25
PORTALS-1557: Multiple css and markup updates

emmanmills:
1. Change content scrolling to auto
2. Add background color for component div
3. Add loading bar
4. Add explore all user link

Screen Shot 2020-08-25 at 9 55 43 AM

### DIFF
--- a/src/lib/containers/UserCardListRotate.tsx
+++ b/src/lib/containers/UserCardListRotate.tsx
@@ -13,6 +13,7 @@ export type UserCardListRotateProps = {
   count: number
   token?: string
   loadingScreen: JSX.Element
+  summaryLink: string
 }
 
 export const getDisplayIds = (ids: string[] = [], count: number = DEFAULT_DISPLAY_COUNT) => {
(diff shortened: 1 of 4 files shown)


src/lib/containers/UserCardListRotate.tsx
```diff
@@ -87,7 +93,15 @@ const UserCardListRotate: React.FunctionComponent<UserCardListRotateProps> = ({s
 
   return (
     <div className="UserCardListRotate">
-      {userIds.length && <UserCardList list={userIds} size={LARGE_USER_CARD} />}
+      {isLoading && loadingScreen}
+      {!isLoading && userIds.length && <UserCardList list={userIds} size={LARGE_USER_CARD} />}
+      <div className="UserCardListRotate__summary">
+        <p>
+          <a className="UserCardListRotate__summary__link btn homepage-button" href="/Explore/People">
```
leem42, line 100: If we could change the href to take in a prop instead. I'm sure that /Explore/People will probably always be the link this goes too, but just so we have the option in the future.

emmanmills: Thanks, this has been updated.

PR #657, opened 2020-08-26
PORTALS-1555: home page buttons are currently part of each component

jay-hodgson:


src/demo/containers/playground/UpsetPlotDemo.tsx
```diff
@@ -9,11 +9,13 @@ export const UpsetPlotDemo: React.FunctionComponent<UpsetPlotDemoProps> = ({toke
   return (
     <UpsetPlot
       token={token}
-      sql={"SELECT distinct individualID, assay FROM syn20821313 where individualID is not null"}
+      sql={"SELECT unnest(individualID), assay FROM syn20821313 WHERE individualID is not null GROUP BY assay, unnest(individualID)"}
```
jay-hodgson, line 12: And fixed sql, now that the underlying table schema has changed.

PR #663, opened 2020-08-27
PORTALS-1597: upset plot style changes

jay-hodgson:


jay-hodgson:
Screen Shot 2020-08-26 at 5 21 06 PM

src/lib/style/components/_upset-plot.scss
```diff
@@ -1,12 +1,45 @@
...
+  
+  // rotate axis label (so that it's vertical), and it move down by 12%
+  text[class^='cChartTextStyle-upset-'], text[class*=' cChartTextStyle-upset-']{
+    text-transform: uppercase;
+  }
+  
+  // rotate axis label (so that it's vertical), and it move down by 12%
+  text[class^='sChartTextStyle-upset-'], text[class*=' sChartTextStyle-upset-']{
+    transform: rotate(270deg) translateX(-12%);
+    text-transform: uppercase;
```
emmanmills, line 27: text-transform doesn't seem to be available for text tag. See https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text. Might want to double check this one. Otherwise, everything else looks okay to me.

jay-hodgson: Nice catch!  It seems to work on Chrome and FF (79), but perhaps this doesn't work on other browser versions.  Changed to txt.toUpperCase() in the js instead.

PR #672, opened 2020-08-28
fix rendering of multi-value content in dataset card

leem42:
before
image

after
image


jay-hodgson:
https://sagebionetworks.jira.com/browse/PORTALS-1429

PR #673, opened 2020-08-28
PORTALS-1525: add allow-list (for tags/categories), and make tags link to site

jay-hodgson:


src/lib/containers/RssFeedCards.tsx
```diff
@@ -103,12 +102,20 @@ export default class RssFeedCards extends React.Component<RssFeedCardsProps, Rss
...
                     <div className="RssFeedItemCategories">
-                    {item['categories'].map((categoryName: any, index: any) => {
-                      if (index >= MAX_CATEGORIES)
-                        return <></>
-                      // else
-                      return <div key={`${item.guid}_${categoryName}`} className="RssFeedItemCategory">{categoryName}</div>
-                    })}
+                      {item['categories'].map((categoryName: any,) => {
+                        // are we allowed to show this category/tag?
+                        const categoryNameLowerCase = categoryName.toLowerCase()
```
leem42, line 107: it looks like categoryName could be of type string instead of any?

jay-hodgson: nice catch, it certainly is.  updated.

PR #678, opened 2020-09-02
PORTALS-1619:  improve error handling

jay-hodgson:
…esponse sent in async job response

### DIFF
--- a/src/lib/containers/UpsetPlot.tsx
+++ b/src/lib/containers/UpsetPlot.tsx
@@ -123,7 +123,11 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
       } catch (err) {
         if (!isCancelled) {
           setIsLoading(false)
-          setError(err.toString())
+          let errorMessage = err['reason']
+          if (!errorMessage) {
+            errorMessage = err.toString()
+          }
+          setError(errorMessage)
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/lib/containers/UpsetPlot.tsx
```diff
@@ -123,7 +123,11 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
       } catch (err) {
         if (!isCancelled) {
           setIsLoading(false)
-          setError(err.toString())
+          let errorMessage = err['reason']
+          if (!errorMessage) {
+            errorMessage = err.toString()
+          }
```
emmanmills, line 129: Just curious, is err a JSON object and you will cast it to string?

jay-hodgson: Not sure what err is.  If it's an error that resulted from getFullQueryTableResults() async call, then it will very likely be a JSON object (that contains the "reason" key).  Is there a better approach to this (unknown runtime error handling)?

emmanmills: Ah, I see. I would probably just hardcode some general error message like "Error retrieving data" if the message is user facing. At the same time I would console.log the error object for inspection.

emmanmills (approved):
Looks good to me!

PR #693, opened 2020-09-14
PORTALS-960

jay-hodgson:


src/__tests__/lib/containers/QueryWrapperMenu.test.tsx
```diff
@@ -42,7 +37,6 @@ describe('it renders with basic functionality', () => {
     menuConfig: [{ sql, facet }],
     rgbIndex: 3,
     stackedBarChartConfiguration: {
```
emmanmills, line 39: Question, do we always need to pass an empty object as a placeholder even if there is nothing in it? Thanks.

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -192,9 +190,8 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
   } else if (isLoadingNewData) {
     return (
       <div className="SRC-loadingContainer SRC-centerContentColumn">
-        {loadingScreen}
         {asyncJobStatus?.progressMessage && (
-          <div>{asyncJobStatus.progressMessage} </div>
+          <div> <span className="spinner" /> {asyncJobStatus.progressMessage} </div>
```
emmanmills, line 194: Is this a different implementation for the loading state because it has a custom message? Thanks.

emmanmills (commented):
Just a couple questions, otherwise looks good to me!

src/__tests__/lib/containers/QueryWrapperMenu.test.tsx
jay-hodgson: good question!  maybe these can be removed, I'll take a look...

src/lib/containers/widgets/facet-nav/FacetNav.tsx
jay-hodgson: This could be a loadingScreen too (with the async message), I just had a spinner since the Table will always have the loading bar (so I thought 2 of these bars right next to each other might look strange, if this loader is ever shown!).  I'd be happy to change if you think a bar would be preferred.

emmanmills: That's okay. I think it makes sense to have a spinner. Thanks.

src/__tests__/lib/containers/QueryWrapperMenu.test.tsx
jay-hodgson: removed the empty stacked bar chart configurations!

jay-hodgson: but missed the test, hold on :)

jay-hodgson: Apparently, if there is no stacked bar chart configuration then the stacked bar chart is not shown today.  And if it's provided (even if it's empty), it will be shown.  So, for now these need to be there.

emmanmills: Got it. Sounds good.

emmanmills (approved):
Looks good to me!

PR #696, opened 2020-09-17
PORTALS-1664: set width of generic card icons (take up space they are given)

jay-hodgson:


jay-hodgson:
This would effect amp-ad icons in a bad way, closing.

PR #700, opened 2020-09-17
PORTALS-1660

jay-hodgson:


src/demo/containers/playground/PeopleProfileDemo.tsx
```diff
@@ -13,6 +14,16 @@ export const PeopleProfileDemo = () => {
...
+          <UserCardListRotate
+            sql={"SELECT * FROM syn13897207 where isFeatured=true"}
+            count={3}
+            useQueryResultUserData={true}
+            size={MEDIUM_USER_CARD}
+            summaryLinkText={"EXPLORE ALL PEOPLE"}
+            summaryLink={"/Explore/People"}            
+          />
+        </div>
       </div>
```
jay-hodgson, line 27: output:
Screen Shot 2020-09-17 at 3 57 04 PM

src/demo/containers/playground/PeopleProfileDemo.tsx
```diff
@@ -13,6 +14,16 @@ export const PeopleProfileDemo = () => {
...
           />
         </div>
+        <div className="row">
+          <UserCardListRotate
+            sql={"SELECT * FROM syn13897207 where isFeatured=true"}
+            count={3}
+            useQueryResultUserData={true}
+            size={MEDIUM_USER_CARD}
+            summaryLinkText={"EXPLORE ALL PEOPLE"}
+            summaryLink={"/Explore/People"}
```
jay-hodgson, line 24: note that you can now set the UserCard size (defaults to LARGE), and can tell it to use the query results to populate (something that the UserCardList already supports, but the UserCardListRotate needs to support for AMP-AD since not all rows contain a UserID!)

src/lib/containers/UserCardList.tsx
```diff
@@ -96,7 +96,7 @@ export default class UserCardList extends React.Component<
         lastName: values[lastNameIndex],
         company: values[institutionIndex],
         ownerId: '',
-        userName: values[firstNameIndex][0],
+        userName: '',
```
jay-hodgson, line 99: This is a bug that I need to fix.  Now I realize this is to support the default picture (first letter of the username), but it needs to handle when the first name is not set.

src/lib/containers/UserCardListRotate.tsx
```diff
@@ -48,12 +50,15 @@ const UserCardListRotate: React.FunctionComponent<UserCardListRotateProps> = ({
     sql,
     count,
     token,
+    useQueryResultUserData = false,
+    size = 'LARGE USER CARD',
```
emmanmills, line 54: Do you think we should use the constant `LARGE_USER_CARD` from 'lib/utils/SynapseConstants'?

src/demo/containers/playground/PeopleProfileDemo.tsx
emmanmills: Looking at the small cards, it seems that the background is transparent so the pink color becomes the card's background. I think it should be white instead.

emmanmills (changes requested):
Just noticed a couple things. Otherwise, looks good to me.

src/demo/containers/playground/PeopleProfileDemo.tsx
jay-hodgson: Good catch, I'll take a look...

jay-hodgson: Updated!  
Screen Shot 2020-09-17 at 4 37 48 PM

src/lib/containers/UserCardListRotate.tsx
jay-hodgson: Yes

emmanmills (approved):
Looks good!

PR #701, opened 2020-09-18
PORTALS-1667, PORTALS-1671

jay-hodgson:


src/lib/utils/functions/sqlFunctions.ts
```diff
@@ -78,10 +101,7 @@ export const insertConditionsFromSearchParams = (
   const tokens: string[][] = lexer.tokenize(sql)
   // we want to either create a where clause or insert into the where clause
   const foundIndex = tokens.findIndex(el => el[0] === 'WHERE')
-  const whereClauseIndex =
-    foundIndex === -1
-      ? tokens.findIndex(el => el[0] === 'FROM') + 2
```
jay-hodgson, line 83: This is one bug.  We can't assume that we should insert the WHERE token 2 tokens after the FROM token (the DOT version notation messes this up).

src/lib/containers/SearchV2.tsx
```diff
@@ -123,11 +123,12 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
     event.preventDefault()
     const { searchText } = this.state
     let { columnName } = this.state
+    const { searchable } = this.props
     if (columnName === '') {
       // default to the first one, will always be defined
       columnName =
         this.props.data?.columnModels?.filter(el =>
-          this.isSupportedColumnAndInProps(el),
```
jay-hodgson, line 130: This was a bug that I introduced.  Now isSupportedColumnAndInProps means include it if it's "searchable" (listed in the props as a search column), or (if no search columns were defined) if it supports search.  isSupportedColumn() now only tells you if it's a column type that supports search.

emmanmills: Got it.

src/lib/containers/SearchV2.tsx
```diff
@@ -149,7 +150,7 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
       const columnSingleValueQueryFilter: ColumnSingleValueQueryFilter = {
         columnName,
         operator: ColumnSingleValueFilterOperator.LIKE,
-        values: [searchText],
+        values: [`%${searchText}%`],
```
jay-hodgson, line 153: This is another bug.  The service expects the wildcard characters to get the response that we expect from the LIKE condition(s)

src/lib/utils/functions/sqlFunctions.ts
emmanmills: This is interesting. Good catch!

emmanmills (approved):
Looks good to me. Thank you.

PR #702, opened 2020-09-21
PORTALS-1668

jay-hodgson:


src/lib/containers/home_page/goals/Goals.tsx
```diff
@@ -144,6 +144,7 @@ export default function (props: GoalsProps) {
           summary,
           link,
           asset,
+          token,
```
jay-hodgson, line 147: This is the main bug fix, the rest is to convert old component to use react hooks instead.

src/lib/containers/QueryCount.tsx
```diff
@@ -14,82 +14,71 @@ export type QueryCountProps = {
...
+          entityId,
+          partMask: SynapseConstants.BUNDLE_MASK_QUERY_COUNT,
+        }
+        const newIsCalculatingQueryCountForSql = {
+          ...isCalculatingQueryCountForSql,  
+        }
+        newIsCalculatingQueryCountForSql[`${sql}-${token}`] = true
+        setIsCalculatingQueryCountForSql(newIsCalculatingQueryCountForSql)
+        debugger
+        SynapseClient.getQueryTableResults(request, token).then(data => {
```
emmanmills, line 56: We should delete the debugger statement.

src/lib/containers/QueryCount.tsx
```diff
@@ -14,82 +14,71 @@ export type QueryCountProps = {
...
+          debugger
```
emmanmills, line 57: Same here, we should remove the debugger statement.

src/lib/containers/home_page/goals/Goals.tsx
emmanmills: Got it! Thanks.

emmanmills (changes requested):
Looks good, just couple small changes. Thank you!

src/lib/containers/QueryCount.tsx
jay-hodgson: Oops, removed.

src/lib/containers/QueryCount.tsx
jay-hodgson: Also removed.

emmanmills (approved):
Looks good!

PR #706, opened 2020-09-26
PORTALS-1658

jay-hodgson:


emmanmills (commented):
A few comments about the css !important rule. It would be better if we can avoid using it. otherwise looks good.

jay-hodgson:
@emmanmills 
> A few comments about the css !important rule. It would be better if we can avoid using it. otherwise looks good.

Thank you, I completely agree (this css started as a copy of the Goals component css, which suffers from all of the same issues).  I've cleaned up the css.  I noticed a few other issues today around home page styling (buttons should be the same style, program cards should be horizontally centered, and the featured data button type text should be capitalized).

emmanmills (approved):
Looks good to me.

PR #709, opened 2020-09-28
PORTALS-1659, PORTALS-1658 followup

jay-hodgson:


emmanmills (approved):
Looks good to me. Thanks!

PR #712, opened 2020-09-30
PORTALS-1666: add facet lock

emmanmills:
1. Add a prop to lock a facet name and all its values on details page
2. If the locked facet name is set, it will be hidden from 
-- facet filter 
-- the pie charts
-- search bar and the select dropdown
3. In the data table, no filter for that facet name column

Screen Shot 2020-09-30 at 1 32 10 PM


jay-hodgson (approved):
Clean approach, well done.

PR #713, opened 2020-09-30
set bootstrap 4 as required bootstrap version (PORTALS-1680)

zimingd:


public/index.html
```diff
@@ -21,11 +21,9 @@
       crossorigin="anonymous"
     ></script>
     <!-- Latest compiled and minified JavaScript -->
-    <script
-      src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"
-      integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa"
-      crossorigin="anonymous"
-    ></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj" crossorigin="anonymous"></script>
```
zimingd, line 24: https://getbootstrap.com/docs/4.5/getting-started/introduction/#js

zimingd: @jay-hodgson as a side note, https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/develop/src/demo/SingleFileBuild/index.html uses
```
  <script
    src="https://unpkg.com/react-bootstrap@next/dist/react-bootstrap.min.js"
    crossorigin
  ></script>
```
I'm not sure if this needs to be changed or not.

jay-hodgson:
Note to reader, build was already lock into using bootstrap 4.5.2 and and react-bootstrap 1.3.0 js (see lock file), but this makes it official in the package json.

public/index.html
```diff
@@ -7,8 +7,8 @@
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
     <link
       rel="stylesheet"
-      href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
-      integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u"
+      href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css"
```
jay-hodgson, line 10: These css changes are drastic, and the components look terrible inside of the updated css.  Please run "yarn start" to start up the demo locally (and compare to https://sage-bionetworks.github.io/Synapse-React-Client/ ).  The modal width seems to also be reduced (causing scroll bars and content issues), as well as using rem instead of px for all fonts (generally a great idea, but we're not ready to promote using 4.5.2 css until we make a lot of overrides).

public/index.html
jay-hodgson: Yes, I think so.  It currently redirects to 1.1.0-rc.0:
https://unpkg.com/react-bootstrap@1.1.0-rc.0/dist/react-bootstrap.min.js

It should probably be replaced with 1.3.0:
https://unpkg.com/react-bootstrap@1.3.0/dist/react-bootstrap.min.js

package.json
```diff
@@ -34,7 +34,7 @@
     "@types/spark-md5": "^3.0.1",
     "@types/xml2js": "^0.4.4",
     "@upsetjs/react": "^1.6.2",
-    "bootstrap": ">=3.4.1",
+    "bootstrap": "^4.5.2",
```
jay-hodgson, line 37: Is this needed as a dependency now that we use react-bootstrap?  I tried taking it out, and didn't see any issues with the build or demo output.

jay-hodgson (changes requested):
bootstrap 4 css breaks output.  And from a cursory test, it looks like we don't need bootstrap as a dependency in SRC now.

zimingd:
Closing in favor of using a className wrapper for bootstrap4 styled code. We will slowly transition to bootstrap 4 instead

PR #716, opened 2020-10-01
PORTALS-1678: option to wrap the matching value with parens

jay-hodgson:
This is to help with the AMP-AD study name match. Ideally these would be synapse IDs


emmanmills (approved):
Looks good to me.

PR #718, opened 2020-10-06
PORTALS-1671: simplify search column config

jay-hodgson:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synapse-react-client",
-  "version": "1.15.27",
+  "version": "1.15.28",
```
emmanmills, line 3: I just merged some icon changes and updated the version to 1.15.28. Not sure if this is going to cause conflict here.

emmanmills (approved):
Just a comment in package.json. Otherwise looks good.

PR #720, opened 2020-10-07
PORTALS-1690

jay-hodgson:


src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -38,6 +38,7 @@ export type DownloadConfirmationProps = {
   updateParentState?: <K extends keyof QueryWrapperState>(
     param: Pick<QueryWrapperState, K>,
   ) => void
+  onExportTable?: () => void
```
emmanmills, line 41: Just curious, is this the same as using Function as a type?

emmanmills (approved):
Looks good to me.

src/lib/containers/download_list/DownloadConfirmation.tsx
jay-hodgson: yes, as far as I know!

PR #721, opened 2020-10-08
PORTALS-1647: If searchable column names are defined in the config, grab the first one (that is not locked)

jay-hodgson:


src/lib/containers/SearchV2.tsx
```diff
@@ -124,12 +124,17 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
...
-        )?.[0].name ?? ''
+      if (searchable) {
+        // If searchable column names are defined in the config, grab the first one (that is not locked)
+        columnName = searchable.filter(colName => colName !== lockedFacet?.facet)[0]        
+      } else {
+        // Otherwise, get the first column model that can be searched.
+        // And for study details page: if lockedFacet is defined, remove it from the search
+        const searchableColumnModels = this.props.data?.columnModels?.filter(el => el.name !== lockedFacet?.facet)
+            .filter(el => this.isSupportedColumn(el),
+          )
```
emmanmills, line 135: Just checking, why is there a comma on line 134?

emmanmills (approved):
Just a comment. Otherwise, looks good!

src/lib/containers/SearchV2.tsx
jay-hodgson: For good luck?  ;)  Deleted it

PR #729, opened 2020-10-22
PORTALS-1352: Collapse all but the first 3 facets in QueryFilter

nickgros:
https://sagebionetworks.jira.com/browse/PORTALS-1352

### DIFF
--- a/src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
+++ b/src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
@@ -86,10 +86,10 @@ export const EnumFacetFilter: React.FunctionComponent<EnumFacetFilterProps> = ({
   onChange,
   facetAliases,
   containerAs = 'Collapsible',
-  collapsed,
+  collapsed = false,
 }: EnumFacetFilterProps) => {
   const [isShowAll, setIsShowAll] = useState<boolean>(false)
-  const [isCollapsed, setIsCollapsed] = useState<boolean>(collapsed || false)
+  const [isCollapsed, setIsCollapsed] = useState<boolean>(collapsed)
   const [isShowDropdown, setIsShowDropdown] = useState<boolean>(false)
   const [showSearch, setShowSearch] = useState<boolean>(false)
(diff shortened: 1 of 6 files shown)


src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -85,9 +86,10 @@ export const EnumFacetFilter: React.FunctionComponent<EnumFacetFilterProps> = ({
   onChange,
   facetAliases,
   containerAs = 'Collapsible',
+  collapsed,
```
conundrumgirl, line 89: you can default 'collapse' to false here instead of ln92

src/lib/containers/widgets/query-filter/QueryFilter.tsx
```diff
@@ -163,16 +163,19 @@ export const QueryFilter: React.FunctionComponent<QueryFilterProps> = ({
     <div className="QueryFilter">
       {isLoading && <div>Loading...</div>}
       {!isLoading &&
-        facets.map(facet => {
+        facets.map((facet, index) => {
           const columnModel = columnModels!.find(
             model => model.name === facet.columnName,
           )
 
+          const shouldStartCollapsed: boolean = index > 2;
```
conundrumgirl, line 171: I think type will be inferred here

src/lib/containers/widgets/query-filter/RangeFacetFilter.tsx
```diff
@@ -19,6 +19,7 @@ export type RangeFacetFilterProps = {
   columnModel: ColumnModel
   onChange: Function
   facetAliases: {} | undefined
+  collapsed: boolean
```
conundrumgirl, line 22: just wondering if it's intentional that it's optional in one EnumFilter and not here

conundrumgirl (approved):
a couple of nitpicks but looks good

src/lib/containers/widgets/query-filter/RangeFacetFilter.tsx
nickgros: Not intentional--making both optional for consistency

PR #736, opened 2020-10-27
update to node-sass 4.13.1

jay-hodgson:
auto-update for patch and minor releases.  checking in results after rebuild


emmanmills (approved):
Looks good!

PR #738, opened 2020-10-27
PORTALS-1302:  if unselected segment, set opacity to .25

jay-hodgson:
Note that if no facet values are selected, opacity should be untouched (all facet values are selected in this case).


emmanmills (approved):
Looks good to me!

PR #745, opened 2020-10-29
remove SRC-adjust-for-bootstrap-margin

jay-hodgson:
I think this must have been a misunderstanding.  If items are in a bootstrap responsive grid, then the row already has a -15px margin (left and right) to correct for the columns which have a 15px padding (left and right).


emmanmills (approved):
Looks good! Thanks.

PR #746, opened 2020-10-29
PORTALS-1420: select the search input element on column selection

jay-hodgson:


emmanmills (approved):
Looks good to me!

jay-hodgson:
Emma recommended that we "keep the input onfocus hook in the code base so we can use it later and keep your fix in SearchV2 until the file is converted to functional component" which sounds right to me.

PR #748, opened 2020-11-02
PORTALS-1699

jay-hodgson:
Bootstrap modal dialogs are already centered, and have better responsive behavior than results from the current css overrides.  Delete.


emmanmills (approved):
Looks good! Thanks.

PR #757, opened 2020-11-11
Single file build token

jay-hodgson:


src/lib/containers/home_page/featured-data/FacetPlotsCard.tsx
```diff
@@ -116,7 +116,7 @@ const FacetPlotsCard: React.FunctionComponent<FacetPlotsCardProps> = ({
       </div>
     }
     const isShowingMultiplePlots = facetPlotDataArray.length > 1
-    const cardTitle = title ?? isShowingMultiplePlots ? selectedFacetValue : unCamelCase(facetDataArray[0].columnName, facetAliases)
+    const cardTitle = title ?? (isShowingMultiplePlots ? selectedFacetValue : unCamelCase(facetDataArray[0].columnName, facetAliases))
```
jay-hodgson, line 119: Also noticed that the card title was wrong when a title was provided, because of the order of operations.  Made explicit (that ternary operator statement is the fallback value, if the title is undefined/null)

src/demo/SingleFileBuild/index.html
```diff
@@ -165,7 +178,7 @@
     }
     const queryWrapperMenuContainer = document.querySelector('#panel1')
     const queryWrapperProps = {
-      token: '',
+      token,
```
jay-hodgson, line 181: I think this only worked if people pasted in their session tokens into this code before!

nickgros (approved):

PR #760, opened 2020-11-13
Portals-1749:  nf study details page

emmanmills:
I don't see this issue in other portals but discovered this in NF. I need to disable the filter button in the Synapse table column header if locked facet value is same as the column name.


jay-hodgson (approved):
Surprised this hasn't come up before!  Perhaps other details page tables did not include the locked facet column?  In any case, this change looks right to me.  Thanks Emma.

PR #761, opened 2020-11-13
SWC-5723

nickgros:
* Fixes an issue with `ReactTooltip`s not rendering in synapse.org
* Adds very specific styling to CopyToClipboardInput button (synapse.org overrides #body > .btn)
* Use custom error handling at the top level of the page component, because the added error boundary fails and the entire page fails to render (unsure why, I suspect it has to do with differences between a JS `Error` and a `SynapseClientError`) when the user tries to load the component when not signed in.

### DIFF
--- a/src/lib/containers/CopyToClipboardInput.tsx
+++ b/src/lib/containers/CopyToClipboardInput.tsx
@@ -66,7 +66,7 @@ export const CopyToClipboardInput: React.FunctionComponent<CopyToClipboardInputP
         ></FormControl>
         <Button
           variant="default"
-          className="SRC-copyToClipboardIcon"
+          id="SRC-copyToClipboardButton"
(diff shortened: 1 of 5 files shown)


src/lib/containers/CopyToClipboardInput.tsx
```diff
@@ -66,7 +66,7 @@ export const CopyToClipboardInput: React.FunctionComponent<CopyToClipboardInputP
         ></FormControl>
         <Button
           variant="default"
-          className="SRC-copyToClipboardIcon"
+          id="SRC-copyToClipboardButton"
```
jay-hodgson, line 69: Ok to have multiple elements with the same id on the page (I thought this was discouraged)?  If this is to increase css specificity score, there may be other ways (include element and css class name?).

jay-hodgson (approved):

PR #762, opened 2020-11-13
PORTALS-1757

jay-hodgson:
Add new way to demo components (live editor).  Output available on our gh-pages now:
https://sage-bionetworks.github.io/Synapse-React-Client/
Created md files for home page components, but can be made for any component.


emmanmills (approved):
Looks good.

PR #763, opened 2020-11-14
bump SRC version to 1.15.43

jay-hodgson:


jay-hodgson:
Prep to release 1.15.43

PR #766, opened 2020-11-16
fix editor cursor location (due to font-family override in SRC)

jay-hodgson:


package.json
```diff
@@ -161,8 +161,8 @@
     "build:js": "tsc --p tsconfig.build.json",
     "build:copy-assets": "cd src/lib && find . -name '*.scss' -o -name '*.css' -o -name '*.svg' | cpio -pdm ../../dist/",
     "build:rollup": "rollup -c && cp -r src/umd dist/",
-    "build:docs": "npx styleguidist build",
-    "start-docs": "npx styleguidist server"
+    "build:docs": "yarn run build-css && npx styleguidist build",
```
jay-hodgson, line 164: build css when building docs/demo site or starting local server.

src/demo/style/DemoStyle.scss
```diff
@@ -1,4 +1,4 @@
-$primary-action-color: #c94281;
-$primary-bgcolor-rgba: rgba(201, 66, 129, 0.03);
+$primary-action-color: rgb(64, 123, 160);
+$primary-bgcolor-rgba: rgba(64, 123, 160, 0.03);
```
jay-hodgson, line 2: Change from hotpink to nf's current color scheme from design.

src/lib/style/base/_base.scss
```diff
@@ -21,12 +21,9 @@ html {
...
 }
 
-body * {
-  font-family: 'Lato';
-}
-
 body {
   color: $gray-regular;
+  font-family: 'Lato';
 }
```
jay-hodgson, line 28: This is the main fix.  "body *" is a bit too aggressive a selector, it applied to <pre><span class="token string/operator/punctuation/..."></span></pre>, and changed the font family in codemirror.  The way codemirror works (today) is that it has a textarea (with absolute position) overlaying a pre, where the pre is used to show the actual code (with syntax highlighting, etc).  Once we mucked with the style, these two were no longer in sync, and the cursor was no longer on the textfield text that the user wants to edit!

emmanmills (approved):
Looks good!

styleguide.config.js
```diff
@@ -1,8 +1,9 @@
 const path = require('path')
+const pkg = require('./package.json');
 const ignoreComponents = ['**/*.Mobile.tsx','**/*.Desktop.tsx']
 
 module.exports = {
-    title: 'Synapse React Client Component Library',
+    title: `Synapse React Components v${pkg.version}`,
```
jay-hodgson, line 6: added current package version

styleguide.config.js
```diff
@@ -36,14 +37,16 @@ module.exports = {
...
-    // This works when running the local server, but not when built.
-    // styles: {
-    //     StyleGuide: {
-    //         content: {
-    //             maxWidth: 1300
-    //         }
-    //     }
-    // },
+    theme: {
+        maxWidth: 1300,
```
jay-hodgson, line 41: found out how to properly customize the maxWidth (of the content)!

emmanmills: Oops! Didn't see this. Do you need to add a unit after 1300?

PR #770, opened 2020-11-18
PORTALS-1748 - Retheming bootstrap components with bootstrap-sass

nickgros:
I'm not 100% confident that this won't break things. In fact, I know it causes changes on Synapse.org related to its custom Bootstrap 3 theme. It can be resolved here: https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4454, but I'm not sure if that change will cause further issues.

* I replaced `node-sass` with `sass`. `node-sass` [is deprecated](https://sass-lang.com/blog/libsass-is-deprecated). This was further motivated by `sass` playing more nicely with rollup when resolving imports in node_modules.
* I pulled in `bootstrap-sass`, which allows us to override [Bootstrap 3 variables](https://getbootstrap.com/docs/3.4/customize/#less-variables)
* To demo the advantages, I found a couple of buttons where we were manually applying Bootstrap 3 classes AND custom styling, and let the bootstrap variables do its magic.


package.json
```diff
@@ -147,8 +148,8 @@
     "predeploy": "yarn run build-gh && cp styleguide/index.html styleguide/404.html && cp favicon.png styleguide/favicon.png",
```
nickgros, line 148: Note that above "postinstall" still calls `npm rebuild node-sass`. I'm not sure what component is still using `node-sass`. `yarn why` tells us that only the `rollup-plugin-scss` is using it (which is overridden in the rollup config), but I ran into issues when running `yarn start`.

src/lib/style/abstracts/_variables.scss
```diff
@@ -1,37 +1,36 @@
 // -----------------------------------------------------------------------------
 // This file contains all application-wide Sass variables.
+// Take care to ensure these don't conflict with variables in Bootstrap 3 (https://getbootstrap.com/docs/3.4/customize/#less-variables)
+//   or Bootstrap 4 (node_modules/bootstrap/scss/_variables.scss)
 // -----------------------------------------------------------------------------
 
 $border-color-gray: #dddddf !default;
-$gray-dark: #dcdcdc !default;
-$gray-light: #f9f9f9 !default;
```
nickgros, line 7: These variables conflicted with Bootstrap 3. Just renamed to $SRC-gray-...

src/lib/style/components/synapse_form_wrapper/_synapse-form-wrapper.scss
```diff
@@ -87,10 +87,10 @@
       padding-left: 0.8rem;
     }
     .panel {
-      background-color: $gray-light;
+      background-color: $SRC-gray-light;
     }
     .file-table {
-      @include calc(margin-bottom, #{$space-unit} * #{3});
+      @include calc(margin-bottom, "#{$space-unit} * #{3}");
```
nickgros, line 93: Couple of places where `sass` couldn't parse this while `node-sass` could. I believe that the change here yields the expected behavior.

src/lib/style/vendors-extensions/_bootstrap3-overrides.scss
```diff
@@ -0,0 +1,14 @@
...
+$icon-font-path: '~bootstrap-sass/assets/fonts/bootstrap/'; // https://github.com/webpack-contrib/sass-loader/issues/40#issuecomment-70269553
+
+// Override variables for theming
+$brand-primary: $primary-action-color;
+
+// Import bootstrap
+@import 'node_modules/bootstrap-sass/assets/stylesheets/bootstrap';
+
+// Bootstrap will override some desired styling (e.g. radio buttons, default text color) if we don't import this:
+@import 'src/lib/template_style/Index.scss';
```
nickgros, line 14: This import fixed the radio buttons (which was the most notable style change that led me to hesitate continuing with this change). It didn't seem like we were pulling it in though. What's the purpose of Index.scss, and should we be importing it here?

src/lib/containers/Login.tsx
```diff
@@ -199,14 +200,14 @@ class Login extends React.Component<Props, State> {
...
           />
           {this.getLoginFailureView()}
-          <button
+          <Button
+            variant="primary"
             onSubmit={this.handleLogin}
             type="submit"
-            className="btn SRC-primary-background-color SRC-hoverWhiteText
-              SRC-whiteText m-1 SRC-google-button SRC-marginBottomTen"
+            className="SRC-google-button SRC-marginBottomTen"
```
nickgros, line 207: On top of this being less code/fewer CSS classes, the button also gets the default bootstrap hover behavior (it previously had no hover behavior).

zimingd: @nickgros 

The react-bootstrap dependency is actually for Bootstrap 4. If you ever run into any `react-bootstrap` tags that get styled weirdly, this is the reason.

`<Button>` will work because it happens to have the same `className` and HTML element type in both Bootstrap 3 and Bootstrap 4

I'm personally hesitant about adding more confusion to the Bootstrap 3/4 ambiguity (i.e. continuing using the Bootstrap 4 react library for more elements intended for Bootstrap 3 styling).

nickgros: This is a good point! I think we still get some benefit because we could still do `<button className="btn btn-primary"> ...` instead of `<button className="btn btn-primary SRC-primary-background-color otherStylesEtc"> ...`

nickgros:
I think instead of manipulating Bootstrap 3, we should try to use the bootstrap-4-backport class that @zimingd added where possible since primary-color theming has been integrated there.

PR #771, opened 2020-11-19
PORTALS-1759

jay-hodgson:


src/lib/containers/AccountLevelBadge.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { ReactComponent as Registered } from '../assets/icons/account-registered.svg'
+import { ReactComponent as Certified } from '../assets/icons/account-certified.svg'
+import { ReactComponent as Validated } from '../assets/icons/account-validated.svg'
+import { Error } from './Error'
+
+export type AccountLevelBadgeProps = {
+  userId: string
+}
+
+export const accountLevelRegisteredLabel:string = 'Registered'
```
nickgros, line 14: Nit: I think the type is inferred

src/lib/containers/AccountLevelBadge.tsx
```diff
@@ -0,0 +1,96 @@
...
+        setError(err)
+      } finally {
+        setIsLoading(false)
+      }
+    }
+
+    getData()
+  }, [userId])
+  if (isLoading) {
+    return <></>
```
nickgros, line 53: Just double checking that this is the desired behavior on load

nickgros (approved):

src/lib/containers/AccountLevelBadge.tsx
jay-hodgson: Went back and forth on whether we should have a loading progress on this component.  Since it typically loads so fast, I thought flashing a loading bar might be more distracting in this case. 
If there was a higher level component that showed many of these in a list (for example), I think that component should have a loading screen.

PR #774, opened 2020-11-24
PORTALS-1762 - component that renders children if in view.

jay-hodgson:
and continues to use the rendered children if scrolled out of view


styleguide.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
     tocMode: 'collapse',
     pagePerSection: true,
     skipComponentsWithoutExample: true,
+    styleguideComponents: {
+        Wrapper: path.resolve(__dirname, 'src/lib/containers/RenderIfInView'),
```
jay-hodgson, line 11: This wraps every component in the styleguidist component demo page with this component (which is really the only way to go since the All Components section is only going to get bigger!)

src/lib/containers/RenderIfInView.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
```
jay-hodgson, line 1: This is my code.  This is the component that I'm going to export for portals (with the ability to force pre-render on the portal side).  Every dynamically generated SynapseComponent will be wrapped in one of these, unless you set some new flag (like forcePreload) to true (which could be done for complex components down the page that need extra time to load!).

emmanmills (approved):
Nice package! I can see why it's useful because it doesn't load everything all at once.

PR #775, opened 2020-11-24
Portals-1761: Governance pledge

emmanmills:
Screen Shot 2020-11-24 at 9 27 54 AM


src/lib/containers/TermsAndConditions.tsx
```diff
@@ -0,0 +1,211 @@
...
+    icon: speakerSvg,
+    label: 'I will not use data for marketing and/or advertising.',
+    description: ''
+  },
+  {
+    icon: flagSvg,
+    label: `I will report any suspected data breach or misuse to <a href="mailto:privacyofficer@sagebase.org">privacyofficer@sagebase.org</a> within 2 business days of initial incident knowledge.`,
+    description: ''
+  },
+]
```
emmanmills, line 70: Not sure if the terms and conditions should live somewhere outside the component. If so, where would be a good place to store this?

src/lib/containers/TermsAndConditions.tsx
```diff
@@ -0,0 +1,211 @@
...
+              <Button variant="secondary" onClick={clearSignature}><RefreshSvg />CLEAR SIGNATURE</Button>
+              <Button variant="secondary" onClick={closeSignatureDialog}>CANCEL</Button>
+              <Button
+                variant="primary"
+                className={submitBtnActive ? "" : "btn-disabled"}
+                active={submitBtnActive}
+                onClick={submitSignature}>Submit Signature
+              </Button>
+            </Modal.Footer>
+          </Modal>
```
emmanmills, line 197: FYI, we may not need the signature pad according to Stockard and she is checking with Governance. But I implemented it anyway just in case.

src/lib/containers/TermsAndConditions.tsx
```diff
@@ -0,0 +1,211 @@
...
+          <div className="terms-signature-action">
+            { trimmedDataUrl
+              ? <img onClick={displaySignatureDialog} src={trimmedDataUrl} />
+              : <Button onClick={displaySignatureDialog} className="btn-tap-sign">Tap to Sign</Button>
+            }
+          </div>
+        </div>
+        <br /><br /><br /><br /><Button onClick={handleSubmit}>TO BE DELETED: Submit form to debug</Button>
+      </form>
```
emmanmills, line 206: This is the button for form submission and the callback function for now is just logging the states. Should I just remove it?

package.json
```diff
@@ -74,6 +75,7 @@
     "react-plotlyjs-ts": "^2.2.1",
     "react-router-dom": "5.1.2",
     "react-scripts": "^3.4.3",
+    "react-signature-canvas": "^1.0.3",
     "react-sizeme": "^2.6.12",
```
emmanmills, line 79: This package is for signature pad.

src/lib/containers/TermsAndConditions.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { ReactComponent as penSvg } from '../assets/icons/terms/pen.svg'
+import { ReactComponent as speakerSvg } from '../assets/icons/terms/speaker.svg'
+import { ReactComponent as flagSvg } from '../assets/icons/terms/flag.svg'
+import { ReactComponent as RefreshSvg } from '../assets/icons/terms/refresh.svg'
+import { Button, Modal } from 'react-bootstrap'
+import SignatureCanvas from 'react-signature-canvas'
+import ReactSignatureCanvas from 'react-signature-canvas'
+
+export type TermsAndConditionsProps = {
+  token?: string
```
jay-hodgson, line 18: I'm having a difficult time understanding the contract between the outer page and this component.  
I don't think you need the session token.
But I think you need to add a callback function in the props that returns a boolean (you would call on every change event, to let the caller know if all items have been consented to).   Then the outer page can set up the enable states based on this value.

emmanmills: I updated the code to use a callback function to send back a boolean to the caller. 

```
export type TermsAndConditionsProps = {
  onFormChange: Function
}
```

src/lib/containers/TermsAndConditions.tsx
emmanmills: Button code has been removed.

src/lib/containers/TermsAndConditions.tsx
jay-hodgson: Suspect that this informed consent text should not change often, but you bring up a good point.  One possibility (that we've used elsewhere in this project) is to use a Synapse Table for the structured content.  Or it could be an external json file.  In this particular case, I don't think we need to externalize this text/copy.

jay-hodgson (approved):

PR #780, opened 2020-11-30
Upgrade from node-sass to dart sass

nickgros:
https://sagebionetworks.jira.com/browse/PORTALS-1767


nickgros:
Portals will need to be updated to use Dart Sass exclusively. It currently uses node-sass to read SRC sass files, which would break with these changes.

PR #797, opened 2020-12-09
Newest changes per conversations with Milen and team.

jonryser:


jay-hodgson:
Ok to merge, since this is just going into the GenUI branch

PR #799, opened 2020-12-10
Released 1.15.52 of SRC using a new GitHub Actions workflow script

jay-hodgson:
Has changes from SWC-5199 as well.
Note that the initial build from Sage-Bionetworks/Synapse-React-Client will fail, since it will not have the NPM secret access token...

### DIFF
--- a/.github/workflows/npmpublish.yml
+++ b/.github/workflows/npmpublish.yml
@@ -1,0 +1,34 @@
+name: "NPM Publish and GitHub tag"
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - run: yarn install
+      - run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+      # Set PACKAGE_VERSION, pull from package.json
+      - name: package-version
+        run: node -p -e '`PACKAGE_VERSION=${require("./package.json").version}`' >> $GITHUB_ENV
+      # Create the tag using the new PACKAGE_VERSION value
+      - name: Create tag
+        uses: actions/github-script@v3
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,34 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+      # Set PACKAGE_VERSION, pull from package.json
+      - name: package-version
+        run: node -p -e '`PACKAGE_VERSION=${require("./package.json").version}`' >> $GITHUB_ENV
+      # Create the tag using the new PACKAGE_VERSION value
+      - name: Create tag
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ github.token }}
```
ahayden, line 27: Should be `secrets.GITHUB_TOKEN`?

.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,34 @@
...
+          script: |
+            github.git.createRef({
```
ahayden, line 29: It is OK to have multiple tags with the same `PACKAGE_VERSION`, or do you only want to add tags when package.json changes?

.github/workflows/npmpublish.yml
jay-hodgson: Hey, my [tag](https://github.com/jay-hodgson/Synapse-React-Client/tags) was created when I tested this on my forked repo (pulling into my main), so it seems to work.  But is secrets.GITHUB_TOKEN preferable???

.github/workflows/npmpublish.yml
jay-hodgson: This is a great point.  Currently, by convention, we only pull into main on release of a new version (develop is our working branch).  But it would be nice to have additional safeguards.  Tagging is really the stretch objective here.  Before this change, there was no way to know that[ what the developer was publishing matches what was in main](https://sagebionetworks.jira.com/browse/PORTALS-792)!

jay-hodgson:
@ahayden said "the security of it looks fine to me", so I think it's ok to merge.  But if we break away from our convention (of only pushing to main on release), then our tags could get messy (and the npm publish will fail, since that version has already been released).

PR #802, opened 2020-12-11
PORTALS-1751 Use DemoStyle.css in Styleguidist

nickgros:
I've been using the styleguide to verify changes to buttons and found this defect. Just using main.css breaks some styles because we are missing Index.scss. This is fixed by using DemoStyle.css (which was used in the react-scripts demo), which pulls in both main.scss and Index.scss.

As an example, see QueryWrapperPlotNav. We lose the .SRC-light-button class.

without Index.scss:
image

with Index.scss:
image


nickgros:
Putting these changes in #822 instead

PR #805, opened 2020-12-14
PORTALS-1786: Limit the image height in ProjectViewCard

nickgros:
Before, a taller-than-expected image would push contents off of the card:

image

After:

image

### DIFF
--- a/src/lib/style/components/_project-view-card.scss
+++ b/src/lib/style/components/_project-view-card.scss
@@ -10,7 +10,15 @@
   text-align: left;
 
   img {
+    display: block;
     max-width: 100%;
+    max-height: 30%;
+    width: auto;
+    height: auto;
+
+    // Center the image
+    margin-left: auto;
+    margin-right: auto;
   }
 
   &__ImagePlaceholder {
--- a/src/lib/style/components/_project-view-card.scss
+++ b/src/lib/style/components/_project-view-card.scss
@@ -13,8 +13,6 @@
     display: block;
     max-width: 100%;
     max-height: 30%;
-    width: auto;
-    height: auto;
 
     // Center the image
     margin-left: auto;


src/lib/style/components/_project-view-card.scss
```diff
@@ -10,7 +10,15 @@
...
   img {
+    display: block;
     max-width: 100%;
+    max-height: 30%;
+    width: auto;
+    height: auto;
+
+    // Center the image
+    margin-left: auto;
+    margin-right: auto;
```
emmanmills, line 21: This solution to use margin to center the image is interesting. I normally would apply `text-align: center` on the image's parent div to center the image, providing that the image is the only element within the parent div. I have to try your solution sometimes.

src/lib/style/components/_project-view-card.scss
```diff
@@ -10,7 +10,15 @@
   text-align: left;
 
   img {
+    display: block;
     max-width: 100%;
+    max-height: 30%;
+    width: auto;
+    height: auto;
```
emmanmills, line 17: Unless somewhere else in the codebase is setting the width and height for the  tag, their default values should just be `auto`. See the Formal Definition section on [width](https://developer.mozilla.org/en-US/docs/Web/CSS/width) & [height](https://developer.mozilla.org/en-US/docs/Web/CSS/height). But it's okay to leave it too in case you want to enforce the CSS rules.

emmanmills (approved):
Looks great!

src/lib/style/components/_project-view-card.scss
nickgros: I have other elements in the div that utilize the parent's `text-align: left`

src/lib/style/components/_project-view-card.scss
nickgros: You're right. I will remove these.

PR #806, opened 2020-12-15
SWC-5346: Shrink mobile Synapse homepage images by about 25%

nickgros:
https://sagebionetworks.jira.com/browse/SWC-5346

### DIFF
--- a/src/lib/style/components/_synapse-homepage.scss
+++ b/src/lib/style/components/_synapse-homepage.scss
@@ -69,14 +69,15 @@
     }
     .HeroContainer {
       .HeroImage {
-        max-height: 25vw;
+        max-height: 250px;
+        max-width: 100%;
       }
     }
     .SectionGraphic {
       display: block;
       margin: auto;
-      max-height: 200px;
-      max-width: 400px;
+      max-height: 175px;
+      max-width: 100%;
     }
 
     &__MarginAutoDesktop {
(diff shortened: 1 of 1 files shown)


src/lib/style/components/_synapse-homepage.scss
```diff
@@ -161,7 +171,7 @@
   @media (max-width: 1100px) {
     // No grid for the HeroContainer at <= 1100px
     .HeroImage {
-      max-height: 250px;
+      max-height: 22vw;
```
emmanmills, line 174: I saw the staging you currently have the hero image with width 700px. If the design team wants that image to be 25% less than the non-mobile version, FYI we could also do this:

$hero-image-width: 700px;
.SynapseHomepage .HeroContainer .HeroImage {
 width: calc($hero-image-width * 0.75)
}

However the mobile viewport could get smaller and smaller so maybe even 25% less than the desktop version will not be enough, so use `vw` or `vh` might work better. Just curious though, why do you use `vw` for the max-height attribute and not the max-width attribute?

nickgros: The '25%' is relative to the mobile image size in the first iteration, so the ratio of image size on mobile vs desktop isn't a design requirement.

I opted to change this and not use vw and use a max-height suggested by design. I also just set max-width to 100% so that the image scales down as the viewport shrinks while in mobile view

emmanmills (approved):
Looks good to me!

PR #808, opened 2020-12-15
Portals-1768: Add direct download column

emmanmills:
Screen Shot 2020-12-15 at 9 52 01 AM


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -791,20 +816,37 @@ export default class SynapseTable extends React.Component<
...
+
+      if (isShowingDownloadColumn) {
+        rowContent.unshift(
+          <td className="SRC_noBorderTop direct-download">
+            <DirectDownload
+              key={"direct-download-"+rowSynapseId}
+              token={token}
+              fileEntityHandle={fileEntityHandle}
+            ></DirectDownload>
+          </td>
```
emmanmills, line 846: I wonder if I should move the icon switching concept outside the <DirectDownload> component. For example:

<DirectDownload
key={"direct-download-"+rowSynapseId}
token={sting}
isAuthorized={boolean}
downloadLink={string}
externalUrl={string}
></DirectDownload>

src/lib/containers/FileEntityHandleQueryWrapper.tsx
```diff
@@ -0,0 +1,89 @@
...
+      mounted = false
+    }
+  }, [token, rows])
+
+  const getFileEntityHandle = () => {
+    return new Promise((resolve, reject) => {
+      const fileHandlePromises = rows?.map(async (row, rowIndex) => {
+        const entityVersionNumber = row.versionNumber?.toString()
+        const rowSynapseId = `syn${row.rowId}`
+        const entity = await SynapseClient.getEntityResult(token, rowSynapseId, entityVersionNumber)
```
jay-hodgson, line 49: Instead of calling getEntityResult to get each entity, would an EntityHeader be enough information (it has the type, so you could tell if it was a FileEntity)?  If so, then you could use SynapseClient.getEntityHeader with a ReferenceList (which is list of SynapseID+version objects), which would make this a single call.

jay-hodgson: For the FileEntites, you might be able to then use SynapseClient.getFiles() to get all of the FileHandles in a single call (set includeFileHandles to true).

emmanmills: I have updated the code to call SynapseClient.getFiles() only once but still need to use getEntityResult to get a list of file handle ids. The list of file handle ids is required to use SynapseClient.getFiles().

jay-hodgson: This is calling getFiles() once per row.  Ideally you would collect all of the FileEntities, and make a single bulk call to getFiles()

src/lib/utils/SynapseClient.ts
```diff
@@ -1585,6 +1606,42 @@ export const getFileHandleContent = (
...
+export const getFileResult = (
+  fileEntity: FileEntity,
+  sessionToken?: string,
+  includeFileHandles?: boolean,
+  includePreSignedURLs?: boolean,
+  includePreviewPreSignedURLs?: boolean,
+): Promise<FileResult> => {
+
+  return new Promise((resolve, reject) => {
+    const fileHandleAssociationList: FileHandleAssociation[] = [
```
jay-hodgson, line 1618: This is an array of FileHandleAssociations.  Currently you're creating one, but this could ask for all of them in a single service call.

src/lib/utils/SynapseClient.ts
```diff
@@ -1585,6 +1606,42 @@ export const getFileHandleContent = (
   })
 }
 
+export const getFileResult = (
+  fileEntity: FileEntity,
```
jay-hodgson, line 1610: This could be an array of FileEntities

src/lib/containers/FileEntityHandleQueryWrapper.tsx
emmanmills: Sorry, the code you were looking at is outdated. Now the getFiles() call is on line 85 and it's called only once.

src/lib/utils/SynapseClient.ts
emmanmills: Actually this function is only used for retrieving the download link now after code refactoring. So I only need 1 file handle data to get the download link when a user click the download button instead of a list. Is there a function in SynapseClient I can re-use to just get just 1 file handle data?

jay-hodgson: Sorry, I was looking at old code.  This is great.

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: It would also mean moving the logic to get the download link on click (since you can't pass these in, because the presigned urls expire).  I like what you've done

src/lib/containers/FileEntityHandleQueryWrapper.tsx
jay-hodgson: My mistake.  Looks great now!

src/lib/utils/SynapseClient.ts
jay-hodgson: May want to change other functions in SynapseClient to use this getFileResult()!  I'm looking at getFileEntityFileHandle() and getFileEntityContent(),

jay-hodgson (approved):

PR #811, opened 2020-12-15
SWC-5348: All links to dashboard say 'View Your Dashboard'

nickgros:


src/lib/containers/SynapseHomepage.tsx
```diff
@@ -144,7 +144,7 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
               variant="primary"
               href={token ? DASHBOARD_LINK : REGISTRATION_LINK}
             >
-              Get Started Now
+              {token ? 'View Your Dashboard' : 'Get Started Now'}
```
nickgros, line 147: Opted to do this individually on each button instead of one variable because we have another button that alternates between 'View Your Dashboard' and 'Sign In'. It would seem like a new variable would belong on that button, but it doesn't.

emmanmills (approved):
Looks good!

PR #814, opened 2020-12-16
Fix type export

nickgros:
See https://github.com/facebook/flow/issues/1379


emmanmills (changes requested):
I think we need to bump the version also.

emmanmills (approved):

PR #823, opened 2020-12-21
PORTALS-1767 (PR 2): Use @use, @forward, namespacing

nickgros:
Depends on #822, https://github.com/Sage-Bionetworks/portals/pull/534.

First, see #822, which aims to replace node-sass with Dart Sass in our toolchain.

`@use` and `@forward` are replacements for `@import` that are supported in Dart Sass (and will never be supported by node-sass because it is deprecated).

This PR uses these where possible. Notably, we can use a namespace for our variables, eliminating the risk of a variable name collision. No styles or behavior should be changed.

### DIFF
--- a/src/lib/style/components/_resources.scss
+++ b/src/lib/style/components/_resources.scss
@@ -1,8 +1,11 @@
+@use '../abstracts/variables' as SRC;
+@use 'sass:color';
+
 // TODO: These colors to be configurable for the other portals.
 .Resources {
   background-color: rgba(64, 123, 160, 0.05);
   a {
-    color: $primary-action-color;
+    color: SRC.$primary-action-color;
   }
   .control-container {
     background: white;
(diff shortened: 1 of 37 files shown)


src/lib/style/components/_resources.scss
```diff
@@ -23,10 +26,12 @@
         padding: 10px;
         &.isSelected,
         &:not(.gap-fill):hover {
-          color: $primary-action-color;
-          // Hex value can also set opacity, using two extra values at the end
-          background-color: #{$primary-action-color}0a;
-          border-right: 2px solid $primary-action-color;
+          color: SRC.$primary-action-color;
+          background-color: color.adjust(
```
nickgros, line 30: see if we are doing the opacity set trick in code

jay-hodgson (approved):

PR #824, opened 2021-01-04
PORTALS-1786: Use object-fit to scale down + keep aspect ratio

nickgros:
https://sagebionetworks.jira.com/browse/PORTALS-1786

### DIFF
--- a/src/lib/style/components/_project-view-card.scss
+++ b/src/lib/style/components/_project-view-card.scss
@@ -11,12 +11,12 @@
 
   img {
     display: block;
-    max-width: 100%;
-    max-height: 30%;
-
-    // Center the image
     margin-left: auto;
     margin-right: auto;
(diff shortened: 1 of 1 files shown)


nickgros:
Before:

image


After: 

image

nickgros:
At full size,

Before: 
image


After:

image

src/lib/style/components/_project-view-card.scss
```diff
@@ -11,12 +11,12 @@
 
   img {
     display: block;
-    max-width: 100%;
-    max-height: 30%;
-
-    // Center the image
     margin-left: auto;
     margin-right: auto;
+    object-fit: contain;
```
nickgros, line 16: This won't work in IE: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

PR #830, opened 2021-01-07
SWC-5331 and SWC-5356

nickgros:
Token permissions should behave like links (no underline, added help cursor for clarity): https://sagebionetworks.jira.com/browse/SWC-5331

Pressing enter shouldn't reload page in SWC. This is fixed by using the React-Bootstrap form tag. https://sagebionetworks.jira.com/browse/SWC-5364

### DIFF
--- a/src/lib/containers/personal_access_token/CreateAccessTokenModal.tsx
+++ b/src/lib/containers/personal_access_token/CreateAccessTokenModal.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Button, FormControl, Modal, ModalBody } from 'react-bootstrap'
+import { Button, Form, FormControl, Modal, ModalBody } from 'react-bootstrap'
 import { SynapseClient } from '../../utils'
 import { AccessTokenGenerationRequest } from '../../utils/synapseTypes/AccessToken/AccessTokenGenerationRequest'
 import { scopeDescriptions } from '../../utils/synapseTypes/AccessToken/ScopeDescriptions'
@@ -103,7 +103,7 @@ export const CreateAccessTokenModal: React.FunctionComponent<CreateAccessTokenMo
           </>
         ) : (
           <div className="SRC-marginFive">
-            <form>
+            <Form>
               <div className="SRC-marginBottomTen">
                 <label className="SRC-boldText">Token Name</label>
                 <FormControl
@@ -135,7 +135,7 @@ export const CreateAccessTokenModal: React.FunctionComponent<CreateAccessTokenMo
                   onChange={() => setModifyAccess(!modifyAccess)}
                 ></Checkbox>
               </div>
-            </form>
+            </Form>
             <div className="SRC-center-text">
               {showErrorMessage && (
                 <ErrorBanner error={errorMessage}></ErrorBanner>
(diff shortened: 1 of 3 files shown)


src/lib/style/components/_personal-access-tokens.scss
```diff
@@ -18,10 +18,7 @@
   .SRC-scopeName {
     margin-inline-start: 3px;
     margin-inline-end: 3px;
-  }
-
-  .SRC-scopeName:hover {
-    text-decoration: underline;
+    cursor: help;
```
jay-hodgson, line 21: Do the scope names act like normal checkbox text (you can click to select the scope)?  

The **help** cursor type is not a commonly-used interface anymore.
We don't even use it for our contextual help widgets (because we want to indicate that they're clickable).

src/lib/containers/personal_access_token/CreateAccessTokenModal.tsx
```diff
@@ -103,7 +103,7 @@ export const CreateAccessTokenModal: React.FunctionComponent<CreateAccessTokenMo
           </>
         ) : (
           <div className="SRC-marginFive">
-            <form>
+            <Form>
```
jay-hodgson, line 106: I assume this is for styling purposes only, right?  Out of curiosity, does it look right without the form (because we really don't use any of the form functionality here)?

src/lib/style/components/_personal-access-tokens.scss
nickgros: These scope names are not in the modal (there's no checkbox). They appear on the cards of tokens that are already created. When you hover over a scope, you get a tooltip describing the scope. The issue was that the scope names appear clickable because of the underline on hover, but there's no click behavior.

I can remove the cursor type entirely; just not sure how to demonstrate that these pieces are interactive, but not clickable

src/lib/containers/personal_access_token/CreateAccessTokenModal.tsx
nickgros: Form is not necessary. Getting rid of it fixes the issue!

src/lib/style/components/_personal-access-tokens.scss
jay-hodgson: I think setting it to **default** is preferable for this case.

jay-hodgson: (not **pointer**, which would indicate that it's clickable.)

jay-hodgson: Communicating that they are interactive should not be the responsibility of the cursor, since by the time you see the cursor you see that it's interactive (from the hover tooltip!).

src/lib/containers/personal_access_token/CreateAccessTokenModal.tsx
jay-hodgson: Yay!

src/lib/style/components/_personal-access-tokens.scss
nickgros: Makes sense! I've set the cursor to default on hover.

jay-hodgson (approved):

PR #833, opened 2021-01-08
add EvaluationEditorPage, a component combining EvaluationEditor and Evaluation. Remove export of EvaluationEditor and EvaluationRoundEditorList

zimingd:


zimingd:
image
image

jay-hodgson:
@zimingd , could you point me to the design file that shows this UX?  I would think the Evaluation and Rounds would be in different UI groups, given the requirement to create the Evaluation before adding rounds.

zimingd:
The design does not explicitly mention anything about creating an new Evaluation round, but it does bundle the EvaluationEditor and EvaluationRoundEditorList together in the same page for editing an existing evaluation round
https://www.figma.com/file/E5zIAPsI2fagDc8btWGU06Ol/Challenge-Evaluation-Queue?node-id=479%3A1821

This is what is currently shown for creating a new round:
image

This is problematic because it still uses the **deprecated** Quota version of limiting submissions

When I hooked up my dev stack to simply make it use the new Editor, this appears because there's zero communication between the two components.

image


Once I click the "Save" button, the EvaluationRound editor does not change
image

jay-hodgson:
What do you think about showing the Add Round button when Creating a new evaluation, but on click it would show the info alert that you added (can be edited once the Evaluation has been saved)?  I'd like to avoid showing the alert unconditionally (when creating a new Evaluation)

jay-hodgson (approved):

PR #834, opened 2021-01-08
PORTALS-1770: use mobile breakpoints, and set a min height per plot row

jay-hodgson:


jay-hodgson:
This PR fixes the card height by setting a minimum height per plot to something more reasonable. 
Before: 
Screen Shot 2020-12-01 at 3 30 25 PM
After:
Screen Shot 2021-01-07 at 4 43 49 PM

src/lib/style/components/_facet-plots-card.scss
```diff
@@ -1,7 +1,8 @@
-.FacetPlotsCard {
+.FacetPlotsCard {  
   background-color: white;
+  overflow: hidden;
```
nickgros, line 3: What's the purpose of `overflow: hidden` here?

nickgros (approved):
Looks great, just one question

src/lib/style/components/_facet-plots-card.scss
jay-hodgson: In the worst case scenario for mobile breakpoints (like the styleguidist context, where the home page component does not take up the full width of the viewport!), this is to guarantee that the component content (like the plot legend labels, for example) do not extend outside of the card boundary.  That content will be cut off in this rare case.

nickgros: Ah, totally makes sense. Thanks!

PR #845, opened 2021-01-13
PORTALS-1802: change 'Unannotated' to 'Not Assigned' when facet value is NOT_SET

jay-hodgson:
Also, changed to use the SynapseConstant value instead of rewriting the NOT_SET value all over the codebase.


nickgros (approved):
Looks good!

PR #846, opened 2021-01-14
PORTALS-1593

nickgros:


src/lib/style/abstracts/_variables.scss
```diff
@@ -8,7 +8,7 @@ $gray-light: #f9f9f9 !default;
 $primary-action-color: rgb(64, 123, 160) !default;
 $primary-bgcolor-rgba: rgba(201, 66, 129, 0.03) !default;
 
-$secondary-action-color: #5960a5 !default;
+$secondary-action-color: rgb(98, 172, 98) !default;
```
nickgros, line 11: Changed default from purple to 'Synapse Green' so any components that inadvertently end up in SWC won't look too out of place.

rollup.config.js
```diff
@@ -64,7 +64,7 @@ export default {
   },
   // NOTE - the order matters for the extensions below
   plugins: [
-    resolve({ extensions, browser: true }),
+    resolve({ extensions, browser: true, preferBuiltins: true }),
```
nickgros, line 67: The default value is `true`, just suppressing log messages that come up when this value is not explicitly set.

jay-hodgson (approved):

PR #850, opened 2021-01-14
SWC-5374 Remove unscoped button style

nickgros:
Before:
image

After:
image

### DIFF
--- a/src/lib/style/components/_download-list.scss
+++ b/src/lib/style/components/_download-list.scss
@@ -68,10 +68,6 @@
   margin-left: 5px;
 }
 
-.btn {
-  margin-left: 15px;
-}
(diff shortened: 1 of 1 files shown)


src/lib/style/components/_download-list.scss
```diff
@@ -68,10 +68,6 @@
   margin-left: 5px;
 }
 
-.btn {
```
nickgros, line 71: This was previously `.action-button`. There is another scoped `.action-button` selector that also applies `margin-left: 15px`. I think that the scoped style is applied to every button that this style was originally meant to apply to, so it is redundant.

jay-hodgson (approved):

PR #852, opened 2021-01-15
SWC-5375

nickgros:
SWC overrides the alert style, but can't override the text color because the selector is too specific. We'll just change all alerts in SRC to use the ErrorBanner style.

SWC will need to be updated to apply `!important` or a [more specific selector](https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4496/files), since SRC CSS is applied after the custom bootstrap theme.

Error message currently looks like this:
image

My main concern with this approach is that it is very brittle. A preferred approach would be importing the SCSS into SWC and overriding the variables. However, this complicates the build process.

### DIFF
--- a/src/lib/containers/ErrorBanner.tsx
+++ b/src/lib/containers/ErrorBanner.tsx
@@ -63,7 +63,7 @@ export const ErrorBanner = (props: ErrorProps) => {
     stringError = error
   }
   return (
-    <div className="Error">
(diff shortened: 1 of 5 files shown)


jay-hodgson:
After talking, @nickgros decided to investigate incorporating the SRC scss into the SWC build process instead (where variables could be overridden).

PR #861, opened 2021-01-21
temporary import fix

zimingd:


zimingd:
I'm looking deeper into imports to determine why `import { Evaluation, EvaluationStatus } from '../../utils/synapseTypes'` would cause an error

emmanmills (approved):
Looks good!

PR #871, opened 2021-01-25
PORTALS-1783: Add custom icon path

emmanmills:
Add an option to change where the svg icons are located.


jay-hodgson (approved):
Let's try it out!  Hopefully overriding the iconBasePath in the portal will do the trick.

PR #874, opened 2021-01-25
PORTALS-1812 

nickgros:
Remove baseUrl so VSCode only suggests relative imports

Also cleaned up the readme a bit


tsconfig.json
```diff
@@ -4,7 +4,6 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "baseUrl": "src",
```
nickgros, line 7: This is the critical line--when `baseUrl` is specified, VSCode will suggest imports relative to this directory, which will break at runtime.

Examples.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Markdown Rendering Example
+
+View the demo app incorporation of markdown [here](./src/demo/containers/App.tsx).
+
+To use the synapse markdown-it component you must pass it a wiki page id and an owner id. Additionally, you can configure an error view to display.
+
+| Props                                      | Explanation                                             |
+| ------------------------------------------ | ------------------------------------------------------- |
+| ownerId: String                            | ownerId for the synapse page                            |
```
jay-hodgson, line 51: I wonder if this documentation should be moved to the component itself in a way where it is output in the styleguidist live demo, and then we just point people to that live demo instead of these examples (which may become stale without us knowing).

PR #875, opened 2021-01-26
PORTALS-1803

jay-hodgson:


src/lib/containers/QueryWrapper.tsx
```diff
@@ -134,6 +135,7 @@ export default class QueryWrapper extends React.Component<
     this.getNextPageOfData = this.getNextPageOfData.bind(this)
     this.updateParentState = this.updateParentState.bind(this)
     this.getInitQueryRequest = this.getInitQueryRequest.bind(this)
+    const showFacetVisualization = props.defaultShowFacetVisualization ?? true
```
emmanmills, line 138: Just curious, where do you use `showFacetVisualization` in the code?

emmanmills (approved):
Thanks, it looks good!

src/lib/containers/QueryWrapper.tsx
jay-hodgson: Down on line 156 in this file.  It used to always set to true, but will now use the input param (if set)

PR #877, opened 2021-01-26
PORTALS-1808: Reformat Graphs Based On Screen Size

leonli33:
This fix is for Jira issue: https://sagebionetworks.jira.com/browse/PORTALS-1808.

Allows the size of the graph containers to shrink as the screen width gets smaller. This solves the problem of having three graphs that are stacked together, with two on top and one on the bottom. Some images of the fix can be seen below:

When the window size is normal:
Screen Shot 2021-01-26 at 12 01 13 PM

When the window size decreases:
Screen Shot 2021-01-26 at 12 03 37 PM

When the "Show All Graphs" button is pressed:
Screen Shot 2021-01-26 at 12 10 18 PM

### DIFF
--- a/src/lib/containers/widgets/facet-nav/FacetNav.tsx
+++ b/src/lib/containers/widgets/facet-nav/FacetNav.tsx
@@ -185,7 +185,7 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
       colorIndex: index,
     }
   })
-  const showMoreState = getShowMoreState()
+  const showMoreButtonState = getShowMoreState()
 
   if (error) {
     return <></>
@@ -200,7 +200,7 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
   } else {
     return (
       <>
-        <div className={`FacetNav ${showFacetVisualization ? '' : 'hidden'}${showMoreState === 'LESS' ? 'less' : ''}`}>
+        <div className={`FacetNav ${showFacetVisualization ? '' : 'hidden'} ${showMoreButtonState === 'LESS' ? 'less' : ''}`}>
           <div className="FacetNav__expanded">
             {expandedFacets.map((facet, index) => (
               <div key={facet.columnName}>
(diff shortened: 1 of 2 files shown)


conundrumgirl (approved):
Nice work! Thanks for the great description with visuals!

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -200,7 +200,7 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
   } else {
     return (
       <>
-        <div className={`FacetNav ${showFacetVisualization ? '' : 'hidden'}`}>
+        <div className={`FacetNav ${showFacetVisualization ? '' : 'hidden'}${showMoreState === 'LESS' ? 'less' : ''}`}>
```
jay-hodgson, line 203: I think you need a space here.  Otherwise, if showFacetVisualization=false and showMoreState='LESS' then we'll end up with the styles 'FacetNav hiddenless'

src/lib/style/components/facet_nav/_facet-nav.scss
```diff
@@ -30,7 +30,7 @@
   }
 }
 
-.container-full-width {
+.container-full-width .FacetNav.less{
```
jay-hodgson, line 33: Why do these changes fix the issue?  Today when showMoreState='LESS' (or any value), this css applies.  What am I missing @conundrumgirl ?

src/lib/containers/widgets/facet-nav/FacetNav.tsx
jay-hodgson: @conundrumgirl  explained it to me, I was confused by the showMoreState.  showMoreState is equal to 'LESS' when all the graphs are showing (indicating that's what the button would indicate).
The bug that I point out above would be a tricky state to get into, but it would happen.  User needs to click show more (show all graphs), and then hide the visualization.

jay-hodgson (approved):

conundrumgirl (approved):

PR #881, opened 2021-01-28
Portals-1783: Fix icon size issues

emmanmills:
1. Redo svg icon component to use material ui icon library
2. Delete unused packages and icon base path

Screen Shot 2021-01-28 at 9 48 59 AM


src/lib/containers/IconSvg.md
```diff
@@ -4,7 +4,6 @@
       options={{
         icon: 'data',
         color: '#e61899',
-        size: 'sm',
```
emmanmills, line 7: Delete the size option because material ui icons will use the surrounding font size to size the icon. But I might put it back later if we need to customize the icon size for bigger or smaller than surrounding fonts.

src/lib/containers/IconSvg.tsx
```diff
@@ -1,63 +1,67 @@
...
-        fill: ${color};
-        height: ${IconSize[size]};
-        padding-left: ${padding === 'left' ? IconPadding[size] : 0};
-        padding-right: ${padding === 'right' ? IconPadding[size] : 0};
-        path {
-          fill: ${color};
-        }
-      `
-    }
+  switch (icon) {
```
jay-hodgson, line 42: Ok, so this switch maps the icon name to either a custom or material UI icon.  Very nice!

jay-hodgson (approved):

PR #884, opened 2021-02-02
SWC-5342: Update login page style

emmanmills:
Screen Shot 2021-02-01 at 4 35 01 PM


src/lib/containers/LoginPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const thisClass = "login-panel-wrapper"
+
+  useEffect(() => {
+    if (mounted) {
+    parent = backgroundElementId ? (document.getElementById(backgroundElementId) as HTMLElement)
+      : (document.getElementById(thisClass)?.parentNode as HTMLElement)
+      if (parent) {
+        parent.classList.add("login-panel-wrapper-bg")
+      }
+    }
```
emmanmills, line 23: The purpose of this block is to set the background image css class to the parent of this component. The parent could be further up in the DOM tree depending on where is the best place to set the background.

src/lib/style/components/_login.scss
```diff
@@ -1,19 +1,155 @@
...
+
 .LoginComponent {
   max-width: 350px;
   background: white;
   font-family: Lato, sans-serif;
 
   .SRC-google-button,
   .SRC-login-button {
-    height: 50px;
+    //height: 50px;
```
jay-hodgson, line 11: ok to delete dead code

emmanmills: Thanks, just deleted it.

src/lib/containers/LoginPage.tsx
jay-hodgson: I'm unclear why you need to adjust the parent container background, instead of defining this component as the full page, but I agree that this approach should work.

emmanmills: I guess I was overthinking this. I wasn't sure how this component will be applied in Synapse so I provided a way to override background in any parent tags just in case. But you brought up a good point so I deleted the code to set background outside this component.

jay-hodgson: No need for useEffect() now

emmanmills: Deleted!

jay-hodgson (approved):

PR #893, opened 2021-02-04
PORTALS-1834: autolink label value if it's a Synapse ID

jay-hodgson:


jay-hodgson:
Like to autolink secondary label value if it looks like a Synapse ID (when it doesn't have explicit label linking rules).
Screen Shot 2021-02-03 at 6 09 38 PM

src/lib/containers/GenericCard.tsx
```diff
@@ -153,11 +153,35 @@ export const renderLabel = (args: {
...
+  }
+
+  if (!labelLink) {
+    // if this looks like a Synapse ID, then autolink.
+    if (str.match(SYNAPSE_REGX)) {
+      // its a synId
+      return <a
+          target="_blank"
+          rel="noopener noreferrer"
+          href={`https://www.synapse.org/#!Synapse:${str}`} className={className} style={style}>
```
emmanmills, line 176: The style object seems to be empty here. Should you delete it?

jay-hodgson: Deleted!

PR #898, opened 2021-02-05
PORTALS-1783: Bug fix for missing icon config

emmanmills:
This happen when the Synapse table status icon value changes and the portal config is not updated with the new value. For example "Partially Available" is changed to "Partially Released". We should always return the status value.


jay-hodgson (approved):
Seems reasonable

PR #901, opened 2021-02-09
PORTALS-1795:  Ability to define a column to use for a custom image (file handle column)

jay-hodgson:


jay-hodgson:
Tested existing AMP-AD and NF, existing generic card configurations rendering is unchanged.
This mechanism can be used for portal owners to define a custom image per card in the Synapse Table/View:
Screen Shot 2021-02-09 at 9 57 16 AM
Screen Shot 2021-02-08 at 6 14 53 PM

src/lib/containers/GenericCard.tsx
```diff
@@ -399,7 +403,7 @@ export default class GenericCard extends React.Component<
       // undefined, take default value from class
       marginTop: isHeader ? '0px' : undefined,
       marginBottom: isHeader ? '0px' : undefined,
-      paddingBottom: showFooter ? undefined : '15px',
+      paddingBottom: showFooter || imageFileHandleIdValue ? undefined : '15px',
```
jay-hodgson, line 406: Set the padding bottom to 15, unless we're showing a footer or an image on the left.

src/lib/containers/GenericCard.tsx
```diff
@@ -438,72 +442,83 @@ export default class GenericCard extends React.Component<
...
+        <div className={'SRC-portalCardMain'}>
+          {imageFileHandleIdValue && <div className="SRC-imageThumbnail">
+            <ImageFileHandle 
+              token={token}
+              fileHandleId={imageFileHandleIdValue}
+              tableEntityConcreteType={tableEntityConcreteType}
+              rowId={data![schema.id]}
+              tableId={tableId}
+            /></div>}
+          {!imageFileHandleIdValue && <div className="SRC-cardThumbnail">
```
jay-hodgson, line 454: Show an icon (or at least space for an icon) if the image is not set.

src/lib/containers/GenericCard.tsx
```diff
@@ -438,72 +442,83 @@ export default class GenericCard extends React.Component<
...
+              )}
+          </div>
         </div>
         {showFooter && (
           <CardFooter
             isHeader={false}
             secondaryLabelLimit={secondaryLabelLimit}
             values={values}
             columnIconOptions={columnIconOptions}
+            className={`${imageFileHandleIdValue ? 'hasImage' : 'hasIcon'}`}
```
jay-hodgson, line 521: Apply the class "hasImage" or "hasIcon" to footer div to change it's indent behavior (if icon, do what it does today, if image then do not try to align with the icon indent).

src/lib/containers/widgets/FileHandleLink.tsx
```diff
@@ -45,7 +45,12 @@ export const FileHandleLink = (props: FileHandleLinkProps) => {
             fileAssociateType,
             fileAssociateId,
             redirect,
-          )
+          ).then(url => {
+            window.open(url, '_blank')
```
jay-hodgson, line 49: SynapseClient.getActualFileHandleByIdURL() used to have this logic, but I want to make the same call with different success handling, so I changed it to return a Promise instead.

src/lib/containers/widgets/ImageFileHandle.tsx
```diff
@@ -0,0 +1,64 @@
+import { FileHandleAssociateType } from '../../utils/synapseTypes'
```
jay-hodgson, line 1: Very similar to FileHandleLink, but even less complicated (no display text value or download icon)

src/lib/style/components/_cards.scss
```diff
@@ -125,9 +144,11 @@ $text-color-lighter: #898989;
 
   .SRC-cardMetadata {
     display: block;
-    background-color: $color-gray-background;
-    margin-top: $baseline-grid;
-    padding: $baseline-grid $baseline-grid $baseline-grid 15%;
```
jay-hodgson, line 130: Now only applies the padding-left of 15% if the card is showing an icon (not an image)

jay-hodgson:
I went and implemented the lazy-load image recipe for this, but I have some concerns.  Using IntersectionObserver has a terrible UX (images flash into view when you scroll down).  But native img loading="lazy" has it's own issue in this context, since we would still get the presigned URL as soon as the component is loaded, but may not start downloading from it for some time (until the user scrolls down to the image).  I don't remember off hand what the expire time is, but I know it's not long (and it may take minutes for the user to scroll down to this card).
So, not lazy loading these for now, we will instead execute n calls to get the images (where n is the page size).

src/lib/utils/SynapseClient.ts
```diff
@@ -387,20 +387,14 @@ export const getActualFileHandleByIdURL = (
   fileAssociateType: FileHandleAssociateType,
   fileAssociateId: string,
   redirect: boolean = true,
-) => {
+): Promise<string> => {
   // get the presigned URL for this file handle and open it in a new tab
```
nickgros, line 391: Nit: Comment is outdated

jay-hodgson:
> I went and implemented the [lazy-load image recipe](https://github.com/thebuilder/react-intersection-observer/blob/master/docs/Recipes.md#lazy-image-load) for this, but I have some concerns. Using IntersectionObserver has a terrible UX (images flash into view when you scroll down). But native img loading="lazy" has it's own issue in this context, since we would still get the presigned URL as soon as the component is loaded, but may not start downloading from it for some time (until the user scrolls down to the image). I don't remember off hand what the expire time is, but I know it's not long (and it may take minutes for the user to scroll down to this card).
> So, not lazy loading these for now, we will instead execute n calls to get the images (where n is the page size).

Rediscovered that we can use IntersectionObserver and load before it's actually in the viewport, so I added that (500px away from the image, it will ask for the presigned URL and load the image)

nickgros (approved):
👍

src/lib/style/components/_cards.scss
```diff
@@ -28,6 +28,9 @@ $text-color-lighter: #898989;
   margin-top: 30px; // just for test page
   margin-bottom: 30px; // just for test page
 
+  .SRC-portalCardMain {
+    display: flex;
```
nickgros, line 32: Based on the images I don't think it would make a difference as the cards are fairly crowded, but consider specifying `justify-content` (unless the default `flex-start` is desired).

emmanmills (approved):
Looks good to me!

PR #909, opened 2021-02-15
Portals-1841: Async Download Bug

emmanmills:
Fix an issue with Study Details page when the correct file download data is being overwritten by data that's already been fetched due to race condition.


jay-hodgson (approved):
This all makes sense, thank you @emmanmills !

PR #912, opened 2021-02-16
PORTALS-1819

jay-hodgson:


src/lib/containers/query_wrapper_plot_nav/FilterAndView.tsx
```diff
@@ -19,37 +18,25 @@ const FilterAndView = (props: QueryWrapperChildProps & OwnProps) => {
...
-  } = props
+  } = props  
   const { showFacetFilter } = topLevelControlsState!
   return (
-    <div className="row">
-      <div
-        className={showFacetFilter ? 'col-xs-12 col-sm-3 col-lg-3' : 'hidden'}
-        style={{ paddingRight: '0px' }}
-      >
-        <QueryFilter {...rest} />
```
jay-hodgson, line 30: QueryFilter moved up to QueryWrapperPlotNav

src/lib/containers/query_wrapper_plot_nav/FilterAndView.tsx
```diff
@@ -19,37 +18,25 @@ const FilterAndView = (props: QueryWrapperChildProps & OwnProps) => {
     cardConfiguration,
     hideDownload,
     ...rest
-  } = props
+  } = props  
   const { showFacetFilter } = topLevelControlsState!
   return (
-    <div className="row">
```
jay-hodgson, line 25: Changed to css grid instead of bootstrap responsive layout

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -394,7 +395,7 @@ export default class SynapseTable extends React.Component<
               </div>
             )}
           {/* FRAGILE, CHANGE WITH CAUTION, see - https://sagebionetworks.jira.com/browse/PORTALS-1539 */}
```
jay-hodgson, line 397: It looks like the "fragile" part of this is around the table being in a div (so the table is always in the DOM).

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -85,7 +85,7 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
       )
       setIsFirstTime(false)
     }
-  }, [data, isFirstTime])
+  }, [data, isFirstTime, facetsToPlot])
```
jay-hodgson, line 88: fixed a warning (reference to facetsToPlot in the useEffect without declaring it here)

src/lib/containers/query_wrapper_plot_nav/TopLevelControls.tsx
```diff
@@ -128,79 +128,81 @@ const TopLevelControls = (
...
+  const showFacetFilter = topLevelControlsState?.showFacetFilter
   return (
-    <h3 className="QueryWrapperPlotNav__title">
-      <div className="QueryWrapperPlotNav__querycount">
-        <QueryCount token={token} name={name} sql={sql} parens={true} />
-      </div>
-      <div className="QueryWrapperPlotNav__actions">
-        {customControls &&
-          customControls.map(customControl => {
+    <div className={`TopLevelControls ${showFacetFilter ? QUERY_FILTERS_EXPANDED_CSS : QUERY_FILTERS_COLLAPSED_CSS}`}>
```
jay-hodgson, line 133: Highly recommend you view this diff in "split" view rather than "unified".  It looks like many changes when in "unified" mode, but in reality it just adds a wrapper div with "TopLevelControls" and the expand/collapse filters class.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -329,7 +328,7 @@ export default class SynapseTable extends React.Component<
       unitDescription,
       token,
       showBarChart,
-      enableLeftFacetFilter,
+      topLevelControlsState,
```
jay-hodgson, line 331: Special prop unnecessary, it has this information already from the QueryWrapperChildProps, so use it instead.

src/__tests__/lib/containers/SynapseTable.test.tsx
```diff
@@ -98,6 +98,14 @@ describe('basic functionality', () => {
     chartSelectionIndex: 0,
     isAllFilterSelectedForFacet: {},
     data: castData,
+    topLevelControlsState : {
```
jay-hodgson, line 101: @emmanmills : is there a better way to do this?  mock it using jest perhaps (instead of this real object)?

src/lib/style/components/_query-filter.scss
```diff
@@ -1,10 +1,11 @@
...
 .QueryFilter {
-  margin: 0rem 1rem 1rem 0;
-  border: 1px solid SRC.$border-color-gray;
-  padding: 1rem;
-
+  &__title {
+    font-weight: bold;
+  }
+  padding: 2rem;
+  background-color: rgba(47, 142, 148, 0.05);
```
emmanmills, line 8: Do you know if you are supposed to use the lighten version of the theme color here?

src/__tests__/lib/containers/SynapseTable.test.tsx
emmanmills: I checked Jest's documentation and it has something called jest.createMockFromModule(moduleName). Is this what you are thinking about? However I am not familiar with the usage.

src/lib/style/components/_query-filter.scss
jay-hodgson: Nice catch, I'll look into that.

src/__tests__/lib/containers/SynapseTable.test.tsx
jay-hodgson: Ok, I'll keep them as stub/real objects for now

src/lib/style/components/_query-filter.scss
jay-hodgson: Updated!

src/lib/style/components/_query-filter.scss
```diff
@@ -1,10 +1,15 @@
...
 .QueryFilter {
-  margin: 0rem 1rem 1rem 0;
-  border: 1px solid SRC.$border-color-gray;
-  padding: 1rem;
-
+  &__title {
+    font-weight: bold;
+  }
+  padding: 2rem;
+  background-color: adjust-color(
```
jay-hodgson, line 8: very light themed primary action color

emmanmills: Ah, I use `background-color: rgba(SRC.$primary-action-color-500, 0.1);`. But this should also work!

PR #915, opened 2021-02-17
PORTALS-1819: followup: adjusted theme color does not work across all portals.

jay-hodgson:


jay-hodgson:
What worked for some primary-action-color portals definitions, did not work for others (sometimes it resulted in white, and other times a loud teal).

jay-hodgson:
@emmanmills recommended another approach (to just reduce the opacity of the 500 variant), which worked like a charm in my testing (amp-ad, nf, and cancer complexity).

PR #918, opened 2021-02-18
Portals-1828: Add data type icons to cards

emmanmills:
Screen Shot 2021-02-17 at 5 03 09 PM


src/lib/containers/IconList.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import IconSvg, { IconSvgOptions } from './IconSvg'
+
+export type IconListProps = {
+  iconConfigs: {
+    [index:string]: IconSvgOptions
+  }
+  iconNames: string
```
jay-hodgson, line 8: I think this thing should have a consistent name.  It's called `dataType` in the GenericCardSchema, `dataTypeIconNames` for the value variable, and `iconNames` here.
Also, can you describe the expected interface/type (on JSON.parse).

src/lib/containers/GenericCard.tsx
```diff
@@ -33,6 +34,7 @@ export type GenericCardSchema = {
   imageFileHandleColumnName?:string
   secondaryLabels?: any[]
   link?: string
+  dataType?: string
```
jay-hodgson, line 37: Is there an example of using this new prop, in the demo app or in a unit test?

src/lib/containers/IconList.tsx
emmanmills: I standardized the names to use `dataTypeIconNames` in GenericCardSchema and value variable inside the Generic Card component. But I think `iconNames` should stay the same for the IconList component because I think the icons in the list are not necessary just for data types. So I think a more generic name like "iconNames" would be more appropriate. What do you think?

`JSON.parse` has been moved to the calling GenericCard component to avoid parsing error in the demo app. So the expected type for `iconNames` is `string[]`

src/lib/containers/GenericCard.tsx
emmanmills: Yes, just added.

jay-hodgson (approved):

PR #922, opened 2021-02-19
Precheck view and note the programmatic req

nickgros:
Appearance in Synapse.org: 

image


jay-hodgson:
I think the descriptions of each permission in parenthesis is suboptimal.  Would rather see them in subdued text.  This also solves the parens inside parens, where it starts to look like lisp :)
This comment does not block the merge.  Here's an example of what I'm thinking of though...
https://forum.gitlab.com/t/question-scopes-which-scopes-required-for-project-exporting/29542

jay-hodgson (approved):

nickgros:
I like your suggestion! Here's the new version:

image

I also cleaned up some of the CSS and refactored a tiny amount of component behavior by introducing useListView (which wasn't added when this component was created)

jay-hodgson (approved):
Looks great!

PR #927, opened 2021-02-24
PORTALS-1853

jay-hodgson:
Header version fo the Generic Card is no longer in a container-fluid.
And set a few max-widths


jay-hodgson:
Stockard had a chance to review the output, and I think we've reached consensus regarding the output.  Merging.

PR #928, opened 2021-02-25
Portals-1830: Fix view more button logic in cards

emmanmills:


src/lib/containers/CardContainer.tsx
```diff
@@ -40,18 +40,53 @@ export type CardContainerProps = {
...
    *
    * @memberof SynapseTable
    */
   const handleViewMore = () => {
-    const queryRequest = props.getLastQueryRequest!()
+    const queryRequest = getLastQueryRequest!()
     let offset = queryRequest.query.offset!
     // paginate forward
-    offset += PAGE_SIZE
+    offset += SynapseConstants.PAGE_SIZE
```
jay-hodgson, line 86: New offset ignores limit?  Expected it to use the configurable limit to figure out what the new offset of the query should be.

src/lib/containers/CardContainer.tsx
```diff
@@ -111,14 +131,11 @@ export const CardContainer = (props: CardContainerProps) => {
...
-  //   2. The hasMoreData prop is false
-  //   3. The limit is set to less than PAGE_SIZE
-  // below we show the view more button by following the opposite logic from above.
-  let showViewMore: boolean =
-    limit >= PAGE_SIZE && data.queryResult.queryResults.rows.length >= PAGE_SIZE
-  showViewMore = showViewMore && props.hasMoreData!
+  // limit is default 25 if it's not set in portal config
+  // pageCount increments by 1 whenever the view more button is clicked
+  const visibleCardCount = pageCount === 1 ? limit : limit * pageCount
+  let showViewMore: boolean = data.queryResult.queryResults.rows.length > visibleCardCount
```
jay-hodgson, line 137: queryResults.rows are the rows returned by the current page, which I'd expect to always be the limit (unless you're on the last page).  So I'd show the View More button if hasMoreData, which I think is determined by the Query Wrapper.
The problem is that there are so many references to SynapseConstants.PAGE_SIZE above this component.  If limit is really configurable, then I'd expect
`hasMoreData = data.queryResult.queryResults.rows.length === limit` rather than using the hard-coded PAGE_SIZE.
And at the end of the day, there may be cases where portal owners don't want to show the cards past the limit, so maybe we should have a way to force us to hide the View More.

PR #929, opened 2021-02-25
PORTALS-1854

jay-hodgson:
Expand descriptionLinkConfig to define if the description should be expanded by default (shown in full), in addition to whether it should be processed as Markdown.
Also add "thumbnailRequiresPadding", which will pad the icon or image by 21px (matches padding-top of card type UI element).

### DIFF
--- a/src/lib/containers/GenericCard.tsx
+++ b/src/lib/containers/GenericCard.tsx
@@ -5,7 +5,7 @@ import {
   CardLink,
   CommonCardProps,
   MarkdownLink,
-  MarkdownValue,
+  DescriptionConfig,
 } from './CardContainerLogic'
 import { unCamelCase } from '../utils/functions/unCamelCase'
 import MarkdownSynapse from './MarkdownSynapse'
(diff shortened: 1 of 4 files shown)


src/lib/containers/GenericCard.tsx
```diff
@@ -421,9 +422,9 @@ export default class GenericCard extends React.Component<
       marginTop: isHeader ? '0px' : undefined,
       marginBottom: isHeader ? '0px' : undefined,
       paddingBottom: showFooter || imageFileHandleIdValue ? undefined : '15px',
-    }
+    }    
     const icon:JSX.Element = <>
-        {imageFileHandleIdValue && <div className="SRC-imageThumbnail">
+        {imageFileHandleIdValue && <div className="SRC-imageThumbnail" style={{padding: genericCardSchemaDefined.thumbnailRequiresPadding ? '21px' : undefined}}>
```
emmanmills, line 427: Although this doesn't break anything, I wonder if setting padding to `undefined` is valid css. Should you just set it to 0 if no extra padding is needed?

emmanmills (approved):

PR #931, opened 2021-02-26
PORTALS-1856: set SearchV2 z-index, and fix custom image padding right bug

jay-hodgson:


src/lib/containers/GenericCard.tsx
```diff
@@ -424,7 +424,7 @@ export default class GenericCard extends React.Component<
       paddingBottom: showFooter || imageFileHandleIdValue ? undefined : '15px',
     }    
     const icon:JSX.Element = <>
-        {imageFileHandleIdValue && <div className="SRC-imageThumbnail" style={{padding: genericCardSchemaDefined.thumbnailRequiresPadding ? '21px' : 'inherit'}}>
+        {imageFileHandleIdValue && <div className="SRC-imageThumbnail" style={{padding: genericCardSchemaDefined.thumbnailRequiresPadding ? '21px' : undefined}}>
```
jay-hodgson, line 427: Setting to an undefined value causes the style not to be overridden (which is the intended behavior in this case)

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.md
```diff
@@ -54,13 +54,13 @@ Configured to show a table:
       ],
     }}
     searchConfiguration={{
-      searchable: ['assay','name','consortium'],
+      searchable: ['assay','name','consortium', 'study', 'dataType', 'tissue'],
```
jay-hodgson, line 57: Changes to the styleguidist demo to cause PORTALS-1856

src/lib/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -95,6 +95,11 @@
     }
   }
 
+  .SearchV2 {
+    // PORTALS-1856: z-index greater than the Show All Graphs button (500)
+    z-index: 600;
```
jay-hodgson, line 100: the fix

emmanmills (approved):
Looks good!

PR #936, opened 2021-03-01
SWC-5388: Replace column value with table entity file download link

emmanmills:
Screen Shot 2021-03-01 at 2 44 45 PM

### DIFF
--- a/src/lib/containers/DirectDownload.tsx
+++ b/src/lib/containers/DirectDownload.tsx
@@ -1,24 +1,37 @@
 import React, { useEffect, useState } from 'react'
 import { Icon } from './row_renderers/utils'
 import {
+  BatchFileRequest,
   FileEntity,
+  FileHandle,
+  FileHandleAssociateType,
+  FileHandleAssociation,
   implementsExternalFileHandleInterface,
 } from '../utils/synapseTypes'
-import { SynapseClient } from '../utils'
+import {
+  getFiles,
+  getFileResult,
+  getEntity
+} from '../utils/SynapseClient'
+import IconSvg from './IconSvg'
 
 export type DirectFileDownloadProps = {
   token?: string
   associatedObjectId: string
   entityVersionNumber?: string
+  associatedObjectType?: FileHandleAssociateType
+  fileHandleId?: string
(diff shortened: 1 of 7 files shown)


src/lib/containers/DirectDownload.tsx
```diff
@@ -30,35 +43,73 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = (props)
...
+    const fileHandleAssociationList: FileHandleAssociation[] = [
+      {
+        associateObjectId: associatedObjectId!,
+        associateObjectType: associatedObjectType!,
+        fileHandleId: fileHandleId!,
+      },
+    ]
+    const batchFileRequest: BatchFileRequest = {
+      includeFileHandles: true,
+      includePreSignedURLs: true,
```
jay-hodgson, line 91: Can includePreSignedURLs be set to false here?

src/lib/containers/DirectDownload.tsx
```diff
@@ -1,24 +1,37 @@
...
-  const {token, associatedObjectId, entityVersionNumber} = props
+  const {token, associatedObjectId, entityVersionNumber, associatedObjectType, fileHandleId, displayFileName} = props
   const [isExternalFile, setIsExternalFile] = useState<boolean>(false)
   const [hasFileAccess, setHasFileAccess] = useState<boolean>(false)
   const [fileEntity, setFileEntity] = useState<FileEntity>()
   const [externalURL, setExternalURL] = useState<string>()
+  const [fileName, setFileName] = useState<string>("")
   let mounted:boolean = true
 
   useEffect( () => {
```
jay-hodgson, line 37: This could be modified to execute the effect if inView changes (and only getFileEntityFileHandle() if mounted and inView are true)

emmanmills: Done.

src/lib/containers/DirectDownload.tsx
emmanmills: Updated. Thanks.

jay-hodgson (approved):

PR #939, opened 2021-03-02
PORTALS-1860: useGetQueryResultBundle update to use `react-query`

nickgros:
Using the `react-query` library here lets us remove a lot of code, since we've constructed an API that is very similar to the `useQuery` API.

It's important to note here that to use the `useQuery` hook, context must be available via an ancestor `QueryClientProvider`. To prevent massively breaking these components that are now using `useQuery`, I wrote a wrapper that will provide each component a `QueryClient`.

However, the performance benefits here may end up being negligible because the `QueryClientProvider` may unmount with the component, losing the entire cache. So when the component re-mounts, the cache is empty, so you don't get to see the component pre-load with stale data.

If we wanted to remove the wrapper (to get the most gains out of `react-query`), we would need to ensure that dependents of SRC wrap these components in a `QueryClientProvider`. This is easy in portals, since it would go towards the top level of the app. It would be harder in SWC, which renders components piecewise. Regardless, I don't like the idea of SRC components breaking in other apps because of a missing Context tag that seemingly has nothing to do with SRC, so we'll need to come up with a new approach.

I also added many tests for the other hooks created for EntityFinder.


src/__tests__/lib/utils/hooks/SynapseAPI/QueryClientProviderWrapper.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+  })
+
+  it('An application that uses useQuery fails to render without the wrapper', async () => {
+    // Mocking console.error not necessary for the test, we just expect it to happen and want to suppress
+    jest.spyOn(console, 'error').mockImplementation(() => {})
+
+    try {
+      expect(mount(<BasicComponentWithUseQuery />)).toThrow(
+        expect.stringContaining(
+          'No QueryClient set, use QueryClientProvider to set one',
```
jay-hodgson, line 27: Is looking for this exact substring going to make this test fragile (moving forward)?

jay-hodgson (approved):

src/__tests__/lib/utils/hooks/SynapseAPI/QueryClientProviderWrapper.test.tsx
nickgros: Probably will be more fragile than necessary. I think since we test both the error and non-error conditions, it's unnecessary to check the message.

PR #940, opened 2021-03-03
SWC-5398: Entity Finder Testing/Cleanup -- DetailsView

nickgros:


src/lib/containers/entity_finder/details/view/DetailsViewRow.tsx
```diff
@@ -81,22 +87,31 @@ export const DetailsViewRow: React.FunctionComponent<DetailsViewRowProps> = ({
...
     <tr
       ref={ref}
-      className={`EntityFinderDetailsView__Row${
-        isSelected ? ' EntityFinderDetailsView__Row__Selected' : ''
-      }${isDisabled ? ' EntityFinderDetailsView__Row__Disabled' : ''}`}
-      style={isHidden ? { display: 'none' } : {}}
+      aria-selected={isSelected}
+      aria-disabled={isDisabled}
+      aria-hidden={isHidden}
+      className="EntityFinderDetailsView__Row"
```
nickgros, line 114: React Testing Library intentionally makes it difficult to inspect the DOM (e.g. validate CSS classes) and instead requires testing the accessibility tree. Thus we use ARIA attributes instead of custom classes. I think it ends up being neater as well.

jay-hodgson: I agree!  This is something we should work towards in the rest of the codebase.

src/lib/containers/entity_finder/details/view/DetailsViewRow.tsx
```diff
@@ -110,87 +125,98 @@ export const DetailsViewRow: React.FunctionComponent<DetailsViewRowProps> = ({
...
-          {!isDisabled && selectButtonType === 'radio' && (
-            <RadioGroup
-              className="SRC-pointer-events-none"
-              options={[{ label: '', value: 'true' }]}
-              value={isSelected.toString()}
-              id=""
-              onChange={() => {}}
-            />
-          )}
+        <td className="IsSelectedColumn" aria-label="is-selected">
```
jay-hodgson, line 128: Should we be moving the scss in this direction too (to use the aria attributes)?
```
[aria-label="is-selected"] {
  //styles
}
```

src/lib/style/components/_entity-finder.scss
```diff
@@ -160,24 +181,23 @@ $-finder-height: 500px;
...
-  &__Row:hover:nth-child(even) {
+  &__Row:hover {
     background: color.change(SRC.$primary-action-color-500, $alpha: 0.2);
   }
 
-  &__Row__Selected,
-  &__Row__Selected:hover,
-  &__Row__Selected:nth-child(even),
-  &__Row__Selected:hover:nth-child(even) {
+  &__Row[aria-selected='true'] {
```
jay-hodgson, line 188: Oh, I guess you already did for some of these definitions!

src/__tests__/lib/containers/entity_finder/details/DetailsView.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+        expect(mockDetailsViewRow).toHaveBeenNthCalledWith(
+          2,
+          expect.objectContaining({ appearance: 'default' }),
+          {},
+        )
+      })
+
+      it('Creates a row with the selected appearance', async () => {
+        renderScreen({
+          selected: [{ targetId: entityHeaders[0].id }], // !
```
jay-hodgson, line 117: Like the "// !" to draw attention to the perturbation

jay-hodgson:
Like the error handling, type solidification, and new tests.  Nice work!

jay-hodgson (approved):

src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -191,141 +205,144 @@ export const EntityFinder: React.FunctionComponent<EntityFinderProps> = ({
...
-                      {configFromTreeView && (
-                        <EntityDetailsList
+          )}
+          {
+            <div style={searchActive ? { display: 'none' } : {}}>
+              <div className="EntityViewReflexContainer">
+                <SizeMe>
+                  {({ size }) => (
+                    <ReflexContainer
+                      key={(!!size.width).toString()}
```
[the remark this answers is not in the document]
nickgros: I don't think I need windowResizeAware, but I need to look into this further. I force the rerender with SizeMe to try to prevent a bug where the ReflexContainer can't determine its parent's height or width, which causes it to render a blank component. This happens often in SWC locally...I think partially because it loads in a modal but I'm not totally sure. It's been a very difficult issue to fix reliably.

PR #948, opened 2021-03-09
PORTALS-1410: show a special UI when the user search returns zero rows

jay-hodgson:


src/lib/containers/CardContainer.tsx
```diff
@@ -96,7 +97,11 @@ export const CardContainer = (props: CardContainerProps) => {
     return <div>{isLoading && loadingScreen}</div>
   } else if (data && data.queryResult.queryResults.rows.length === 0) {
     // data was retrieved from the backend but there is none to show.
-    // show "no results" UI (see PORTALS-1497)
+    const queryRequest = props.getLastQueryRequest!()
+    if (queryRequest.query.additionalFilters) {
+      return <SearchResultsNotFound />
```
jay-hodgson, line 102: Used in cards when additionalFilters is set

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -89,7 +90,6 @@ export type SynapseTableState = {
   mapEntityIdToHeader: Dictionary<EntityHeader>
   mapUserIdToHeader: Dictionary<Partial<UserGroupHeader & UserProfile>>
   isColumnSelectionOpen: boolean
-  isUserModifiedQuery?: boolean //flag to signal that the selection criterial has been defined by user and if no records are returned do not hide the table
```
jay-hodgson, line 92: This was trying to track if a user action caused it to result in zero rows, but I could not find a way to do this.  Maybe if you quickly select facet values (across facet keys)?  Facet values that would result in zero rows are not returned by the service.

PR #950, opened 2021-03-10
Portals-1490: Render entity names instead of ids

emmanmills:
Screen Shot 2021-03-10 at 2 00 08 PM

### DIFF
--- a/src/__tests__/lib/containers/SynapseTable.test.tsx
+++ b/src/__tests__/lib/containers/SynapseTable.test.tsx
@@ -352,6 +352,7 @@ describe('basic functionality', () => {
     const BOOLEAN_LIST_INDEX = 5
     const INTEGER_LIST_INDEX = 6
     const FILEHANDLEID_INDEX = 7
+    const ENTITYIDLIST_INDEX = 8
     const MOCKED_STRING = 'MOCKED_VALUE'
     const MOCKED_STRING_LIST = '["MOCKED_VALUE1","MOCKED_VALUE2"]'
     const MOCK_DATE_VALUE = '1581360939000'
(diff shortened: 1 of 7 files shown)


src/__tests__/lib/containers/SynapseTable.test.tsx
```diff
@@ -606,27 +608,33 @@ describe('basic functionality', () => {
...
           <div>
             {renderTableCell({
-              entityColumnIndicies,
-              userColumnIndicies,
-              dateColumnIndicies,
-              booleanListColumnIndicies,
-              fileHandleIdColumnIndicies,
-              dateListColumnIndicies,
-              otherListColumnIndicies,
+              ...tableCellProps,
```
emmanmills, line 634: The table cell props were duplicated in several places so I refactored it.

jay-hodgson: Much cleaner, thank you.

src/lib/utils/SynapseClient.ts
```diff
@@ -827,6 +827,22 @@ export const getEntity: GetEntity = <T>(
   ) as Promise<T>
 }
 
+/**
+ * Get a list of entity headers given by entity ids
+ * http://rest-docs.synapse.org/rest/GET/entity/type.html
+ */
+export const getEntityTypeByIds = <T extends PaginatedResults<EntityHeader>> (
```
jay-hodgson, line 834: These are not just the entity types that are being returned.  Could you rename to getEntityHeadersByIds?

src/lib/containers/EntityIdList.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (mounted && inView) {
+      getEntityTypes()
+    }
+    return () => {
+      mounted = false
+    }
+  }, [entityIdList, inView])
+
+  const getEntityTypes = async () => {
+    const entityIds = entityIdList.join(",")
```
jay-hodgson, line 28: Prefer this transform (de-structuring) to occur in the getEntityHeadersByIds.  Could you pass the entityIdList array to that function?

src/lib/utils/SynapseClient.ts
emmanmills: This is done.

src/lib/containers/EntityIdList.tsx
emmanmills: This is done.

PR #961, opened 2021-03-25
PORTALS-1857

jay-hodgson:


src/__tests__/lib/containers/GenericCard.test.tsx
```diff
@@ -263,7 +263,7 @@ describe('it makes the correct URL for the title', () => {
   it('creates a link to synapse', () => {
     const synId = 'syn12345678'
     const synLink = `https://www.synapse.org/#!Synapse:${synId}`
-    const { href, target } = instance.getTitleParams(
+    const { href, target } = instance.getLinkParams(
```
jay-hodgson, line 266: Renamed (use same function for CTA link and card title)

src/lib/containers/CardContainerLogic.tsx
```diff
@@ -32,9 +32,11 @@ export type MarkdownLink = {
   matchColumnName: string
 }
 
-export type CTAButtonCardLink = {
-  buttonText: string
-  linkConfig: CardLink
+export type CTACardLink = {
```
jay-hodgson, line 35: Remove "Button" from name, since it's no longer a button.  Simplify so that it only supports link text and a table/view column that contains the URLs (or Synapse IDs!)

src/lib/containers/GenericCard.tsx
```diff
@@ -469,7 +468,19 @@ export default class GenericCard extends React.Component<
...
+    if (ctaLinkConfig) {
+      const ctaLinkValue: string = data[schema[ctaLinkConfig.link]]
+      const { href: newCtaHref, target: newCtaTarget } = this.getLinkParams(
+        ctaLinkValue,
+        undefined, //card link config
+        data,
+        schema,
+      )
+      ctaHref = newCtaHref
+      ctaTarget = newCtaTarget
```
jay-hodgson, line 482: Is there a better way of doing this?

emmanmills: I am wondering if you want to set `ctaLinkValue` outside the `if (ctaLinkConfig)` block because I am guessing whether you have the `ctaLinkConfig`, you still want to surfac the `ctaLinkValue` right? Also, what happen if `data[schema[ctaLinkConfig.link]]` is not defined?  What do you see there?

jay-hodgson: Without a ctaLinkConfig I don't know where to look for the ctaLinkValue.
Not sure what happens when data[schema[ctaLinkConfig.link]] is undefined, I'll test and make sure that it doesn't render a link in that case!

jay-hodgson: Changed to fallback to empty string (like so many other cases in this file), which results in the link not rendering.

emmanmills (approved):
Looks good. Thanks!

PR #970, opened 2021-04-12
new StyleGuidist component wrapper

jay-hodgson:
introduce a memoryrouter for components that rely on the router environment (like Prompt, Switch, Link, and others)


conundrumgirl (approved):
Thank you!

PR #975, opened 2021-04-16
PORTALS-1905: show request button regardless logging in or not

emmanmills:
Screen Shot 2021-04-16 at 1 55 51 PM

### DIFF
--- a/src/lib/containers/access_requirement_list/AcceptedRequirements.tsx
+++ b/src/lib/containers/access_requirement_list/AcceptedRequirements.tsx
@@ -29,7 +29,7 @@ export type AcceptedRequirementsProps = {
     | ManagedACTAccessRequirement
   accessRequirementStatus: AccessRequirementStatus | undefined
   showButton?: boolean
-  showRequestAccessButton?: boolean
+  isRequestAccessButton?: boolean
   onHide?: Function
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/containers/access_requirement_list/AcceptedRequirements.tsx
```diff
@@ -183,7 +185,23 @@ export default function AcceptedRequirements({
           )}
         </div>
       </div>
-      {token && showButton && (
+      { // Show request access button for ManagedACTAccessRequirement component without logged in
+        isRequestAccessButton && showButton && (
+        <div className={`button-container ${isApproved ? `hide` : `default`}`}>
+          <div className="accept-button-container">
+            <button className="accept-button" onClick={onAcceptClicked}>
```
jay-hodgson, line 192: If anonymous (no token), I'd expect onAcceptClicked to pop up a message instead sending to Synapse.org (or opening up your new wizard).

jay-hodgson: Way to avoid duplicating the UI code?  (isRequestAccessButton || token) && showButton && ...

emmanmills: I am unclear whether it's okay to keep the current workflow to redirect to Synapse.org and over there, user will be asked to first sign in. If  the workflow needs to stay in portals, then the changes should probably go with the new feature branch.

emmanmills:
Fix in new feature branch instead (ref: https://sagebionetworks.jira.com/browse/PORTALS-1840)

PR #976, opened 2021-04-16
PORTALS-1906: show the download confirmation UI in the correct location and layout

jay-hodgson:


emmanmills (approved):
Looks good to me!

PR #985, opened 2021-04-19
small fixes for spacing (and making the box shadow visible)

jay-hodgson:


jay-hodgson:
Before (now that items are forced into grid boundary!):
Screen Shot 2021-04-19 at 12 48 37 PM

After:
Screen Shot 2021-04-19 at 12 51 10 PM

jay-hodgson:
Current production looks like this (where items are not forced into the css grid):
Screen Shot 2021-04-19 at 12 54 54 PM

src/lib/style/components/_cards-rotate.scss
```diff
@@ -12,17 +10,8 @@ $card-rotate-margin: 21px;
...
       overflow: hidden;
-    }
-
-    & > .SRC-grid-item:first-child {
-      margin-right: $card-rotate-margin;
-    }
-
-    & > .SRC-grid-item:last-child {
-      margin-left: $card-rotate-margin;
+      padding: 5px;
```
emmanmills, line 14: Another way to force equal width for all 3 columns, maybe use something like `grid-template-columns: repeat(3, minmax(0, 1fr));`?

jay-hodgson: I think we've have the same box-shadow issue if we changed to use your grid definition.

emmanmills: The 42px spacing is from the original design, so instead we can apply to the padding instead of the margin withhin each column.

emmanmills: Ah, okay. Thanks.

emmanmills (approved):

PR #994, opened 2021-04-26
PORTALS-1912: if image scale is set to 100, set to auto.  add katex preprocessing (…

jay-hodgson:
…same as Synapse.org).  Also, only process the math element in the DOM once.

### DIFF
--- a/src/lib/containers/MarkdownSynapse.tsx
+++ b/src/lib/containers/MarkdownSynapse.tsx
@@ -262,13 +262,18 @@ export default class MarkdownSynapse extends React.Component<
       '[id^="mathjax-"]',
     )
     // go through all obtained elements and transform them with katex
+    const regEx = new RegExp(/\\[()[\]]/, 'g') // Look for a '\' followed by either '(', ')', '[', or ']'. We delete these strings since they interfere with katex processing.
     mathExpressions.forEach(element => {
-      element.textContent &&
-        katex.render(element.textContent, element, {
+      if (element.textContent && !element.getAttribute('processed')) {
+        // only process a math element once, used to double/triple process
+        element.setAttribute('processed', 'true')
+        const textContent = element.textContent.replace(regEx, '')
+        return katex.render(textContent, element, {
           // @ts-ignore
           output: 'html',
           throwOnError: false,
         })
+      }
     })
   }
   /**
--- a/src/lib/containers/widgets/SynapseImage.tsx
+++ b/src/lib/containers/widgets/SynapseImage.tsx
@@ -108,7 +108,12 @@ class SynapseImage extends React.Component<
   public render() {
     const { params } = this.props
     const { align = '', altText = 'synapse image' } = params
-    let scale = `${Number(params.scale) ?? 100}%`
+
+    let scale = 'auto'
+    if (params.scale && params.scale !== '100') {
+      scale = `${Number(params.scale)}%`
+    }
+    
     const alignLowerCase = align.toLowerCase()
     let className = ''
     if (alignLowerCase === 'left') {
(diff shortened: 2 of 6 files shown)


jay-hodgson:
Here's the Hackthon wiki page after the changes (with fixed inline math processing, and showing the default image scale):
Screen Shot 2021-04-26 at 4 01 45 PM

src/lib/utils/SynapseClient.ts
```diff
@@ -1114,6 +1114,11 @@ export const getPresignedUrlForWikiAttachment = (
     BackendDestinationEnum.REPO_ENDPOINT,
   )
 }
+
+export const isInSynapseExperimentalMode = ():boolean => {
```
jay-hodgson, line 1118: Also, add a way to ask if Synapse Experimental Mode is on (if the cookie exists).  Will use this to start to use the new Download List services.

styleguide.config.js
```diff
@@ -102,15 +102,21 @@ module.exports = {
             {
                 src:"https://cdn.jsdelivr.net/npm/sanitize-html@1.20.0/dist/sanitize-html.min.js",
             },
+            {
+                src:"https://cdn.jsdelivr.net/npm/katex@0.10.0/dist/katex.min.js",
```
jay-hodgson, line 106: Load katex in the StyleGuide

styleguide.config.js
```diff
@@ -102,15 +102,21 @@ module.exports = {
...
         links: [
             {
               rel: 'stylesheet',
-              href:
-                'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css'
+              href: 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css'
+            },
+            {
+                rel: 'stylesheet',
+                href: 'https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/katex.min.css'
```
jay-hodgson, line 116: And load the Katex css

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -262,13 +262,18 @@ export default class MarkdownSynapse extends React.Component<
       '[id^="mathjax-"]',
     )
     // go through all obtained elements and transform them with katex
+    const regEx = new RegExp(/\\[()[\]]/, 'g') // Look for a '\' followed by either '(', ')', '[', or ']'. We delete these strings since they interfere with katex processing.
```
jay-hodgson, line 265: This syntax is a way to indicate that the math should be display inline, but we handle this outside of the latex (in the markdown it's either an inline math element or a block element).
https://www.overleaf.com/learn/latex/mathematical_expressions

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -262,13 +262,18 @@ export default class MarkdownSynapse extends React.Component<
...
     mathExpressions.forEach(element => {
-      element.textContent &&
-        katex.render(element.textContent, element, {
+      if (element.textContent && !element.getAttribute('processed')) {
+        // only process a math element once, used to double/triple process
+        element.setAttribute('processed', 'true')
```
jay-hodgson, line 269: Only run the element content through katex processing once.

src/lib/containers/widgets/SynapseImage.tsx
```diff
@@ -108,7 +108,12 @@ class SynapseImage extends React.Component<
   public render() {
     const { params } = this.props
     const { align = '', altText = 'synapse image' } = params
-    let scale = `${Number(params.scale) ?? 100}%`
+
+    let scale = 'auto'
+    if (params.scale && params.scale !== '100') {
+      scale = `${Number(params.scale)}%`
```
jay-hodgson, line 114: Only use the scale parameter if it's set and not '100'.  Default value is 'auto'

src/lib/utils/SynapseClient.ts
emmanmills: This is a nice addon, we can use this for other feature testing too.

emmanmills (approved):
Looks good. Thanks.

PR #996, opened 2021-04-27
PORTALS-1896

nickgros:


src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -108,58 +102,76 @@ export function extractPlotDataArray(
...
+        sessionToken,
+      )
+      for (const header of response.results) {
+        map.set(header.id, header.name)
+      }
+    } else if (columnType === ColumnType.USERID) {
+      const response = await SynapseClient.getGroupHeadersBatch(
+        filteredValues,
+        sessionToken,
+      )
```
nickgros, line 132: Both of these requests return the deprecated `PaginatedList` response that possibly give an inaccurate `totalNumberOfResults` based on offset, but the offset is unknown because failures are omitted. There's no offset parameter on the service, so we can't just get the next page.

To get all results reliably, we may have to do our own reconciliation where we double-request entities to see if they were omitted because of a failure or because of pagination.

I don't think it's a huge deal to not paginate here since we make requests in descending order by count (so the largest pie slices will be requested first).

[Filed a backend ticket here](https://sagebionetworks.jira.com/browse/PLFM-6714).

jay-hodgson:
What's the current behavior when we have many unique facet values for an EntityID or UserID type column (so many that they cannot be resolved to populate your label map)?

jay-hodgson (approved):

PR #997, opened 2021-04-27
hints for treeshaking (material ui icons) reduces portal build by .7 MB

jay-hodgson:


package.json
```diff
@@ -62,7 +62,7 @@
     "markdown-it-synapse-table": "^1.0.6",
     "moment": "^2.24.0",
     "moment-timezone": "^0.5.32",
-    "plotly.js-basic-dist": "^1.54.7",
+    "plotly.js-basic-dist-min": "^1.58.4",
```
jay-hodgson, line 65: this ends up having no effect on portals build size, but I left it in since I think we should probably be importing the min dist

PR #1000, opened 2021-04-29
PORTALS-1913: add special renderer for UserId_List type columns (and UserId columns)

jay-hodgson:


jay-hodgson:
If the View/Table column type is UserID_List, then we automatically render small user cards:
Screen Shot 2021-04-29 at 10 58 39 AM

src/lib/containers/GenericCard.tsx
```diff
@@ -83,7 +85,7 @@ export const getCutoff = (summary: string) => {
   return { previewText }
 }
 
-export const renderValueOrMultiValue = ({
+export const getValueOrMultiValue = ({
```
jay-hodgson, line 88: And renamed this function (since we are not actually rendering anything in this function, but rather parsing out the values to render).  Also returns columnModelType (so caller can decide if it wants to use a specialized renderer for the returned value)

nickgros (approved):

PR #1003, opened 2021-04-30
PORTALS-1821

nickgros:
Depends on #1002 

image

After clicking View All:

image

Once the viewport narrows enough, flex-grow causes the graphs expand to fill the width
image

Kind of weird side effect of flex grow, if there's an odd number of graphs:
image

### DIFF
--- a/src/lib/style/components/facet_nav/_facet-nav.scss
+++ b/src/lib/style/components/facet_nav/_facet-nav.scss
@@ -12,9 +12,9 @@
     justify-content: space-between;
 
     > div {
-      margin: 5px;
-      flex-grow: 0;
-      flex-basis: 48%;
+      margin: 3px;
+      flex-grow: 1;
+      flex-basis: 49%;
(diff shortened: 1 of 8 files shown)


src/lib/style/components/facet_nav/_facet-nav.scss
```diff
@@ -2,43 +2,32 @@
...
-
-.container-full-width .FacetNav.less{
-  .FacetNav {
-    &__row {
-      display: flex;
-      flex-wrap: wrap;
 
-      > div {
-        flex: 0 0 380px;
-      }
```
nickgros, line 41: Always using flex, so I don't think we need this anymore

jay-hodgson: Ok

jay-hodgson:
Has many (all?) of the changes from #1002 .  Why does the build fail?

PR #1007, opened 2021-05-03
PORTALS-1830: For pagination, use the request limit (instead of a hard-coded page size)

jay-hodgson:


emmanmills (approved):
Looks good. Thanks!

PR #1010, opened 2021-05-04
PORTALS-1911: in alpha mode only, expose a column where a single item…

jay-hodgson:
… can be added to the Download List v2 (cart)

### DIFF
--- a/src/lib/containers/AddToDownloadListV2.tsx
+++ b/src/lib/containers/AddToDownloadListV2.tsx
@@ -1,0 +1,40 @@
+import React from 'react'
+import {
+  AddBatchOfFilesToDownloadListResponse,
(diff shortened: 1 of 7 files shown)


jay-hodgson:
When the SynapseTestWebsite cookie available and set to true, this PR adds a new "add to cart" column:
Screen Shot 2021-05-04 at 2 00 02 PM

And when the "add to cart" button is clicked, it tries to add that row to the new Download List via [this API](http://rest-docs.synapse.org/rest/POST/download/list/add.html).
Request:
Screen Shot 2021-05-04 at 2 00 59 PM
Response:
Screen Shot 2021-05-04 at 2 01 09 PM

emmanmills (approved):
Looks good! Thanks.

PR #1013, opened 2021-05-05
PORTALS-1917: fix search filter added by the client when the column w…

jay-hodgson:
…as found


emmanmills (commented):
Looks good! Thanks.

emmanmills (approved):

PR #1018, opened 2021-05-06
SWC-5589:  use Access Token instead of Session Token

jay-hodgson:
Mostly renaming, but these are the interesting bits:
[SynapseClient.ts](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1018/files#diff-9378feef6a76b985b375207d91c3f08213bafdcc17a7ecb7cb2a91f4c8c9cd3f)
[LoginResponse.ts](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1018/files#diff-bba6ad6e64fadc9563a5166c502fc1b1cfc4e8ad81dd08066bd2f0d075b7c9e8)


emmanmills (approved):
Looks good!

jay-hodgson:
Changes after review was about reverting newly introduced code (and updated a styleguidist md file, which should not block release).

PR #1019, opened 2021-05-06
Portals-1840: Support ManagedACTAccessRequirement full workflow in portals

emmanmills:


package.json
```diff
@@ -41,6 +41,7 @@
     "@upsetjs/react": "^1.6.2",
     "bootstrap": "^4.6.0",
     "column-resizer": "^1.3.5",
+    "downshift": "^6.1.2",
```
emmanmills, line 44: This package is for the UserSearchBox component to auto-suggest the accessors. https://github.com/downshift-js/downshift

src/lib/containers/UserCardSmall.tsx
```diff
@@ -75,6 +120,7 @@ export const UserCardSmall: React.FunctionComponent<UserCardSmallProps> = ({
       >
         {avatar}
         {`@${userProfile.userName}`}
+        {showAccountLevelIcon && <span className={"account-level-icon"}>{icon}</span>}
```
emmanmills, line 123: Add account level icon in small user card but only show it when `showAccountLevelIcon` is true.

src/lib/containers/access_requirement_list/AcceptedRequirements.tsx
```diff
@@ -181,23 +181,23 @@ export default function AcceptedRequirements({
...
+              <div className="not-accept-button-container">
+                <button className="not-accpet-button" onClick={() => onHide?.()}>
+                  I do not accept
+                </button>
+              </div>
+            </div>
+          )}
         </div>
-      )}
+      </div>
```
emmanmills, line 200: Move the markup to make the buttons left-aligned rather than center-aligned.

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -217,6 +236,24 @@ export default function AccessRequirementList({
...
+      setManagedACTAccessRequirement(managedACTAccessRequirement)
+    }
+    if (researchProjectId) {
+      setresearchProjectId(researchProjectId)
+    }
+    if (formSubmitRequestObject) {
+      setFormSubmitRequestObject(formSubmitRequestObject)
+    }
+    setRequestDataStep(step)
+  }
```
emmanmills, line 255: This function is called in the child components when we need to pass data among sibling components.

src/lib/containers/DirectDownloadButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  className?: string
+  variant?: string  // This allows you to change the look of the button (see react bootstrap doc)
+}
+
+const DirectDownloadButton: React.FC<DirectFileDownloadButtonProps> = props => {
+
+  const {id, token, fileHandleAssociation, className, variant, fileName}  = props
+  let mounted:boolean = true
+
+  useEffect( () => {
```
jay-hodgson, line 20: Does this do anything?

src/lib/containers/DirectDownloadButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }, [token])
+
+  const getDownloadLink = async () => {
+    if (!fileHandleAssociation.fileHandleId || !token) return
+
+    const batchFileRequest:BatchFileRequest = {
+      requestedFiles: [fileHandleAssociation],
+      includePreSignedURLs: true,
+      includeFileHandles: true,
```
jay-hodgson, line 35: I think includeFileHandles can be false in this case.

src/lib/containers/UserCardSmall.tsx
```diff
@@ -23,13 +28,53 @@ export const UserCardSmall: React.FunctionComponent<UserCardSmallProps> = ({
...
   withAvatar = false,
   avatarSize = 'SMALL',
   imageURL,
   ...rest
 }) => {
+
+  const [userBundle, setUserBundle] = useState<UserBundle | undefined>()
   const target = useRef(null)
+  
+  let icon = <Registered />
```
jay-hodgson, line 42: Thought icon would be a state variable.  Why don't we do this more often (just use a local variable that we update in the effect)?

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -217,6 +236,24 @@ export default function AccessRequirementList({
         return undefined
     }
   }
+
+  const requestDataStepCallback = (props:requestDataStepCallbackProps) => {
+    const {managedACTAccessRequirement, step, researchProjectId, formSubmitRequestObject} = props
+    if (managedACTAccessRequirement) {
+      // to identify which managedACTAccessRequirement we are looking at in step 1 & 2 forms
+      setAccessRequirementId(String(managedACTAccessRequirement.id))
```
jay-hodgson, line 244: This seems unnecessary if you're passing around the entire managedACTAccessRequirement.  Am I missing something?

src/lib/containers/DirectDownloadButton.tsx
emmanmills: No except when the token is updated. You think it's okay to delete it?

src/lib/containers/access_requirement_list/managedACTAccess/CancelRequestDataAccess.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const {formSubmitRequestObject, token, onHide} = props
+  const [alert, setAlert] = useState<AlertProps | undefined>()
+  const [showCloseBtn, setShowCloseBtn] = useState<boolean>(false)
+
+  const handleSave = async () => {
+    if (formSubmitRequestObject) {
+      try {
+        const resp = await updateDataAccessRequest(formSubmitRequestObject, token)
+        if (resp.etag) {  // save success, close dialog
```
jay-hodgson, line 25: Did you run into a case where the backend responded with a 200 but left etag undefined?  I did not expect to see this case (I would have thought that your catch would have been sufficient).

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
```diff
@@ -0,0 +1,298 @@
...
+      step: 1
+    })
+  }
+
+  const onAcceptClicked = async () => {
+    if (
+      accessRequirement.concreteType ===
+        SUPPORTED_ACCESS_REQUIREMENTS.ManagedACTAccessRequirement ||
+      accessRequirement.concreteType ===
+        SUPPORTED_ACCESS_REQUIREMENTS.ACTAccessRequirement
```
jay-hodgson, line 79: I don't think SUPPORTED_ACCESS_REQUIREMENTS.ACTAccessRequirement should be handled by any of your new code.  This type of Access Requirement does not have an access request in Synapse (it's handled out-of-band, typically either by email or Jira).

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
```diff
@@ -0,0 +1,298 @@
...
+          gotoSynapseAccessRequirementPage()
+        }
+      } else {
+        requestDataStepCallback?.({
+          managedACTAccessRequirement: accessRequirement,
+          step: 4
+        })
+      }
+
+      // PORTALS-1483: and close the dialog.
```
jay-hodgson, line 116: Just delete this code.  You're right that requesting access should not close the dialog (PORTALS-1483 is applicable when it sends the user off to request access in another window).

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+    setIsApproved(propsIsApproved)
+    if (accessRequirementStatus?.currentSubmissionStatus) {
+      setSubmissionState(accessRequirementStatus.currentSubmissionStatus.state)
+      showSubmissionStatusAlert(accessRequirementStatus.currentSubmissionStatus.state)
+    }
+
+  }, [propsIsApproved])
+
+  const gotoSynapseAccessRequirementPage = () => {
```
jay-hodgson, line 67: gotoSynapseAccessRequirementPage() could be renamed.  something like showRequestAccess()

src/lib/containers/DirectDownloadButton.tsx
emmanmills: Done.

src/lib/containers/UserCardSmall.tsx
emmanmills: Good idea. Created a state variable to store the icon.

src/lib/containers/UserSearchBox.tsx
```diff
@@ -0,0 +1,95 @@
...
+            role: 'searchbox'
+          })} style={{marginBottom: '0'}} />
+          <ul {...getMenuProps()} className={isOpen ? "users-visible" : ""} role='list'>
+            { isOpen ? users.filter((user:FormattedUserHeader) => !inputValue
+              || `${user.firstName} ${user.lastName}`.includes(inputValue)
+              || user.userName.includes(inputValue))
+                  .map((item, index) => (
+                    <li key={`userSearchBox-${index}`} {...getItemProps({key:item.ownerId, index, item})}>
+                      {<>
+                        <a
```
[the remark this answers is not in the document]
emmanmills: Sounds good to me. Changed to use the small UserCard.

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
emmanmills: I did this because managedACTAccessRequirement.id is a number type and all the APIs ask for a string type. So I create an extra variable just to avoid type conversion in several places. But for clarity's sake, I removed the redundant variable.

src/lib/containers/access_requirement_list/managedACTAccess/CancelRequestDataAccess.tsx
emmanmills: I haven't seen it yet. What would be a better way to check if saving is successful if etag may not be reliable?

src/lib/containers/DirectDownloadButton.tsx
jay-hodgson: When the token is updated, what does this do?

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep1.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    try {
+      updateResearchProject(requestObj, token).then((researchProject) => {
+        requestDataStepCallback?.({
+          managedACTAccessRequirement,
+          step: 2,
+          researchProjectId: researchProject.id
+        })
+      }).catch(e => {
+        console.log("RequestDataAccessStep1: Error updating research project data: ", e)
```
jay-hodgson, line 67: Ideas to reduce duplicate code here?

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -0,0 +1,575 @@
...
+  const getAccessorsData = (dataAccessRequestData:RequestInterface) => {
+    const accessorChanges = dataAccessRequestData.accessorChanges
+    const ids:string[] = []
+
+    if (!accessorChanges || !accessorChanges.length) {
+      ids.push(user.ownerId)
+      // Add current user to submission request object
+      setFormSubmitRequestObject(prevState => {
+        return Object.assign({}, prevState, {
+          accessorChanges: [{
```
jay-hodgson, line 117: I remember talking about this :)

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -0,0 +1,575 @@
...
+        if (!requestedFileTypes[id]) {
+          requestedFileTypes[id] = []
+        }
+        requestedFileTypes[id].push("attachments")
+      })
+    }
+
+    // Fetch the required doc file names and save them in the state variables
+    if (batchFileRequest.requestedFiles.length) {
+      getFiles(batchFileRequest, token).then(resp => {
```
jay-hodgson, line 199: Thanks for making this a batch call.  It was complicated, but this looks great.

src/lib/containers/DirectDownloadButton.tsx
emmanmills: It will refresh the component. If the token is changed and it also refreshes the page, I guess I can delete this code.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -0,0 +1,575 @@
...
+        })
+      }) // end getFiles
+    }
+  }
+
+  const handleSubmit = async () => {
+    if (formSubmitRequestObject) {
+      try {
+        const resp = await updateDataAccessRequest(formSubmitRequestObject, token)
+        if (resp.etag) { // save success
```
jay-hodgson, line 245: Same question: have you encountered a case when you have a response where etag is undefined?  I'd expect an error from the updateDataAccessRequest call.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
```diff
@@ -0,0 +1,298 @@
...
+          message: (<strong>You have submitted a data access request.</strong>)
+        })
+        break
+      case SUBMISSION_STATE.APPROVED:
+        setAlert({
+          key: 'success',
+          message: (<strong>Your data access request has been approved.</strong>)
+        })
+        break
+      case SUBMISSION_STATE.REJECTED:
```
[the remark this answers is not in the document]
emmanmills: Okay, I updated the code to show the rejected reason instead.

jay-hodgson:
If the previous submission was rejected, does the user have a way to create/update a new submission?

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -0,0 +1,575 @@
...
+    accessorsArr.push(selectedAccessor)
+    setFormSubmitRequestObject(prevState => {
+      return Object.assign({}, prevState, {
+        accessorChanges: accessorsArr
+      })
+    })
+  }
+
+  return (<>
+    <Form className={"access-request-form2"} onSubmit={e => e.preventDefault()}>
```
jay-hodgson, line 395: I wonder if this really needs to be a Form (I assume it's for css only).  If you change this to a div (with the "access-request-form2" css class), does it still just work?

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep1.tsx
emmanmills: I refactored the code out to a new function.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
emmanmills: Got it. I deleted the check for SUPPORTED_ACCESS_REQUIREMENTS.ACTAccessRequirement

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
emmanmills: Done.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccess.tsx
emmanmills: Done. Updated function name.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
emmanmills: I added an else block to show an error notification if resp.etag is not defined.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
emmanmills: I'd like to utilize the bootstrap form elements and the library does provide all the CSS for you so it's convenient. But I also prefer to keep the form tag because there are other form elements like buttons and input fields on the same page so I thought it's semantically meaningful to enclose them all within a form tag.

src/lib/containers/access_requirement_list/managedACTAccess/CancelRequestDataAccess.tsx
jay-hodgson: The http response code should indicate the if the save was successful.  If the RPC was successful (200 level response) then we can assume there were no errors.  SynapseClient will only give you the response if it was successful.

PR #1027, opened 2021-05-11
Fix EntityFinder input customization

nickgros:
#1019 slightly changed how bootstrap inputs look

This broke the visual styling of the customized entity finder input. This is easily fixed by not using bootstrap here


emmanmills (approved):
Looks good! Thanks.

PR #1032, opened 2021-05-11
PORTALS-1916

jay-hodgson:


src/lib/utils/SynapseClient.ts
```diff
@@ -92,12 +92,17 @@ import { VersionInfo } from './synapseTypes/VersionInfo'
 import { SearchQuery, SearchResults } from './synapseTypes/Search'
 import { ResearchProject } from './synapseTypes/ResearchProject'
 import {
-  ManagedACTAccessRequirementStatus,
   RequestInterface,
```
emmanmills, line 95: Was there a merge conflict? This line is required for the ManagedACTAccess component.

src/lib/utils/SynapseClient.ts
```diff
@@ -2229,7 +2234,7 @@ export const getAccessRequirement = (
 export const getAccessRequirementStatus = (
   sessionToken: string | undefined,
   requirementId: string | number,
-): Promise<AccessRequirementStatus | ManagedACTAccessRequirementStatus> => {
+): Promise<AccessRequirementStatus> => {
   const url = `repo/v1/accessRequirement/${requirementId}/status`
```
emmanmills, line 2238: Same as above. This line is required for the ManagedACTAccess component.

src/lib/utils/SynapseClient.ts
jay-hodgson: Nice catch, this was a bad merge from conflict resolution

src/lib/utils/SynapseClient.ts
jay-hodgson: added this back in

emmanmills (approved):
Looks good

PR #1044, opened 2021-05-17
SWC-5589: recommend validating access token on app initialization.  s…

jay-hodgson:
…ervice calls seem to be failing in a different way when passing an invalid access token (resulting in a difficult to debug CORS error)


emmanmills (approved):
Looks good!

PR #1046, opened 2021-05-20
Clean up ErrorBoundary code

nickgros:
Currently, we're using the same fallback component everywhere. For now, let's just package it into a custom error boundary component, and use that.

I also moved some code for managing errors to a new utils file


src/lib/containers/ErrorBanner.tsx
```diff
@@ -1,30 +1,20 @@
...
-
-function isSynapseClientError(
-  error: string | Error | SynapseClientError,
-): error is SynapseClientError {
-  return (error as any).status !== undefined
-}
-
-function isError(error: string | Error | SynapseClientError): error is Error {
-  return (error as any).message !== undefined
-}
```
nickgros, line 22: moved to utils

src/lib/containers/ErrorBanner.tsx
```diff
@@ -85,3 +75,29 @@ export const ErrorBanner = (props: ErrorProps) => {
...
+ *
+ * Use with {@link react-error-boundary#handleError | handleError}
+ * @param props
+ * @returns
+ */
+export const SynapseErrorBoundary: React.FC<
+  Optional<ErrorBoundaryPropsWithComponent, 'FallbackComponent'>
+> = (props: Optional<ErrorBoundaryPropsWithComponent, 'FallbackComponent'>) => (
+  <ErrorBoundary FallbackComponent={ErrorFallbackComponent} {...props} />
+)
```
nickgros, line 103: Now, use this instead of `ErrorBoundary` because our default fallback component is baked-in

src/lib/utils/types/Optional.ts
```diff
@@ -0,0 +1,25 @@
...
+ *   bar: string
+ *   baz?: string // this is now optional
+ * }
+ *
+ * To make multiple properties optional, use a String Literal union:
+ * type OptionalBarBaz = Optional<FooBarBaz, 'bar' | 'baz'>
+ *
+ * @see https://stackoverflow.com/a/54178819
+ */
+export type Optional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
```
nickgros, line 25: Surprisingly, TypeScript doesn't ship a utility type that does this

jay-hodgson (approved):

PR #1047, opened 2021-05-20
SWC-5612 - Store application-wide data (including auth token) in context

nickgros:
## Summary of changes
 
These changes let us pull the authentication token (as well as some other fields) from context, instead of passing them through props. This dramatically cleans up the code by eliminating some excessive prop drilling. This also reduces the risk of error: it was easy to not pass a token via props since it was optional in most contexts. Now the token, is only accessed when it is used.

This will be a breaking change. Any application that uses SRC will need to wrap all SRC components in a  `SynapseContextProvider`.

The application data provided via this context currently includes these values:
* access token (which remains undefined if the user is not logged in)
* whether time should be displayed in UTC
* whether experimental mode is on

The context also provides a `react-query` QueryClient, so our components can now utilize a shared cache when using `react-query`.

This can be extended to provide more data fairly easily, if the need arises.

## A note on testing

I discovered that Enzyme doesn't fully support the React 16+ context API (see [Enzyme React 16 support](https://github.com/enzymejs/enzyme/issues/1553) and https://github.com/enzymejs/enzyme/issues/2189). In particular, `contextType` (which is used exclusively on class components) doesn't work with the shallow renderer. I rewrote tests for a few components to use `mount`, but for other class components, I left the access token as a prop. An access token can still be passed to the component via context using a `SynapseContextConsumer`. Additionally, components accepting access tokens as a prop must still be wrapped in context in live settings, because they may have child components that access context.

Tangentially, [Facebook no longer uses Enzyme in any new tests](https://github.com/enzymejs/enzyme/issues/2429#issuecomment-817750268). I think we should consider at least avoiding the shallow renderer (we can still use Jest mocks to reduce the test surface of complex components).

## Checklist

- [x] Ensure build works
- [x] Ensure tests pass
- [x] Export appropriate context objects in index/rollup
- [x] Test build (authenticated requests are made, responds to login/logout)
  - [x] Works in portals (after releasing, update https://github.com/Sage-Bionetworks/portals/pull/615)
  - [x] Ensure that it works in SWC (after releasing, update https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4595)
  - [x] Ensure that Styleguidist doesn't break
- [x] Remove QueryClient wrappers from homepage components (these are now redundant, since consumers should wrap components in a QueryClientProvider via SynapseWrapper)
- [x] Update README with instructions to use the wrapper.

For follow up:

- [ ] Investigate removing QueryClient wrappers from other components (e.g. EntityFinder, new download list).


README.md
```diff
@@ -128,6 +132,12 @@ Note there are a number of CDNs required to finish this functionality-
 <script src="https://unpkg.com/markdown-it-br@1.0.0/dist/markdown-it-br.min.js"></script>
 ```
 
+## Usage
+
+Virtually all components require a SynapseContext (exported....)
+
+Example: 
+
```
nickgros, line 140: TODO: Write this section

src/__tests__/lib/containers/MarkdownSynapse.test.tsx
```diff
@@ -1,30 +1,79 @@
...
+})
+
+jest.mock('../../../lib/containers/widgets/SynapseImage', () => {
+  return {
+    __esModule: true,
+    default: function MockSynapseImage() {
+      return <div role="img"></div>
+    },
+  }
+})
```
nickgros, line 38: Simple example for how we can use mocks to reduce testing surface while using RTL's `render` or Enzyme's `mount`

src/__tests__/lib/containers/SynapseTable.test.tsx
```diff
@@ -31,11 +31,18 @@ import syn16787123Json from '../../../mocks/syn16787123.json'
...
 const createShallowComponent = (
   props: SynapseTableProps & QueryWrapperChildProps,
 ) => {
-  const wrapper = mount<SynapseTable>(<SynapseTable {...props} />)
+  const wrapper = mount<SynapseTable>(<SynapseTable {...props} />, {
+    wrappingComponent: SynapseContextProvider,
+    wrappingComponentProps: {
+      synapseContext: MOCK_CONTEXT_VALUE,
+    },
+  })
```
nickgros, line 45: TODO: Should use this style for all mounted components. I think this is preferable because `wrapper.instance()` just works. It probably works better with `wrapper.setContext()` as well

src/mocks/MockSynapseContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+ */
+export const SynapseTestContext = jest
+  .fn()
+  .mockImplementation(({ children }) => {
+    return (
+      <SynapseContextProvider synapseContext={MOCK_CONTEXT_VALUE}>
+        {children}
+      </SynapseContextProvider>
+    )
+  })
```
nickgros, line 28: Handy prebuilt pieces to use across all tests.

src/lib/utils/hooks/SynapseAPI/QueryClientProviderWrapper.tsx
```diff
@@ -1,35 +0,0 @@
...
- * To use react-query's useQuery, each component must have access to the QueryClientProvider context.
- * Since we only export individual components, we may not have the required context. The downside of this strategy
- * is that the component MUST use this supplied QueryClient, and cannot configure their own.
- *
- * We should remove this wrapper if we can ensure that these components will always receive a QueryClient via context
- * in every place they are used (including dependent applications, e.g. Portals and SWC)
- *
- * Adapted from https://medium.com/@jrwebdev/react-higher-order-component-patterns-in-typescript-42278f7590fb
- */
-export const withQueryClientProvider = <P extends object>(
```
nickgros, line 26: No longer needed as we now require all components to be wrapped in <SynapseContextProvider>, which itself contains a <QueryClientProvider>

src/lib/utils/SynapseContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+  synapseContext?: SynapseContextType
+  queryClient?: QueryClient
+}
+
+/**
+ * Provides context necessary for most components in SRC
+ * @param param0
+ * @returns
+ */
+export const SynapseContextProvider: React.FunctionComponent<SynapseContextProviderProps> = ({
```
nickgros, line 33: Optional reading: [here's a blog post](https://kentcdodds.com/blog/how-to-use-react-context-effectively) by Kent C. Dodds that has guided many of the design decisions made here.

src/lib/utils/SynapseContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+): React.FC<P> =>
+  function _(props: P) {
+    return (
+      <SynapseContextConsumer>
+        {synapseContext => (
+          <Component {...(props as P)} synapseContext={synapseContext} />
+        )}
+      </SynapseContextConsumer>
+    )
+  }
```
nickgros, line 80: TODO: I think I can delete this

src/lib/utils/SynapseContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export type SynapseContextType = {
+  accessToken?: string
+  isInExperimentalMode: boolean
+  utcTime: boolean
+}
+
+export const SynapseContext = React.createContext<
+  SynapseContextType | undefined
+>(undefined)
```
nickgros, line 21: This needs to be exported to use it in class components.

TODO: Add comment stating the same

src/lib/utils/SynapseConstants.ts
```diff
@@ -88,3 +88,7 @@ export const USER_BUNDLE_MASK_IS_ACT_MEMBER = 0x2
 // SessionStorage keys for info from ids
 export const USER_PROFILE_STORAGE_KEY = 'INFO_FROM_IDS_USER_PROFILE'
 export const ENTITY_HEADER_STORAGE_KEY = 'INFO_FROM_IDS_ENTITY_HEADER'
+
+// TODO: Move to SynapseConstants
```
nickgros, line 92: rm todo

src/lib/rollup.index.ts
```diff
@@ -37,6 +38,7 @@ const SynapseComponents = {
   SynapseHomepage,
   EntityFinder,
   ErrorPage,
+  SynapseContextProvider,
```
nickgros, line 41: TODO: Export all relevant context objects

src/lib/index.ts
```diff
@@ -86,6 +87,7 @@ const SynapseComponents = {
   ProjectViewCarousel,
   EntityFinder,
   ExperimentalMode,
+  SynapseContextProvider,
```
nickgros, line 90: TODO: Export all relevant context objects

src/lib/containers/synapse_form_wrapper/SynapseFormWrapper.md
```diff
@@ -6,7 +6,6 @@ fileNamePath="study.submission_name"
 formUiSchemaEntityId="syn20692911"
 formNavSchemaEntityId="syn20968007"
 isWizardMode={true}
-token={accessToken}
```
nickgros, line 9: todo: still need!

src/lib/containers/synapse_form_wrapper/SynapseFormWrapper.tsx
```diff
@@ -10,6 +10,17 @@ import { FileEntity, FormData } from '../../utils/synapseTypes/'
...
+ * TODO: SWC-5612 - Replace token prop with SynapseContext.accessToken
+ * 
+ * This wasn't done because Enzyme's shallow renderer is not currently
+ * compatible with the `contextType` field in the React 16+ context API.
+ * 
+ * This can be fixed by rewriting tests to not rely on the shallow renderer.
+ * 
+ * See here: https://github.com/enzymejs/enzyme/issues/1553
+ */
+
```
nickgros, line 23: e.g. comment I left on the few classes I didn't want to rewrite tests for

src/lib/utils/hooks/SynapseAPI/useEntityBundle.ts
```diff
@@ -21,14 +22,14 @@ const ALL_FIELDS: EntityBundleRequest = {
...
   version?: number,
   options?: UseQueryOptions<EntityBundle, SynapseClientError, EntityBundle>,
 ) {
+  const { accessToken } = useSynapseContext()
   return useQuery<EntityBundle, SynapseClientError>(
-    ['entitybundle', entityId, version, bundleRequest],
+    [accessToken, 'entitybundle', entityId, version, bundleRequest],
     () =>
       SynapseClient.getEntityBundleV2(
         entityId,
```
nickgros, line 35: I think this is awesome: this query function is a hook itself, so we can call `useContext` here. The caller of `useGetEntityBundle` doesn't even need the access token!

src/lib/utils/SynapseContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { QueryClient, QueryClientProvider } from 'react-query'
+
+const defaultQueryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      staleTime: 30 * 1000, // 30s
+      retry: false, // SynapseClient knows which queries to retry
+    },
+  },
+})
```
nickgros, line 11: Moved this from QueryClientWrapper, we'll provide it here.

src/lib/utils/SynapseContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  )
+}
+
+export const SynapseContextConsumer = SynapseContext.Consumer
+
+export function useSynapseContext(): SynapseContextType {
+  const context = useContext(SynapseContext)
+  if (context === undefined) {
+    throw new Error(
+      'useSynapseContext must be used within a SynapseContextProvider',
```
jay-hodgson, line 56: like it, fail fast

src/lib/containers/QueryWrapper.tsx
```diff
@@ -12,6 +12,18 @@ import {
...
+
+/**
+ * TODO: SWC-5612 - Replace token prop with SynapseContext.accessToken
+ *
+ * This wasn't done because Enzyme's shallow renderer is not currently
+ * compatible with the `contextType` field in the React 16+ context API.
+ *
+ * This can be fixed by rewriting tests to not rely on the shallow renderer.
+ *
+ * See here: https://github.com/enzymejs/enzyme/issues/1553
```
jay-hodgson, line 24: ok

jay-hodgson:
Fully reviewed SynapseContext.tsx, useEntityBundle.ts, HasAccess.tsx, FacetNavPanel.tsx.
As well as README.md, SynapseClient.ts, and MockSynapseContext.tsx.
I also did a cursory review of the rest of the files, and didn't identify any issues.  Nice use of context.

jay-hodgson (approved):

PR #1048, opened 2021-05-24
SWC-5597: note, this uses a service from Synapse stack 358, so the cu…

jay-hodgson:
…rrent prod (357) will return a 404


jay-hodgson:
Screen Shot 2021-05-24 at 2 36 51 PM

emmanmills (approved):
Looks good!

PR #1056, opened 2021-05-27
PORTALS-1934

jay-hodgson:
Use the new HAS_LIKE functiionality

### DIFF
--- a/src/lib/containers/SearchV2.tsx
+++ b/src/lib/containers/SearchV2.tsx
@@ -14,6 +14,8 @@ import { ColumnModel, ColumnType } from '../utils/synapseTypes'
 import {
   ColumnSingleValueQueryFilter,
   ColumnSingleValueFilterOperator,
+  ColumnMultiValueFunction,
+  ColumnMultiValueFunctionQueryFilter,
   QueryFilter,
 } from '../utils/synapseTypes/Table/QueryFilter'
 
@@ -104,19 +106,6 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
     }
   }
 
-  public static addEscapeCharacters = (searchText: string) => {
-    // We have to escape the following characters
-    // ' % \
-    let escapedSearchText = searchText
-    // escape ' by adding additional '
-    escapedSearchText = escapedSearchText.split("'").join("''")
-    // escape % by adding \
-    escapedSearchText = escapedSearchText.split('%').join(`\%`)
-    // escape \ by adding \
-    escapedSearchText = escapedSearchText.split('\\').join('\\\\')
-    return escapedSearchText
-  }
-
   public search = (event: React.SyntheticEvent<HTMLFormElement>) => {
     // form completion by default causes the page to reload, so we prevent that
     event.preventDefault()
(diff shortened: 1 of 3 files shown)


jay-hodgson:
Here's what the request looks like using the new (multi-value) and old (single value) types of additionalFilter.  The 'study' column is a STRING_LIST type column, and 'name' is a STRING column.
Screen Shot 2021-05-27 at 1 26 48 PM

src/lib/containers/SearchV2.tsx
```diff
@@ -104,19 +106,6 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
     }
   }
 
-  public static addEscapeCharacters = (searchText: string) => {
```
jay-hodgson, line 107: not used, so deleted.

src/lib/containers/SearchV2.tsx
```diff
@@ -151,14 +140,27 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
       return false
     })
     if (indexOfColumn === -1) {
-      const columnSingleValueQueryFilter: ColumnSingleValueQueryFilter = {
```
jay-hodgson, line 154: Used to always add a new ColumnSingleValueQueryFilter to the additionalFilters array.  Now, look at the columnModel that is associated to the selected columnName.  If the column type ends with '_LIST', then it's a multi-value column, so we should add a ColumnMultiValueFunctionQueryFilter instead.  Otherwise, do what it used to do (add a ColumnSingleValueQueryFilter)

src/lib/containers/SearchV2.tsx
```diff
@@ -287,7 +289,7 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
                 const isSelected =
                   (columnName === '' && index === 0) || columnName === name
                 return (
-                  <div className="radio">
+                  <div className="radio" key={`search-field-${name}-${index}`}>
```
jay-hodgson, line 292: compiler was complaining that a key was not set.

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -224,9 +224,9 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
             el.columnName === columnName
               ? el.values.filter(el => el !== value)
               : el.values,
```
jay-hodgson, line 226: We can't assume this is a ColumnSingleValueQueryFilter anymore!  Now, when a value is removed from an existing filter, use the original filter values (either going to have an operator if it's a single value filter, or function if it's a multi-value filter).

src/lib/utils/synapseTypes/Table/QueryFilter.ts
```diff
@@ -1,17 +1,29 @@
...
+// A filter using a single value against a column. Values within the values list have an OR relationship
 export interface ColumnSingleValueQueryFilter extends QueryFilter {
   concreteType: 'org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter'
   operator: ColumnSingleValueFilterOperator // Determines the filter operation to perform
   values: string[] // Values to used with the filter.
-  // Multiple values will be joined in an OR relationship.
+}
+
+// A filter that applies multi-value columns functions.
+export interface ColumnMultiValueFunctionQueryFilter extends QueryFilter {
```
jay-hodgson, line 17: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/ColumnMultiValueFunctionQueryFilter.html

emmanmills (approved):
Looks good!

PR #1057, opened 2021-05-28
SWC-5625

nickgros:
Error boundary in table
image

Not in table, triggered by boundary:
image

Not in table, not triggered by boundary
image

### DIFF
--- a/src/lib/style/components/_error.scss
+++ b/src/lib/style/components/_error.scss
@@ -1,6 +1,9 @@
 .Error {
-  div {
+  div.alert {
     text-align: center;
+    display: flex;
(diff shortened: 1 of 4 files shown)


src/lib/style/components/_error.scss
```diff
@@ -1,6 +1,9 @@
 .Error {
-  div {
+  div.alert {
     text-align: center;
+    display: flex;
+    justify-content: space-around;
```
nickgros, line 5: Might look better with `space-between` but I don't know if I want the Reload button hidden so far off the screen in the table component.

jay-hodgson (approved):

PR #1058, opened 2021-06-01
SWC-5629

nickgros:
image

### DIFF
--- a/src/lib/style/bootstrap4_backports/_overrides.scss
+++ b/src/lib/style/bootstrap4_backports/_overrides.scss
@@ -76,19 +76,16 @@
 }
 
 // Modals
-.modal-content {
-  padding: 5px 30px;
-  .modal-header {
-    display: flex;
-    justify-content: space-between;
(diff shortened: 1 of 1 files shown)


src/lib/style/bootstrap4_backports/_overrides.scss
```diff
@@ -76,19 +76,16 @@
...
-    align-items: center;
-    padding: 10px 0px;
-
-    .modal-title {
-      order: -99;
-    }
-    button.close {
-      order: 99;
+body:not(#body) {
+  // We don't want these styles in Synapse.org, it has its own modal styling
```
nickgros, line 80: Should we continue using this pattern going forward?

jay-hodgson: Maybe we should be more explicit.  Perhaps we set the html element in SWC to id="synapse.org"?  Then this would say body:not(#synapse.org)...

PR #1063, opened 2021-06-03
Add/improve type definitions

nickgros:
General improvements and work supporting SWC-5580

### DIFF
--- a/src/lib/utils/synapseTypes/Entity.ts
+++ b/src/lib/utils/synapseTypes/Entity.ts
@@ -7,16 +7,16 @@ export interface Entity {
   attachments?: AttachmentData[] // 	This field is deprecated and will be removed in future versions of Synapse
   name: string // The name of this entity.
   description?: string // The description of this entity.
-  id?: string // The unique immutable ID for this entity. A new ID will be generated for new Entities. Once issued, this ID is guaranteed to never change or be re-issued
+  readonly id?: string // The unique immutable ID for this entity. A new ID will be generated for new Entities. Once issued, this ID is guaranteed to never change or be re-issued
   uri?: string // This field is deprecated and will be removed in future versions of Synapse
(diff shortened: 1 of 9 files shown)


src/lib/utils/synapseTypes/Entity.ts
```diff
@@ -4,19 +4,34 @@ import { ConcreteEntityType } from './ConcreteEntityType'
...
+  readonly createdBy?: string
+  /** The ID of the user that last modified this entity. */
+  readonly modifiedBy?: string
+  /** The ID of the Entity that is the parent of this Entity. */
+  parentId?: string
+  /** Indicates which implementation of Entity this object represents. */
+  readonly concreteType: ConcreteEntityType
+  /** @deprecated This field is deprecated and will be removed in future versions of Synapse */
+  attachments?: AttachmentData[]
+  /** @deprecated This field is deprecated and will be removed in future versions of Synapse */
```
nickgros, line 29: We should use this comment style when possible because VSCode (hopefully other IDEs too?) displays/hints this info inline:

image

jay-hodgson: Like this idea.  Any way to help devs work towards this style (other than this comment)?  Something in VSCode configuration?  Generating ts definitions from the schema would certainly do the trick :)

src/lib/utils/synapseTypes/Schema/JsonSchemaObjectBinding.ts
```diff
@@ -0,0 +1,25 @@
+export type JsonSchemaVersionInfo = {
+  organizationId: string
+  organizationName: string
+  schemaId: string
+  schemaName: string
+  versionId: string
+  $id: string
```
jay-hodgson, line 7: This notation takes me back to the days of jQuery

PR #1065, opened 2021-06-04
SWC-5620

nickgros:
First see new type definitions #1063 (which is rolled into this branch)

Adds modal for previewing an entity.

Still need to add tests.


src/mocks/msw/handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+  mockProjectEntityBundle,
+  MOCK_FILE_ENTITY_ID,
+  MOCK_PROJECT_ID,
+} from '../entity/mockEntity'
+import { mockUserBundle, mockUserProfileData, MOCK_USER_ID } from '../user/mock_user_profile'
+
+const handlers = [
+  rest.options('*', async (req, res, ctx) => {
+    return res(ctx.status(200))
+  }),
```
nickgros, line 18: Is there a more sensible way to handle the pre-flight OPTIONS requests?

src/mocks/msw/handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  rest.get(
+    `${getEndpoint(
+      BackendDestinationEnum.REPO_ENDPOINT,
+    )}repo/v1/userProfile/:id`,
```
nickgros, line 23: In the near future, I'd like to 
* Reference API paths using constants, so these handlers reference the same values as SynapseClient
* Split these handlers up and group them by related services, just to keep things tidy

src/__tests__/lib/containers/entity/metadata/AnnotationsTable.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+        )}repo/v1/entity/:entityId/bundle2`,
+
+        async (req, res, ctx) => {
+          return res(
+            ctx.status(200),
+            ctx.json({ ...mockFileEntityBundle, annotations: undefined }),
+          )
+        },
+      ),
+    )
```
nickgros, line 74: This is how you change `msw` behavior for the duration of an individual test (as long as `server.restoreHandlers()` is called in `afterEach`).

src/lib/containers/UserCard.tsx
```diff
@@ -66,74 +67,37 @@ export const UserCard: React.FunctionComponent<UserCardProps> = (
   // We fetch the image right away in case it is an expiring presigned URL
   const imageURL = usePreFetchResource(preSignedURL)
 
+  const { data: profileAndImage } = useGetUserProfileWithProfilePic(
+    principalId!,
+    {
+      enabled: !!principalId,
+    },
+  )
```
nickgros, line 75: We already had this hook, which does everything the deleted effect (below) was doing, minus some sessionStorage manipulation, so let's just use it here.

src/lib/containers/UserCardMedium.tsx
```diff
@@ -39,306 +33,306 @@ export type UserCardMediumProps = {
...
 
-  constructor(props: UserCardMediumProps) {
-    super(props)
-    this.state = {
-      showModal: false,
-      isContextMenuOpen: false,
-      ORCIDHref: undefined
-    }
-  }
+export const UserCardMedium: React.FC<UserCardMediumProps> = ({
```
nickgros, line 67: Refactored to a function component. I was having trouble reasoning with the class implementation. I ultimately didn't need to do this, but I didn't realize that until I had already done it 😬

Can revert if desired

src/lib/containers/entity/metadata/EntityModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export type EntityModalProps = {
+  readonly show: boolean
+  readonly entityId: string
+  readonly onClose: () => void
+  readonly initialTab?: EntityModalTabs
+}
+
+export const EntityModal: React.FC<EntityModalProps> = ({
```
nickgros, line 29: TODO: Add tests

src/lib/testutils/TestingLibraryUtils.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+      <SynapseContextProvider
+        synapseContext={wrapperProps}
+        queryClient={queryClient}
+      >
+        {children}
+      </SynapseContextProvider>
+    )
+  }
+}
```
nickgros, line 40: We should use this going forward for all RTL tests that need the Context wrapper

src/lib/utils/functions/FileHandleUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+    case CloudProviderFileHandleConcreteTypeEnum.GoogleCloudFileHandle:
+      return `gs://${(fileHandle as GoogleCloudFileHandle).bucketName}`
+    default:
+      throw new Error(
+        `Couldn't determine location name for file handle: ${JSON.stringify(
+          fileHandle,
+        )}`,
+      )
+  }
+}
```
nickgros, line 43: I need to test this and validate that this is the correct behavior

src/lib/utils/hooks/SynapseAPI/useEntityBundle.ts
```diff
@@ -29,7 +29,7 @@ export default function useGetEntityBundle(
 ) {
   const { accessToken } = useSynapseContext()
   return useQuery<EntityBundle, SynapseClientError>(
-    [accessToken, 'entitybundle', entityId, version, bundleRequest],
+    [accessToken, 'entity', entityId, 'bundle', version, bundleRequest],
```
nickgros, line 32: Thinking forward, our query keys should make sense when invalidating the cache in response to a mutation.

For example, when an entity is mutated, we would invalidate the query key `[accessToken, 'entity', entityId], which (I think) would blow away the bundle, plus whatever other future queries are nested in this way

src/lib/utils/synapseTypes/EntityHeader.ts
```diff
@@ -1,20 +1,38 @@
...
+ * The full path for an entity. Note that the root entity, syn4489, may be included.
+ * The docs do not indicate that most fields are omitted from the EntityHeader objects, but our custom type
+ * signature captures this.
+ *
+ * [EntityPath | Synapse REST Docs](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityPath.html)
+ */
 export type EntityPath = {
-  path: EntityHeader[] // The list of all entities in this entity's path. The first element is the root parent and the last element (n) is the entity.
+  path: Pick<EntityHeader, 'name' | 'id' | 'type'>[] // The list of all entities in this entity's path. The first element is the root parent and the last element (n) is the entity.
 }
```
nickgros, line 38: I caught this nuance in the entity path service while creating some of the mocks, so any `EntityHeader`-related changes in this PR were caused by this.

src/mocks/msw/handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+    `${getEndpoint(
+      BackendDestinationEnum.REPO_ENDPOINT,
+    )}repo/v1/entity/:entityId/bundle2`,
+    async (req, res, ctx) => {
+      let response
+      if (req.params.entityId === MOCK_FILE_ENTITY_ID) {
+        response = mockFileEntityBundle
+      } else if (req.params.entityId === MOCK_PROJECT_ID) {
+        response = mockProjectEntityBundle
+      }
```
nickgros, line 56: Here we take the request and use the ID to map it to a specific mock object. The IDs are exposed, so a test suite can reference the `MOCK_FILE_ENTITY_ID` constant. In my mind, this makes it easier to configure multiple mocks returned from the same service, with a cost of increased configuration/mock maintenance.

Is this sensible?

nickgros:
Splitting into two PRs

PR #1067, opened 2021-06-04
Add typescript-eslint

nickgros:


jay-hodgson:
Agree with this direction.  https://github.com/typescript-eslint/typescript-eslint

jay-hodgson (approved):

PR #1068, opened 2021-06-04
PORTALS-1925

jay-hodgson:


jay-hodgson:
Screen Shot 2021-06-03 at 3 19 05 PM

src/lib/utils/synapseTypes/DownloadListV2/ActionRequiredCount.ts
```diff
@@ -0,0 +1,22 @@
+// http://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/download/ActionRequiredCount.html
```
jay-hodgson, line 1: Not yet used.  Introduced new types for future work (John is still working on the Actions Required services)

PR #1069, opened 2021-06-07
References to Synapse.org should conditionally point to staging

nickgros:


jay-hodgson:
Today the portal endpoint is relative if this code is being run on a known Synapse.org portal domain ('www.synapse.org', 'staging.synapse.org', or 'tst.synapse.org')

PR #1073, opened 2021-06-07
PORTALS-1942: Fix unstable tests

nickgros:
The following test files occasionally fail sporadically in CI (but are very hard to reproduce locally):
```
(Not comprehensive, many missing) 
FacetNav.test.tsx
useGetEntityChildren.test.tsx
TotalQueryResults.test.tsx
```

I originally tried to introduce `msw` and started by reorganizing our mocks. The mock reorganization results in the bulk of the changed-line count.

Using `msw` across all tests is going to be a big project, because a lot of tests break. I also found it very difficult to run `msw` in individual test suites, leaving others alone.

Out of desperation, I decided to upgrade react-scripts to v4 and that seems to have fixed it for reasons unknown. I suspect it has something to do with Jest or JSDOM, which were both upgraded.

I'd like to run the Travis CI build 5 times and get 5 successes to consider this fixed:
CI results:
✅✅✅✅✅


package.json
```diff
@@ -136,14 +136,15 @@
     "core-js": "^3.1.4",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
-    "eslint": "6.6.0",
```
nickgros, line 139: eslint is included n react-scripts

src/__tests__/lib/containers/EntityIdList.test.tsx
```diff
@@ -3,36 +3,39 @@ import { mockAllIsIntersecting } from 'react-intersection-observer/test-utils'
...
+import { mockFileEntityHeader } from '../../../mocks/entity/mockEntity'
+import { getEntityHeadersByIds } from '../../../lib/utils/SynapseClient'
+
+const SynapseClient = require('../../../lib/utils/SynapseClient')
+
+SynapseClient.getEntityHeadersByIds = jest
+  .fn()
+  .mockResolvedValue({ results: [mockFileEntityHeader] })
 
 describe('EntityIdList: basic functionality', () => {
```
nickgros, line 21: This test suite was giving me issues so I altered it a bit

Instead of two enzyme tests that 
* verify rendering
* mock useState

I replaced this with one RTL test that mocks SynapseClient and verifies that the state string is visible in the rendered component

src/__tests__/lib/containers/download_list/DownloadListTable.test.tsx
```diff
@@ -10,28 +10,19 @@ import {
...
-import { act } from 'react-dom/test-utils'
+import React from 'react'
+import { mount } from 'enzyme'
 import {
   MOCK_CONTEXT_VALUE,
   SynapseTestContext,
 } from '../../../../mocks/MockSynapseContext'
+import { resolveAllPending } from '../../../../lib/testutils/EnzymeHelpers'
 
 describe('it performs all functionality ', () => {
```
nickgros, line 21: This was another test giving me trouble. I decided to replace ReactDOM test utils with enzyme's mount, which was a simple change and seemed to resolve the issues I was encountering

src/__tests__/lib/utils/SynapseClient.test.tsx
```diff
@@ -106,30 +86,10 @@ describe('it works at integration level testing', () => {
...
-      .then((data) => {
-        expect(data.entity).toBeDefined()
-        expect(data.restrictionInformation).toBeDefined()
-        expect(data.fileHandles).toBeDefined()
-      })
-      .catch((err) => {
+      .catch(err => {
         fail(err.reason)
       })
   })
```
nickgros, line 95: This API is deprecated, so why test it

src/__tests__/lib/utils/SynapseClient.test.tsx
```diff
@@ -5,68 +5,48 @@ import {
   FileHandleAssociateType,
   PaginatedResults,
 } from '../../../lib/utils/synapseTypes/'
-import { SynapseClient, SynapseConstants } from '../../../lib/utils/'
+import { SynapseClient } from '../../../lib/utils/'
 import { FunctionReturningPaginatedResults } from '../../../lib/utils/SynapseClient'
 
 describe('it works at integration level testing', () => {
```
nickgros, line 11: I don't really like these tests. In general, they seem to be testing the Synapse backend rather than the behavior of the client itself. Would like to eventually replace with MSW, but not right now.

src/__tests__/lib/utils/SynapseClient.test.tsx
```diff
@@ -5,68 +5,48 @@ import {
...
-        )
-      })
-      .catch((resp) => {
-        // invalid access token
-        expect(resp['status']).toBe(401)
-      })
+    return SynapseClient.getVersion().then(data => {
+      expect(data.version).toBeDefined()
+    })
   })
```
nickgros, line 30: e.g. this is really just testing that we get a 401 when Synapse could easily change to return a 404 and still be valid behavior.

package.json
```diff
@@ -136,14 +136,15 @@
...
+    "eslint-plugin-jest": "^24.3.6",
```
nickgros, line 139: Some additional lint rules for Jest

package.json
```diff
@@ -136,14 +136,15 @@
...
     "eslint-plugin-jest-dom": "^3.6.5",
     "eslint-plugin-react": "latest",
     "eslint-plugin-react-hooks": "^2.5.0",
     "eslint-plugin-testing-library": "^3.10.2",
     "identity-obj-proxy": "^3.0.0",
     "jest-mock-promise": "^1.1.10",
     "jest-when": "^3.2.1",
+    "msw": "^0.29.0",
```
nickgros, line 147: Not used yet, but I'd like to start using it soon so I didn't remove it from package.json

package.json
```diff
@@ -179,7 +180,7 @@
     "start-js": "react-scripts start",
     "start": "npm-run-all -p watch-css serve:docs",
     "start-deprecated-demo": "npm-run-all -p watch-css start-js",
-    "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!(lodash-es|jest*)/)\" , --env=jsdom-fourteen --setupFiles ./src/setupTests.js",
+    "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!(lodash-es|jest*)/)\" , --env=jsdom --setupFiles ./src/setupTests.js --resetMocks false",
```
nickgros, line 183: Breaking change in upgrading Jest from v24 to v26 requires setting `--resetMocks=false`. Additionally, the default version of JSDOM is now v16, so we don't need to specify jsdom-fourteen anymore

package.json
```diff
@@ -189,7 +190,7 @@
     "build:js": "tsc --p tsconfig.build.json",
     "build:copy-assets": "cd src/lib && find . -name '*.scss' -o -name '*.css' -o -name '*.svg' | cpio -pdm ../../dist/",
     "build:rollup": "rollup -c && cp -r src/umd dist/",
-    "serve:docs": "npx styleguidist server --open",
+    "serve:docs": "DISABLE_ESLINT_PLUGIN=true npx styleguidist server --open",
```
nickgros, line 193: By default, ESLint runs before the styleguidist server launches. This can take a long time (seemingly longer with CRA v4/ESLint v7), and it's not very useful since we aren't actively monitoring/using our lint rules at this time.

So this effectively speeds up `yarn start`. We can change this if we decide to spend time configuring our linter and cleaning up existing lint issues.

README.md
```diff
@@ -358,12 +358,13 @@ There are open issues in the microsoft vscode repository that block the ability
     "test",
     "--runInBand",
     "--no-cache",
-    "--env=jsdom-fourteen",
+    "--env=jsdom",
     "--transformIgnorePatterns",
     // https://github.com/microsoft/vscode/issues/81944
```
jay-hodgson, line 363: Not sure why this is highlighted as red.  That makes me think this part of the string was deleted, which it wasn't.

nickgros: The red highlight is because it's invalid JSON syntax (as far as GitHub's markdown parser can tell). `launch.json` allows JSON-with-comments so it's not a real issue.

src/__tests__/lib/utils/SynapseClient.test.tsx
jay-hodgson: That's reasonable.  Are we using the updated API (entity bundle2 services)?  Is there a test?

nickgros: We have a call to SynapseClient that uses the v2 API. Like most of the SynapseClient functions, we don't have an explicit test for it. The Entity Finder is one example of a component that is using it.

tsconfig.json
```diff
@@ -6,9 +6,12 @@
...
     "module": "esnext",
     "target": "es5",
-    "lib": ["es6", "dom"],
+    "lib": [
+      "es6",
+      "dom"
+    ],
     "sourceMap": true,
-    "jsx": "preserve",
+    "jsx": "react",
```
jay-hodgson, line 14: Trying to get rid of Babel? :)

jay-hodgson (approved):

PR #1074, opened 2021-06-07
PORTALS-1943: was unable to get the entity when the dot version notation was used

jay-hodgson:


emmanmills (approved):
Looks good! Thanks.

PR #1075, opened 2021-06-08
PORTALS-1944: get entity headers on component update

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -178,6 +178,7 @@ export default class SynapseTable extends React.Component<
       : true
   }
   componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps) {
+    this.getEntityHeadersInData(false)
```
jay-hodgson, line 181: This fixes PORTALS-1944, adding back a line to componentDidUpdate() deleted here:
https://github.com/Sage-Bionetworks/Synapse-React-Client/commit/4d259565dc27075e85fa85f4c2631ff12791a85f#diff-b5ba738f96377d79f0c1384e300dcfef65e8fe1b917d084f6086c15bf0caf53fL178

src/lib/utils/SynapseClient.ts
```diff
@@ -887,9 +887,21 @@ export const getEntityHeaders = (
   references: ReferenceList,
   accessToken?: string,
 ) => {
+  // if references contains entity IDs with dot notation, fix the reference object
+  const fixedReferences = references.map(reference => {
```
jay-hodgson, line 891: I saw another component getEntityHeaders failing because it was giving a list of entity IDs with dot notation, but didn't track down the component that was calling it.  I thought I'd just fix it here for good measure.

src/lib/containers/table/SynapseTable.tsx
nickgros: Is this ever called with `true`? We may be able to remove the parameter (but this looks good for a hotfix)

PR #1079, opened 2021-06-14
SWC-5620 pre-work

nickgros:
None of this is directly related to SWC-5620 but all of these changes made developing/testing SWC-5620 easier. Most of the changes are related to mocks and running msw.

Summary:
* Refactored UserCard (it was hard to test UserCard with msw in new components for SWC-5620)
	* Now uses existing hook to fetch user data and profile picture. Previously, UserCard was using a sessionStorage cache, but now it uses react-query and duplicated logic to fetch a profile picture was removed.
* Refactored UserCardMedium
	* Converted from class component to function component
	* This was unnecessary, but it made it easier for me to reason with the component's lifecycle. I can revert if desired.
* UserCardTests.test.tsx changes
	* Now uses Mock Service worker. The worker only gets invoked for one or two tests, but it's a start.
	* Use a more reliable, reusable wrapper for React Testing Library tests
* Started converting SynapseClient API paths to constants
	* We need the API paths for Mock Service Worker, so these should be reusable constants.
* Reorganized mocks and added Mock Service Worker handlers
	* Some of the handlers are unused here but will be used in SWC-5620 tests.
	* Mocks have been fleshed out a bit
* Various type improvements to better reflect actual server responses

### DIFF
--- a/src/__tests__/lib/containers/UserCardTests.test.tsx
+++ b/src/__tests__/lib/containers/UserCardTests.test.tsx
@@ -16,12 +16,12 @@ import UserCardContextMenu, {
   MenuAction,
 } from '../../../lib/containers/UserCardContextMenu'
 import { SEPERATOR } from '../../../lib/utils/SynapseConstants'
-import { resolveAllPending } from '../../../lib/testutils/EnzymeHelpers'
-import { act } from 'react-dom/test-utils'
 import { Avatar, AvatarProps } from '../../../lib/containers/Avatar'
 import { render, waitFor, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { SynapseTestContext } from '../../../mocks/MockSynapseContext'
+import { server } from '../../../mocks/msw/server'
+import { createWrapper } from '../../../lib/testutils/TestingLibraryUtils'
 
 jest.mock('../../../lib/utils/hooks/usePreFetchImage', () => {
   return {
@@ -74,6 +74,11 @@ const createMountedComponent = (props: UserCardProps) => {
   return { wrapper, instance }
 }
 
+// Handle the msw lifecycle:
+beforeAll(() => server.listen())
+afterEach(() => server.restoreHandlers())
+afterAll(() => server.close())
+
 describe('it renders the different sized cards without failing', () => {
   const props = {
     userProfile: mockUserProfileData,
@@ -93,18 +98,20 @@ describe('it renders the different sized cards without failing', () => {
     expect(wrapper.find(UserCardSmall)).toHaveLength(1)
   })
 
-  it('renders a medium card', () => {
+  it('renders a medium card', async () => {
     const size = SynapseConstants.MEDIUM_USER_CARD
-    const { wrapper } = createMountedComponent({ ...props, size })
-    expect(wrapper).toBeDefined()
-    expect(wrapper.find(UserCardMedium)).toHaveLength(1)
+    render(<UserCard {...props} size={size} />, { wrapper: createWrapper() })
+    // This is the final uncancellable async request that gets sent in the medium and large cards
+    // We wait for this explicitly to prevent the "not wrapped in act(...)" warning
+    await waitFor(() => screen.getByText('ORCID', { exact: false }))
   })
 
-  it('renders a large card', () => {
+  it('renders a large card', async () => {
     const size = SynapseConstants.LARGE_USER_CARD
-    const { wrapper } = createMountedComponent({ ...props, size })
-    expect(wrapper).toBeDefined()
-    expect(wrapper.find(UserCardLarge)).toHaveLength(1)
+    render(<UserCard {...props} size={size} />, { wrapper: createWrapper() })
+    // This is the final uncancellable async request that gets sent in the medium and large cards
+    // We wait for this explicitly to prevent the "not wrapped in act(...)" warning
+    await waitFor(() => screen.getByText('ORCID', { exact: false }))
   })
 })
 describe('it creates the correct UI for the avatar', () => {
@@ -157,12 +164,8 @@ describe('it creates the correct UI for the small card', () => {
   })
 
   it('shows a medium user card when mouse enters', async () => {
-    render(
-      <SynapseTestContext>
-        <UserCard {...props} />
-      </SynapseTestContext>,
-    )
-    expect(screen.getByText(`@${props.userProfile.userName}`)).not.toBeNull()
+    render(<UserCard {...props} />, { wrapper: createWrapper() })
+    await waitFor(() => screen.getByText(`@${props.userProfile.userName}`))
 
     // There is no medium user card, so we shouldn't be able to find the full name anywhere
     expect(() =>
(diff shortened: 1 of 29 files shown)


src/mocks/msw/handlers.ts
```diff
@@ -0,0 +1,81 @@
...
+    async (req, res, ctx) => {
+      let response
+      if (req.params.entityId === MOCK_FILE_ENTITY_ID) {
+        response = mockFileEntityBundle
+      } else if (req.params.entityId === MOCK_PROJECT_ID) {
+        response = mockProjectEntityBundle
+      }
+      return res(ctx.status(200), ctx.json(response))
+    },
+  ),
```
nickgros, line 69: Here, we map specific IDs to specific mocks. The ID constants (`MOCK_FILE_ENTITY_ID`) are exposed so tests can use them.

This isn't used in tests in this PR, but I'm wondering if this seems like a good pattern to use, or if it will become a headache to maintain our mocks.

TODO: Return 404 if not found

src/mocks/msw/handlers.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (req.params.entityId === MOCK_FILE_ENTITY_ID) {
+        response = mockFileEntityBundle
+        status = 200
+      } else if (req.params.entityId === MOCK_PROJECT_ID) {
+        response = mockProjectEntityBundle
+        status = 200
+      }
+      return res(ctx.status(status), ctx.json(response))
+    },
+  ),
```
nickgros, line 76: Here, we map specific IDs to specific mocks. The ID constants (`MOCK_FILE_ENTITY_ID`) are exposed so tests can use them.

This isn't used in tests in this PR, but I'm wondering if this seems like a good pattern to use, or if it will become a headache to maintain our mocks.

src/mocks/entity/mockEntity.ts
```diff
@@ -44,3 +83,144 @@ export const mockFolderEntityHeader: EntityHeader = {
...
 }
+
+export const mockProjectEntityBundle: EntityBundle = {
+  entity: {
+    name: 'A Mock Project',
+    id: MOCK_PROJECT_ID,
+    etag: '7849ff2c-1c93-4104-adcf-9e6d6b0c50b5',
+    createdOn: '2020-11-18T20:05:06.519Z',
+    modifiedOn: '2021-05-19T12:44:35.457Z',
+    createdBy: `${MOCK_USER_ID}`,
```
nickgros, line 94: In general, I'm trying to associate all of our mocks so when a component loads multiple related objects, all data ends up being valid.

src/lib/utils/synapseTypes/EntityHeader.ts
```diff
@@ -1,20 +1,38 @@
...
+ * The full path for an entity. Note that the root entity, syn4489, may be included.
+ * The docs do not indicate that most fields are omitted from the EntityHeader objects, but our custom type
+ * signature captures this.
+ *
+ * [EntityPath | Synapse REST Docs](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityPath.html)
+ */
 export type EntityPath = {
-  path: EntityHeader[] // The list of all entities in this entity's path. The first element is the root parent and the last element (n) is the entity.
+  path: Pick<EntityHeader, 'name' | 'id' | 'type'>[] // The list of all entities in this entity's path. The first element is the root parent and the last element (n) is the entity.
 }
```
nickgros, line 38: Realized that the API only returns an abbreviated entity header so this is now reflected in the type

src/lib/utils/functions/getEndpoint.ts
```diff
@@ -10,7 +10,7 @@ type EndpointObject = {
 }
 
 export const PRODUCTION_ENDPOINT_CONFIG: EndpointObject = {
-  REPO: 'https://repo-prod.prod.sagebase.org/',
+  REPO: 'https://repo-prod.prod.sagebase.org',
```
nickgros, line 13: Mock Service Worker is picky about the difference between `https://domain.com/resource` and `https://domain.com//resource`, and we had both the single and double slash styles in our client code. Synapse backend doesn't care about it, so it was never a problem before now.

Getting rid of the slash here makes it harder to accidentally double-slash.

I double-checked that all `'repo'`, `'file'`, and `'auth'` strings in `SynapseClient` are now preceded by a slash, so I am hopeful that this doesn't break anything.

src/lib/utils/SynapseClient.ts
```diff
@@ -332,14 +352,14 @@ export const addFilesToDownloadList = (
   updateParentState?: any,
 ) => {
   return doPost<AsyncJobId>(
-    `file/v1/download/list/add/async/start`,
+    `/file/v1/download/list/add/async/start`,
```
nickgros, line 355: For all of these added slashes, see https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1079/files#r651002737

src/lib/utils/APIConstants.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Separating API endpoints into their own constants file
+ */
+
+import { getEndpoint, BackendDestinationEnum } from './functions/getEndpoint'
+
+export const BACKEND_ENDPOINT = `${getEndpoint(
+  BackendDestinationEnum.REPO_ENDPOINT,
+)}`
```
nickgros, line 9: I didn't incorporate this into the strings for now, because the methods in `SynapseClient` accept the endpoint as an argument. I think it would make sense to do so eventually.

src/lib/utils/APIConstants.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const REPO = `/repo/v1`
+export const AUTH = `/auth/v1`
+export const FILE = `/file/v1`
+
+export const ENTITY = `${REPO}/entity`
+export const ENTITY_ID = (id: string | number) => `${REPO}/entity/${id}`
```
nickgros, line 16: Thoughts on using functions to reuse template literals?

src/lib/testutils/TestingLibraryUtils.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+      <SynapseContextProvider
+        synapseContext={wrapperProps}
+        queryClient={queryClient}
+      >
+        {children}
+      </SynapseContextProvider>
+    )
+  }
+}
```
nickgros, line 40: We should use this wrapper for all React Testing Library tests that require context.

src/__tests__/lib/containers/UserCardTests.test.tsx
```diff
@@ -172,29 +175,26 @@ describe('it creates the correct UI for the small card', () => {
...
+    userEvent.hover(screen.getByText(`@${props.userProfile.userName}`))
+
+    // The card should appear, which would let us see first/last name
+    // we have to wrap in a waitFor because of the delay
+    await waitFor(() =>
+      screen.getAllByText(
+        `${props.userProfile.firstName} ${props.userProfile.lastName}`,
+      ),
+    )
+    await waitFor(() => screen.getByText('ORCID', { exact: false }))
```
nickgros, line 187: This fixes the `not wrapped in act(...)` warning, but I realized we see this warning so much because we don't have a mechanism for cancelling our web requests in `useEffect` when we tear down a component. Should we look into doing this with [fetch](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) or [axios](https://axios-http.com/docs/cancellation)?

src/lib/containers/UserCard.tsx
```diff
@@ -66,74 +67,37 @@ export const UserCard: React.FunctionComponent<UserCardProps> = (
...
-          .catch(err => {
-            console.warn('failed to get user profile picture ', err)
-          })
-      } else {
-        setIsLoading(false)
-      }
-    }
-
-    if (!userProfile && principalId) {
-      const cachedProfileString = sessionStorage.getItem(
```
jay-hodgson, line 113: In synapse.org, will any cache be referenced for the UserCard data?  
I think this was using a session storage cache established on synapse.org.

nickgros: Aha: https://github.com/Sage-Bionetworks/SynapseWebClient/blob/f70b3a781694c2bd5002f0722c66bf07b3029e6b/src/main/java/org/sagebionetworks/web/shared/WebConstants.java#L244

src/__tests__/lib/containers/UserCardTests.test.tsx
jay-hodgson: Do you mean cancelling the calls done by SynapseClient (which use fetch)?

src/lib/utils/APIConstants.ts
jay-hodgson: Ok with me (although I'm not sure about the style to have all letters capitalized for this function)

jay-hodgson: What do others on the web team think?

jay-hodgson (approved):

src/lib/utils/hooks/SynapseAPI/useUserBundle.ts
```diff
@@ -1,22 +1,48 @@
...
+  const cachedValue = sessionStorage.getItem(sessionStorageCacheKey)
+
+  const defaultOptions: QueryObserverOptions<
+    UserProfileAndImg,
+    SynapseClientError
+  > = {
+    // Use the sessionStorage cache to pre-populate profile data before the request succeeds.
+    placeholderData: cachedValue
+      ? { userProfile: JSON.parse(cachedValue) as UserProfile }
+      : undefined,
```
nickgros, line 34: just a note, react-query also has `initialData`, but that gets cached, while `placeholderData` does not

src/__tests__/lib/containers/UserCardTests.test.tsx
nickgros: Yes

jay-hodgson (approved):

PR #1080, opened 2021-06-14
SWC-5620 - Entity Metadata modal

nickgros:
Depends on #1079 

Metadata tab
image

Annotations tab, experimental mode off
image

Annotations tab, experimental mode on. The schema shows up.
image

The link just points to the raw json file

### DIFF
--- a/src/lib/style/components/_entity-metadata.scss
+++ b/src/lib/style/components/_entity-metadata.scss
@@ -1,24 +1,9 @@
 @use '../abstracts/variables' as SRC;
+@use '../abstracts/mixins' as SRCMixins;
 
 .EntityMetadata {
   .Tabs {
-    display: flex;
-    justify-content: space-between;
-
-    .Tab {
-      cursor: pointer;
-      text-align: center;
-      flex-grow: 1;
-      padding: 5px 0px;
-      transition: border-bottom 300ms ease-in;
-    }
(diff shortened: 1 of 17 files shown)


src/lib/utils/functions/FileHandleUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+    case CloudProviderFileHandleConcreteTypeEnum.GoogleCloudFileHandle:
+      return `gs://${(fileHandle as GoogleCloudFileHandle).bucketName}`
+    default:
+      throw new Error(
+        `Couldn't determine location name for file handle: ${JSON.stringify(
+          fileHandle,
+        )}`,
+      )
+  }
+}
```
nickgros, line 43: I'm not 100% sure if this is the right thing to do in every case

src/lib/style/components/_entity-metadata.scss
```diff
@@ -0,0 +1,72 @@
+@use '../abstracts/variables' as SRC;
+
+.EntityMetadata {
+  .Tabs {
```
jay-hodgson, line 4: I think this is a core style that we should share (for example, I should change the Download List v2 UI to use these instead of rolling my own).

jay-hodgson (approved):

PR #1084, opened 2021-06-17
Portals-1933: Account level icon layout issue

emmanmills:
Create react components for account icons to fix an issue with viewBox attribute not showing up.


jay-hodgson:
So happy that this solved the problem!  I find it interesting that changing the way the svg is imported avoids the svg _optimization_ in the UMD build.  Nice work.

PR #1086, opened 2021-06-21
PORTALS-1945

jay-hodgson:
Download Cart - added support for the different types of Actions Required.


jay-hodgson:
Screen Shot 2021-06-21 at 11 54 48 AM

Screen Shot 2021-06-21 at 11 55 27 AM

Screen Shot 2021-06-21 at 11 55 53 AM

src/__tests__/lib/containers/download_list_v2/MeetAccessRequirementCard.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { render, screen } from '@testing-library/react'
```
jay-hodgson, line 1: Basic tests, using the Mock Service Worker (msw) library to intercept rpcs.

src/lib/assets/icons/download_list/easy.svg
```diff
@@ -0,0 +1,33 @@
+<svg width="136" height="74" viewBox="0 0 136 74" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jay-hodgson, line 1: Icons that describe how difficult the action is.  Worked with Emily and Stockard to categorize the different scenarios.  See https://sagebionetworks.jira.com/browse/PORTALS-1947 for more info.

src/lib/containers/IconSvg.tsx
```diff
@@ -76,6 +77,8 @@ const getIcon = (options:IconSvgOptions) => {
       return <GetApp style={muiSvgStyle}></GetApp>
     case 'cart':
       return <ShoppingCartOutlined style={muiSvgStyle}></ShoppingCartOutlined>
+    case 'delete':
+      return <Delete style={muiSvgStyle}></Delete>
```
jay-hodgson, line 81: Used to Clear the Download List v2

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -42,10 +42,11 @@ type AccessRequirementAndStatus = {
 }
 
 export type AccessRequirementListProps = {
-  entityId: string
+  entityId: string  // will show this entity info
   accessRequirementFromProps?: Array<AccessRequirement>
   onHide?: Function
   renderAsModal?: boolean
+  numberOfFilesEffected?: number // if provided, will show this instead of the entity information
```
jay-hodgson, line 49: From the Actions Required context, the design calls for showing the number of files effected by the unmet AR instead of a specific entity.  For example, "12 Files" instead of "file1.bam", where the 12 files were files that the user added to their Download List v2.  For ManagedACTAccessRequirements, this has the potential to be a little clunky for the user, since the email notifications will point to a specific file (and we select essentially a random entity that is covered by this AR).  If the user goes through the process, they will be granted access to the target files, so I'm not suggesting for us to change the ManagedACTAccessRequirement services at this time.

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -343,7 +351,7 @@ export default function AccessRequirementList({
               )}
               {isSignedIn && (
                 <p>
-                  You have signed with Sage Platform (Synapse) user account as{' '}
+                  You have signed in with the Sage Platform (Synapse) user account {' '}
```
jay-hodgson, line 354: Small unrelated bug fix.  See https://sagebionetworks.jira.com/browse/SWC-5648

src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -0,0 +1,127 @@
+import * as React from 'react'
```
jay-hodgson, line 1: The cart page pull together the Actions Required list and the Available For Download files table.  It also handles the case when these are empty (based on the Download List v2 statistics), so those components are not loaded if they are empty.  This makes Clearing the list easy too (since it refetches the stats, and will not show the child components if they are empty).

src/lib/containers/download_list_v2/DownloadCartPageWrapper.md
```diff
@@ -0,0 +1,3 @@
+```jsx
+accessToken ? <DownloadCartPageWrapper /> : <>Sign in to view this component</>
```
jay-hodgson, line 2: Replaced subcomponent styleguidist demo files with the full Download Cart Page.

src/lib/containers/download_list_v2/DownloadListActionsRequired.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useEffect } from 'react'
```
jay-hodgson, line 1: Infinite scroll list of actions required.  For each row, will either render a MeetAccessRequirementCard (if there's an unmet AR to gain access to files in the Download List v2), or RequestDownloadCard (if, somehow, a user was able to add a file to the Download List for which they do not have the DOWNLOAD access type).

src/lib/containers/download_list_v2/DownloadListStats.tsx
```diff
@@ -36,7 +36,7 @@ export default function DownloadListStats() {
         </div>
       )}
       {isFetching && (
-        <div className="EntityFinderDetailsView__Placeholder">
+        <div className="placeholder">
```
jay-hodgson, line 39: Was reusing a style defined for the EntityFinderDetailsView.  Pulled this out into a mixin, and use it in a general "placeholder" css class.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -207,15 +204,10 @@ export default function DownloadListTable(props: DownloadListTableProps) {
         </ReactBootstrap.Table>
       )}
       {isFetching && (
-        <div className="EntityFinderDetailsView__Placeholder">
+        <div className="placeholder">
           <SynapseSpinner size={30} />
         </div>
       )}
-      {!isFetching && allRows.length === 0 && (
```
jay-hodgson, line 214: Handled by higher up component (the Download Cart Page)

src/lib/containers/download_list_v2/MeetAccessRequirementCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+import AccessRequirementList, { SUPPORTED_ACCESS_REQUIREMENTS } from '../access_requirement_list/AccessRequirementList'
+
+export type MeetAccessRequirementCardProps = {
+  accessRequirementId: number
+  count: number
+}
+export const TERMS_OF_USE_TITLE = 'Requires Acceptance of Data-Specific Terms of Use'
+export const SELF_SIGN_TITLE = 'Requires Acceptance of Data-Specific Terms of Use'
+export const ACT_TITLE = 'Requires Approval of Data-Specific Access Requirements'
+export const LOCK_TITLE = 'Access Restricted'
```
jay-hodgson, line 19: Pulled out the AR titles so that I could easily verify these were on the screen when expected (in the unit tests).

src/lib/containers/download_list_v2/MeetAccessRequirementCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <>
+      {!isError && !isFetching && content}
+      {isFetching && (
+        <div className="placeholder">
+          <SynapseSpinner size={30} />
+        </div>
+      )}
+      {isShowingAccessRequirement && (
+        <AccessRequirementList entityId={ar.subjectIds[0].id}
```
jay-hodgson, line 105: Clicking on the START button will show the AccessRequirementList dialog that contains that single Access Requirement.  An Access Requirement is associated to a number of subjects, but I simply pick the first entity (since I'm not sure what file in the download list is associated to this AR).

src/lib/containers/download_list_v2/RequestDownloadCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <div className="RequestDownloadCart actionRequiredCard">
+          <Icon type={VARIABLE_DIFFICULTY} />
+          <div className="metadata">
+            <div className="title">{REQUEST_DOWNLOAD_TITLE}</div>
+            <div className="fileCount">{count} File(s)</div>
+            <div className="description">
+              You must be granted the download permission on <strong>{entityHeader?.name}</strong> in order to download this set of files.
+            </div>
+          </div>
+          <a className="startButtonContainer" href={`https://www.synapse.org/#!Synapse:${entityHeader?.id}`} target='_blank' rel='noreferrer'>
```
jay-hodgson, line 50: This entity points to the "benefactor" for the file(s) that the user wants to download, which is typically a Folder or Project.  So this link will open up a new window and show the container that has the Sharing Settings that need to be modified (to grant the current user the DOWNLOAD access type)

src/lib/utils/APIConstants.ts
```diff
@@ -24,6 +24,8 @@ export const ENTITY_BUNDLE_V2 = (id: string | number, version?: number) =>
 export const ENTITY_ACCESS = (id: string | number) =>
   `${REPO}/entity/${id}/access`
 
+export const ENTITY_HEADERS = `${REPO}/entity/header`
```
jay-hodgson, line 27: Used in SynapseClient, and to set up msw response.

src/lib/utils/SynapseClient.ts
```diff
@@ -906,25 +905,25 @@ export const getEntityHeadersByIds = <T extends PaginatedResults<EntityHeader>>(
...
+})
 
   return doPost(
-    '/repo/v1/entity/header',
+    ENTITY_HEADERS,
     { references: fixedReferences },
     accessToken,
     undefined,
     BackendDestinationEnum.REPO_ENDPOINT,
-  ) as Promise<PaginatedResults<EntityHeader>>
```
jay-hodgson, line 927: Fix IDE warning.

src/lib/utils/SynapseClient.ts
```diff
@@ -2563,19 +2582,16 @@ export const searchEntities = (query: SearchQuery, accessToken?: string) => {
...
-export const getAvailableFilesToDownload = (
-  request: AvailableFilesRequest,
-  accessToken: string | undefined = undefined,
-): Promise<DownloadListQueryResponse> => {
-  const downloadListQueryRequest: DownloadListQueryRequest = {
-    concreteType:
-      'org.sagebionetworks.repo.model.download.DownloadListQueryRequest',
-    requestDetails: request,
-  }
+const getDownloadListJobResponse = (
```
jay-hodgson, line 2585: Reused for all Download List v2 async query job calls.
See http://rest-docs.synapse.org/rest/POST/download/list/query/async/start.html

src/lib/utils/SynapseClient.ts
```diff
@@ -2594,44 +2610,59 @@ export const getAvailableFilesToDownload = (
...
+
+/**
+ * Get Download List v2 available files to download
+ * http://rest-docs.synapse.org/rest/POST/download/list/query/async/start.html
+ */
+export const getAvailableFilesToDownload = (
+  request: AvailableFilesRequest,
+  accessToken: string | undefined = undefined,
+): Promise<DownloadListQueryResponse> => {
+  return getDownloadListJobResponse(accessToken, request)
```
nickgros, line 2638: Would it be possible to specify which QueryResponseDetails object (maybe with a generic) we get in each SynapseClient method instead of casting when the function is called?

src/lib/style/components/entity_finder/_details-view.scss
```diff
@@ -79,15 +80,7 @@
...
-    display: flex;
-    flex-direction: column;
-    justify-content: center;
-    align-items: center;
-    text-align: center;
-    height: 80%;
-    font-size: large;
-    margin: 15px;
-
+    @include SrcMixins.placeholder;
```
nickgros, line 83: Nice! Thanks for this!

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
nickgros: I'm unfamiliar with the ManagedACTAccessRequirements component, but this behavior makes sense to me in the context of the download list.

nickgros: Nit: I think effected -> affected

src/lib/containers/download_list_v2/AvailableForDownloadTable.tsx
```diff
@@ -2,16 +2,12 @@ import React from 'react'
...
 
 /**
  * Table of the files added to the Download Cart that are currently available for download.
  */
-export default function AvailableForDownloadTable(
-  props: DownloadListTableProps,
-) {
+export default function AvailableForDownloadTable() {
   const { accessToken } = useSynapseContext()
   const queryClient = new QueryClient({
```
nickgros, line 12: Unless you need the configured stale time, consider removing this QueryClient and the provider entirely to take advantage of having the cache shared outside of this component. We can also think about changing the default stale time.

Also, the staleTime comment is off by an order of magnitude 🙂

src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <div>
+              <div className="subSectionOverview">
+                <div className="container">
+                  <div className="subSectionContainer">
+                    <span className="subSectionTitle">Complete Your Download</span>
+                    <DownloadListStats />
+                  </div>
+                  <p className="description">Downloading your files programmatically is the quickest and most efficient way to get all of your files, 
+                  both internal and externally hosted. Metadata will always be included in your download automatically when downloading programmatically. 
+                  If you choose to download as .zip files, you can download external files individually at any time.</p>
```
nickgros, line 114: Not code related, but this copy is confusing to me and I think this needs another look before this is surfaced.

src/lib/containers/download_list_v2/DownloadCartPage.tsx
nickgros: Super cool to use the stats to determine to render the sub-components without having to do cache invalidation. I'm wondering what happens when

1. User has items in the download list (cache is populated in AvailableForDownloadTable with some nonzero staleTime)
2. User clears the download list (unmounting AvailableForDownloadTable)
3. User adds an item to the download list before the staleTime expires

If you used a persistent cache, I suspect a new AvailableForDownloadTable would retrieve the old download list.

Right now, I actually think it works because you don't use a persistent cache; you create a new QueryClient every time you render an AvailableForDownloadTable. At first glance, it also seems like none of the hooks you're using would benefit much from cache sharing, so you can ignore [this comment](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1086/files#r655661980). So I have no problem with this, but I think we should capture this complexity in a comment so we don't mess with this later and wonder why the cache no longer magically works.

src/lib/containers/download_list_v2/DownloadCartPageWrapper.tsx
```diff
@@ -2,18 +2,18 @@ import React from 'react'
...
  * "Wrapped" by a query client provider (for caching) and an error boundary.
  */
-export default function DownloadListStatsWrapper() {
+export default function DownloadCartPageWrapper() {
   const { accessToken } = useSynapseContext()
   const queryClient = new QueryClient({
     defaultOptions: {
       queries: {
-        staleTime: 60 * 1000, // 1m
+        staleTime: 20 * 1000, // 1m
```
nickgros, line 16: Is `useGetDownloadListStatistics` in the DownloadCartPage the only hook that uses this query client? Do you need that in its own cache? You could specify a shorter stale time as a parameter in the hook function if you need that.

src/lib/containers/download_list_v2/DownloadListActionsRequired.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, [status, isFetching, hasNextPage, fetchNextPage, inView])
+
+  const allRows = data
+    ? ([] as ActionRequiredCount[]).concat.apply(
+        [],
+        data.pages.map(
+          page => (page.responseDetails as ActionRequiredResponse).page,
+        ),
+      )
+    : []
```
nickgros, line 52: I think I came up with this monstrosity, maybe I should come up with something that makes more sense 🙂

src/lib/containers/download_list_v2/MeetAccessRequirementCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+  }: MeetAccessRequirementCardProps
+) => {
+  const handleError = useErrorHandler()
+  const {
+    data,
+    isFetching,
+    isError,
+    error: newError,
+  } = useGetAccessRequirement(accessRequirementId)
+  const ar:AccessRequirement = data as AccessRequirement
```
nickgros, line 34: Could you instead do this?

```suggestion
  const {
    data: ar,
    isFetching,
    isError,
    error: newError,
  } = useGetAccessRequirement(accessRequirementId)
```

I think you lose that it can be potentially undefined if you cast with `as AccessRequirement`. I think you've guarded against it sufficiently with `isError` and `isFetching`, but I'd rather let the TS compiler handle that in general when possible.

src/lib/containers/download_list_v2/RequestDownloadCard.tsx
nickgros: Kind of weird UX since we don't yet have any way to submit a request...I guess that's probably coming later?

src/lib/style/components/_download-cart-page.scss
```diff
@@ -0,0 +1,103 @@
...
+    align-items: center;
+    .active {
+      padding-top: 5px;
+      border-bottom: 5px solid;
+      border-bottom-color: SRC.$secondary-action-color;
+      button {
+        color: #515359;
+      }
+      .iconImg {
+        filter: invert(1)
```
nickgros, line 35: Does this work in all browsers we try to support? I think support for filter is weird

nickgros (approved):
Had some questions and comments but nothing that needs changed, looks great!

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
jay-hodgson: Right you are!

src/lib/containers/download_list_v2/RequestDownloadCard.tsx
jay-hodgson: That's the hope.  It's been a long-time issue, and does not have an easy answer (because of the way responsibilities are delegated on the curation side).  We may need a way to specify a principal as a data request handler?

src/lib/utils/SynapseClient.ts
jay-hodgson: If I understand what you're looking for correctly, you'd like it to be easier for the caller (so they don't need to get cast the QueryResponseDetails).  I'll update the code and you can let me know if you like it...

src/lib/containers/download_list_v2/DownloadCartPage.tsx
jay-hodgson: Agreed.  I think it was done in haste in the design mockup, but I consider it a work in progress.

src/lib/containers/download_list_v2/DownloadCartPageWrapper.tsx
jay-hodgson: I'll just remove the wrapper!

src/lib/style/components/_download-cart-page.scss
jay-hodgson: unknown, but I think this is a copy-paste issue.  removing from here.

PR #1089, opened 2021-06-25
PORTALS-1948: Includes necessary changes to integrate into SWC

jay-hodgson:


src/demo/containers/App.tsx
```diff
@@ -142,7 +142,7 @@ export default class App extends React.Component<{}, AppState> {
       <SynapseContextProvider
         synapseContext={{
           accessToken: token,
-          isInExperimentalMode: SynapseClient.getIsInExperimentalModeFromCookie(),
+          isInExperimentalMode: SynapseClient.isInSynapseExperimentalMode(),
```
jay-hodgson, line 145: Two functions did the same thing, removed getIsInExperimentalModeFromCookie()

src/lib/assets/icons/download_list/easy.svg
```diff
@@ -1,33 +0,0 @@
-<svg width="136" height="74" viewBox="0 0 136 74" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jay-hodgson, line 1: Moved these to the static Synapse S3 assets bucket since they were not rendering properly when used in SWC.

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -33,37 +35,54 @@ export type DownloadConfirmationState = {
...
 }
+
+type QueryOrFolder = {
+  queryBundleRequest?:QueryBundleRequest
+  folderId?: string
+}
+
 export type DownloadConfirmationProps = {
-  fnClose?: Function
+  fnClose?: () => void
```
jay-hodgson, line 47: TS warning

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -33,37 +35,54 @@ export type DownloadConfirmationState = {
...
   getLastQueryRequest?: () => QueryBundleRequest
+  folderId?: string
```
jay-hodgson, line 49: Added support for configuring this component with a folderId (might be a projectId), so that the user can add all child files of a container.

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -33,37 +35,54 @@ export type DownloadConfirmationState = {
...
-  const req: AddFilesToDownloadListRequest = {
-    concreteType:
-      'org.sagebionetworks.repo.model.file.AddFileToDownloadListRequest',
-    query: query,
-  }
   try {
-    const result = await SynapseClient.addFilesToDownloadList(req, token)
-    const ownerId = result.downloadList.ownerId
-    return [StatusEnum.SUCCESS, ownerId]
+    if (!SynapseClient.isInSynapseExperimentalMode()) {
```
jay-hodgson, line 65: If not in experimental mode, do what it used to (add files from a query, or now folder, to the old Download List).

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -33,37 +35,54 @@ export type DownloadConfirmationState = {
...
+      const req: AddFilesToDownloadListRequest = {
+        concreteType:
+          'org.sagebionetworks.repo.model.file.AddFileToDownloadListRequest',
+        query: queryBundleRequest?.query,
+        folderId: folderId,
+      }
+      await SynapseClient.addFilesToDownloadList(req, token)
+    } else {
+      const req:AddToDownloadListRequest = {
```
jay-hodgson, line 74: If in experimental mode, add to DL v2!

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -120,64 +139,86 @@ const StatusConstruct: UiStateDictionary = {
...
+  folderId,
   fnClose,
   updateParentState,
   topLevelControlsState,
   onExportTable,
+  downloadCartPageUrl='/DownloadCart',
 }) => {
   const { accessToken } = useSynapseContext()
   const { showDownloadConfirmation = true } = topLevelControlsState ?? {}
+  const [status, setStatus] = useState<StatusEnum>(accessToken ? StatusEnum.LOADING_INFO : StatusEnum.SIGNED_OUT)
```
jay-hodgson, line 151: Pulled out status into a standalone state variable.

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -120,64 +139,86 @@ const StatusConstruct: UiStateDictionary = {
...
-      status: StatusEnum.INFO,
-    })
-  }
+    updateStats(queryCount, sumFileSizes?.sumFileSizesBytes)
+  }, [queryOrFolder, setStatus, updateStats])
 
   // UseEffect memoization only works for arguments where a direct === comparison can be made
   // This fails drastically with the queryBundleRequest object which is a complex object of many fields that
   // change, we could use a custom comparitor but this also introduces risk
   useDeepCompareEffect(() => {
```
jay-hodgson, line 219: This is why this is a bit more complicated than I'd like.  The useDeepCompareEffect() requires a js Object (it will throw an error that tells you to instead use useEffect() if only primitives are provided).  In the case when the queryBundleRequest is undefined (if configured to add Folder/Project children), then this error is thrown at runtime :(
To avoid this, I created a new object that contains the folderId and queryBundleRequest, and define that in the dependencies for this hook.  Then everyone is happy :)

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -266,6 +303,13 @@ export const DownloadConfirmation: React.FunctionComponent<DownloadConfirmationP
         return <></>
     }
   }
+  if (showDownloadList && SynapseClient.isInSynapseExperimentalMode()) {
+    // go to the Download Cart Page
+    if (downloadCartPageUrl)
+      window.location.href = downloadCartPageUrl
```
jay-hodgson, line 309: For the first time, I'm going to have a route available to all portals (they do not need to define it in the config)

src/lib/containers/download_list/ShowDownload.tsx
```diff
@@ -30,8 +30,9 @@ function ShowDownload({ to }: ShowDownloadProps & RouteComponentProps) {
   const [showDownloadModal, setShowDownloadModal] = useState(false)
   const idForToolTip = 'SHOW_DOWNLOAD_TOOLTIP'
   const tooltipText = 'Click to view items in your download list.'
+  const isInExperimentalMode = SynapseClient.isInSynapseExperimentalMode()
   useEffect(() => {
-    if (!accessToken) {
+    if (!accessToken || isInExperimentalMode) {
```
jay-hodgson, line 35: Do not show V1 if in experimental mode.

src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -106,7 +110,7 @@ export default function DownloadCartPage() {
                 <div className="container">
                   <div className="subSectionContainer">
                     <span className="subSectionTitle">Complete Your Download</span>
-                    <DownloadListStats />
+                    <DownloadListStats numBytes={data.sumOfFileSizesAvailableForDownload} numFiles={data.numberOfFilesAvailableForDownload}/>
```
jay-hodgson, line 113: This was the cyclical loading loop that showed up when integrating into SWC.  The child component also asked for the Download List stats and when loaded it caused the child to be re-mounted and start the query again.  Result: Accidental DOS attack on Synapse.

src/lib/containers/download_list_v2/DownloadListActionsRequired.tsx
```diff
@@ -20,7 +20,6 @@ export default function DownloadListActionsRequired() {
     fetchNextPage,
     isError,
     error: newError,
-    refetch,
```
jay-hodgson, line 23: Fix warning (unused)

src/lib/containers/download_list_v2/DownloadListActionsRequired.tsx
```diff
@@ -82,7 +81,7 @@ export default function DownloadListActionsRequired() {
         <div
           className="DownloadListActionsRequired"
         >
-            {allRows.map(item => {
+            {allRows.map((item:ActionRequiredCount) => {
```
jay-hodgson, line 84: Fix warning (missing type)

src/lib/containers/download_list_v2/DownloadListStats.tsx
```diff
@@ -1,43 +1,24 @@
...
-    isFetching,
-    isError,
-    error: newError,
-  } = useGetDownloadListStatistics()
-  useEffect(() => {
-    if (isError && newError) {
-      handleError(toError(newError))
-    }
-  }, [isError, newError, handleError])
+export type DownloadListStatsProps = {
```
jay-hodgson, line 4: Now the stats component just takes the number of files and the byte value as props.

src/lib/containers/download_list_v2/ShowDownloadV2.tsx
```diff
@@ -65,11 +66,10 @@ function ShowDownload({ to }: ShowDownloadProps & RouteComponentProps) {
       />
     </>
   )
-
-  return <Link className="Download-Link SRC-userImgSmall" to={to}>
+  return <a className="Download-Link v2" href={to} rel="noopener noreferrer">
```
jay-hodgson, line 69: Using outside of React router (SWC)

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -121,9 +118,9 @@ const defaultIcons = {
   [CHART2]: chart2Svg,
   [FILE]: fileSvg,
   [EXTERNALLINK]: ExternalLink,
-  [EASY_DIFFICULTY]: easySvg,
-  [MEDIUM_DIFFICULTY]: mediumSvg,
-  [VARIABLE_DIFFICULTY]: variableSvg,
+  [EASY_DIFFICULTY]: 'https://s3.amazonaws.com/static.synapse.org/images/easy.svg',
+  [MEDIUM_DIFFICULTY]: 'https://s3.amazonaws.com/static.synapse.org/images/medium.svg',
+  [VARIABLE_DIFFICULTY]: 'https://s3.amazonaws.com/static.synapse.org/images/variable.svg',
```
jay-hodgson, line 123: This was surprisingly easy to fix.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -86,7 +85,6 @@ export interface Dictionary<T> {
 export type SynapseTableState = {
   sortedColumnSelection: SortItem[]
   columnIconSortState: number[]
-  isDownloadConfirmationOpen: boolean
```
jay-hodgson, line 89: This was never updated to true, so I removed.

src/lib/index.ts
```diff
@@ -97,6 +99,8 @@ const SynapseComponents = {
   ProjectViewCarousel,
   EntityFinder,
   ExperimentalMode,
+  DownloadCartPage,
+  ShowDownloadV2,
```
jay-hodgson, line 103: Exported for SWC.

jay-hodgson, line 103: Exported for portals.

src/lib/rollup.index.ts
```diff
@@ -48,6 +51,9 @@ const SynapseComponents = {
   SynapseHomepage,
   EntityFinder,
   ErrorPage,
+  DownloadCartPage,
+  ShowDownloadV2,
+  DownloadConfirmation
```
jay-hodgson, line 56: Exported for SWC.

jay-hodgson: (replacing the SWC implementation of the DownloadConfirmation with our SRC one!)

jay-hodgson:
Screen Shot 2021-06-25 at 9 56 00 AM
Screen Shot 2021-06-25 at 9 51 35 AM

Screen Shot 2021-06-25 at 9 56 09 AM
Screen Shot 2021-06-25 at 9 51 46 AM

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -120,64 +139,86 @@ const StatusConstruct: UiStateDictionary = {
...
   const { showDownloadConfirmation = true } = topLevelControlsState ?? {}
+  const [status, setStatus] = useState<StatusEnum>(accessToken ? StatusEnum.LOADING_INFO : StatusEnum.SIGNED_OUT)
   const [state, setState] = useState<DownloadConfirmationState>({
     fileCount: 0,
-    fileSize: 0,
-    status: accessToken ? StatusEnum.LOADING_INFO : StatusEnum.SIGNED_OUT,
+    fileSize: 0
   })
+  const [queryOrFolder] = useState<QueryOrFolder>({
+    queryBundleRequest: getLastQueryRequest!(),
```
emmanmills, line 157: How come this one doesn't have a `setQueryOrFolder` variable?

jay-hodgson: Right, it never updates this value.  I thought if I used a const then the object would be redefined on each render (which would cause the useEffect to fire).  This was my attempt to single object instance between renderings to avoid re-render (similar to why people use the useCallback() wrapper for inline functions).  Please let me know if there's a better way to do this!

emmanmills: Ah, I see. That makes sense. Thanks.

emmanmills (approved):
Looks good!

PR #1092, opened 2021-06-25
PORTALS-1941: Add accessor renewal radio buttons

emmanmills:


src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -66,6 +68,10 @@ export type AlertProps = {
   message: string | JSX.Element
 }
 
+type requestedFileTypesMap = {
+  [key: string]: string[]
+}
+
```
emmanmills, line 74: Create this type to replace the `any` type to fix eslint warning.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -403,12 +431,20 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
...
       })
     })
+
+    // Update accessor renewal radio buttons if available
+    if (isRenewal) {
+      const filteredRadio:AccessorChange[] = accessorRadioValue.filter(
+        item => item.userId !== pid
+      )
+      setAccessorRadioValue(filteredRadio)
+    }
```
emmanmills, line 441: Remove the renew/revoke radio buttons if an accessor is deleted.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -120,8 +128,28 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
     // get data access required docs data to display file names
     getFilesData(dataAccessRequestData)
 
+    // renewal case
     if (dataAccessRequestData.concreteType === 'org.sagebionetworks.repo.model.dataaccess.Renewal') {
       setIsRenewal(true)
+      if (dataAccessRequestData.accessorChanges?.length) {
```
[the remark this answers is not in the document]
jay-hodgson: This seems generally dangerous.  In the case of a Renewal, could you confirm that these are not already set to RENEW_ACCESS from the backend?  
My concern is when a user saves a request that's in progress.  I think newly added accessors (who should have GAIN_ACCESS), and existing accessors who you've previously changed to REVOKE_ACCESS, will be flipped to RENEW_ACCESS when this UI is reopened.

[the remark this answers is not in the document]
jay-hodgson: If they initially come back as RENEW_ACCESS, then this code can be safely deleted.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
jay-hodgson: This should never happen, since you should only be able to delete newly added accessors (accessors who are going to have GAIN_ACCESS in the request).  For existing accessors, there should be no option to delete - the submitter needs to pick Renew or Revoke (Revoke is the auditable delete).

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -509,6 +540,25 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
     }
   }
 
+  // For renewal only
+  const onAccessorRadioBtnChange = (accessType: AccessType, userId: string) => {
```
jay-hodgson, line 544: To verify, one of these options should be selected by default.  My hope is that the backend is returning "RENEW_ACCESS" for existing accessors!

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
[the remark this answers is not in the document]
emmanmills: I do see RENEW_ACCESS coming from the backend if a data access request has been approved. So I have updated the code to copy the user access types from the backend API to the state variable rather than setting them all to RENEW_ACCESS.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
emmanmills: This is fixed. A user that has a renewal data access type will not have the delete "X" button.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
emmanmills: The renew radio button will be selected by default if the user has a renewal data access type.

src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -79,14 +90,15 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
...
-  const [accessorProfiles, setAccessorProfiles] = useState<UserProfile[]>([])
   const [
     formSubmitRequestObject,
     setFormSubmitRequestObject,
-  ] = useState<RequestInterface|RenewalInterface>()
+  ] = useState<RequestInterface | RenewalInterface>()
   const [alert, setAlert] = useState<AlertProps | undefined>()
   const [isRenewal, setIsRenewal] = useState<boolean>(false)
-  const requestedFileTypes = {}
+  const [accessors, setAccessors] = useState<Accessor[]>([])
```
emmanmills, line 99: Update the state variable to not just contain a list of user profiles (for display usernames) but also the user's access type (for display radio buttons If `isRenewal === true`)

jay-hodgson (approved):

PR #1104, opened 2021-07-02
Release 2.0.10

jay-hodgson:


jay-hodgson:
Unstable test, need to comment out if we cannot stabilize.

> it works at integration level testing › Call to invalid rest endpoint results in error

PR #1106, opened 2021-07-02
PORTALS-1673: sort by assay name, and fix demo of upset plot (lock down to a view version)

jay-hodgson:


emmanmills (approved):
Looks good!

PR #1110, opened 2021-07-07
PORTALS-1956

jay-hodgson:


src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -153,13 +153,21 @@ export const DownloadConfirmation: React.FunctionComponent<DownloadConfirmationP
     fileCount: 0,
     fileSize: 0
   })
-  const [queryOrFolder] = useState<QueryOrFolder>({
-    queryBundleRequest: getLastQueryRequest!(),
+  const lastQueryRequest = getLastQueryRequest!()
```
nickgros, line 156: is `getLastQueryRequest` memoized? If not, `lastQueryRequest` may change on each render and would kill performance with your effect below

nickgros (approved):

src/lib/containers/download_list/DownloadConfirmation.tsx
jay-hodgson: Verified new code in useDeepCompareEffect() is called only when I'd expect it (which is very rare).

PR #1123, opened 2021-07-13
PORTALS-1964: if event metadata is rootClose, then close the popup menu

jay-hodgson:


src/lib/containers/table/table-top/ColumnSelection.tsx
```diff
@@ -30,34 +33,24 @@ export const ColumnSelection: React.FunctionComponent<ColumnSelectionProps> = (
...
-    facetAliases,
-    show: showFromProps,
-    onChange,
+    facetAliases
   } = props
 
-  let [show, setShow] = useState(false)
-
-  let [usedShow, usedSetShow] =
-    showFromProps !== undefined ? [showFromProps, onChange!] : [show, setShow]
```
jay-hodgson, line 41: I think this was previously used by other components.  Now it's a subcomponent of TopLevelControls, which is a subcomponent of QueryWrapperPlotNav.  From what I see in the codebase, the show from props (showFromProps) is no longer set.

src/lib/containers/table/table-top/ColumnSelection.tsx
```diff
@@ -30,34 +33,24 @@ export const ColumnSelection: React.FunctionComponent<ColumnSelectionProps> = (
...
-
-  const onDropdownClick = (_show: boolean, _event: any, metadata: any) => {
-    // Any click event for the Dropdown will close the dropdown (assuming its open), so we have
-    // to handle the onToggle event and manually manage the dropdown open state. If metadata
-    // is defined the event occuring is inside the dropdown which we then want to keep open, otherwise
-    // we close it.
-    if (metadata.source) {
-      usedSetShow(true)
+  const [show, setShow] = useState(false)
+  const onDropdownClick = (_show: boolean, _event: React.SyntheticEvent<Dropdown, Event>, metadata: MetadataEvent) => {
```
jay-hodgson, line 40: Improved types, and recognize "rootClose" source value to close the popup menu.

nickgros (approved):
Looks good. The MetadataEvent type really helps clarify how this is supposed to work.

PR #1128, opened 2021-07-16
PORTALS-1607

jay-hodgson:


src/lib/style/base/_core.scss
```diff
@@ -1223,3 +1216,74 @@ hr ~ .SRC-wrapper {
...
+    }
+  }
+
+  // do not apply special styling to dropdown links or anchors in tabs
+  .dropdown, .nav-tabs, .dropdown-menu {
+    a:not(.btn) {
+      @include ignoreLink;
+    }
+  }
+}
```
emmanmills, line 1289: Have you checked if the new link style will render correctly for the top menu in mobile view (< 768px)? Thanks.

Screen Shot 2021-07-16 at 2 00 51 PM

jay-hodgson: Menu looks ok to me.  Thanks for bringing this up though, the style change could certainly have unintended side effects in buried UI like this.

jay-hodgson: Screen Shot 2021-07-16 at 2 17 19 PM

emmanmills (approved):
Thanks, looks good to me!

PR #1130, opened 2021-07-20
PORTALS-1962: Remove alpha mode for managed ACT access requirement form

emmanmills:


jay-hodgson (approved):

jay-hodgson:
Please do not merge this until after the next synapse.org release (we might want to patch staging this week)

PR #1131, opened 2021-07-21
PORTALS-1970

nickgros:
Fixing custom styling for checkboxes and radio groups.

Using our custom components:
image

Synapse Form:

image

image

QueryWrapperPlotNav:

image

### DIFF
--- a/src/lib/containers/widgets/Checkbox.tsx
+++ b/src/lib/containers/widgets/Checkbox.tsx
@@ -9,14 +9,18 @@ export type CheckboxProps = {
   isSelectAll?: boolean
   children?: React.ReactChild
   onClick?: (event: React.SyntheticEvent<HTMLDivElement>) => void
+  disabled?: boolean
 }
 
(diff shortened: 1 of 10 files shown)


src/lib/containers/widgets/Checkbox.tsx
```diff
@@ -9,14 +9,18 @@ export type CheckboxProps = {
...
   props: CheckboxProps,
 ) => {
-  const { checked: propsChecked = false, isSelectAll = false } = props
+  const {
+    checked: propsChecked = false,
+    isSelectAll = false,
+    disabled = false,
+  } = props
   const [checked, setChecked] = useState<boolean>(propsChecked)
-  const [focused, setFocused] = useState<boolean>(false)
```
nickgros, line 19: Removed 'focus' state from both checkboxes and radios. We now just look at the `:focus` state of the input element instead of relying on a class set by React.

jay-hodgson:
Looks great, thanks Nick

PR #1133, opened 2021-07-22
SWC-5664: separate parent container param from query request

jay-hodgson:
These are the two use cases, now integration tested in SWC

### DIFF
--- a/src/lib/containers/download_list/DownloadConfirmation.tsx
+++ b/src/lib/containers/download_list/DownloadConfirmation.tsx
@@ -22,6 +22,7 @@ import { useSynapseContext } from '../../utils/SynapseContext'
 import { AddToDownloadListRequest } from '../../utils/synapseTypes/DownloadListV2/AddToDownloadListRequest'
 import { useGetEntityChildren } from '../../utils/hooks/SynapseAPI/useGetEntityChildren'
 import useGetQueryResultBundle from '../../utils/hooks/SynapseAPI/useGetQueryResultBundle'
+import { useGetDownloadListStatistics } from '../../utils/hooks/SynapseAPI/useGetDownloadListStatistics'
 
 enum StatusEnum {
   LOADING_INFO,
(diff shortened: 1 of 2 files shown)


jay-hodgson:
This is for a feature that is behind the alpha flag on the portals and synapse.org

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -22,6 +22,7 @@ import { useSynapseContext } from '../../utils/SynapseContext'
 import { AddToDownloadListRequest } from '../../utils/synapseTypes/DownloadListV2/AddToDownloadListRequest'
 import { useGetEntityChildren } from '../../utils/hooks/SynapseAPI/useGetEntityChildren'
 import useGetQueryResultBundle from '../../utils/hooks/SynapseAPI/useGetQueryResultBundle'
+import { useGetDownloadListStatistics } from '../../utils/hooks/SynapseAPI/useGetDownloadListStatistics'
```
jay-hodgson, line 25: Refetch the statistics after adding items to the cart (this will update other UI on the site, like the download notification count in the header)

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -200,14 +188,14 @@ export const DownloadConfirmation: React.FunctionComponent<DownloadConfirmationP
     if (isSuccess && entityChildrenData) {
       updateStats(entityChildrenData.totalChildCount, entityChildrenData.sumFileSizesBytes)
     }
-  }, [updateStats, queryOrFolder, isSuccess, entityChildrenData])
+  }, [updateStats, folderId, isSuccess, entityChildrenData])
```
jay-hodgson, line 191: This effect is specific to the folder (to get statistics about the children of the container)

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -219,7 +207,7 @@ export const DownloadConfirmation: React.FunctionComponent<DownloadConfirmationP
     if (queryResultBundle) {
       updateStats(queryResultBundle.queryCount, queryResultBundle.sumFileSizes?.sumFileSizesBytes)
     }
-  }, [updateStats, queryOrFolder, queryResultBundle])
+  }, [updateStats, lastQueryRequest, queryResultBundle])
```
jay-hodgson, line 210: And this effect is specific to the query request (get the stats on the rowset that this query would return)

emmanmills (approved):
Looks good!

PR #1135, opened 2021-07-23
SWC-5675: load profile pictures on-the-fly

jay-hodgson:
preview pictures only, as needed


jay-hodgson:
Verified that integration with Synapse.org still works (we hand over the same type of picture URL, one that redirects to the presigned URL).  And it now loads the preview image as needed (on popover for example) from a stable URL that relies on a redirect to the actual presigned URL.
Screen Shot 2021-07-22 at 8 23 37 PM

emmanmills (approved):
Looks good!

PR #1141, opened 2021-07-26
SWC-5678

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5678)

Improving the error message when the annotation key collides with an entity property.

image


src/lib/utils/synapseTypes/Entity.ts
```diff
@@ -54,9 +55,9 @@ export interface EntityJson extends Record<string, unknown> {
 }
 
 /**
- * A string array of the keys in the EntityJson object above.
+ * A string array of all possible keys used by Synapse in Entity objects (objects that inherit this interface: https://docs.synapse.org/rest/org/sagebionetworks/repo/model/Entity.html).
  * This object is used to determine which fields are standard and which are annotations,
- * so it's important that this array contains all of the keys in the object above.
+ * so it's important that this array contains all keys in the objects that implement the linked interface above.
  */
 export const entityJsonKeys = [
```
nickgros, line 62: We use these to warn/disallow creating annotations that will be overwritten by entity properties. Should we fetch these at runtime using [GET /REST/resources/effectiveSchema](https://docs.synapse.org/rest/GET/REST/resources/effectiveSchema.html)?

Seems like another justification for [PLFM-6760](https://sagebionetworks.jira.com/browse/PLFM-6760) 🙂

jay-hodgson (approved):

PR #1143, opened 2021-07-27
Apply the id for the radio group to each option

nickgros:


emmanmills (approved):
I tested this locally and it works!

PR #1146, opened 2021-07-28
SWC-5691

nickgros:
https://sagebionetworks.jira.com/browse/SWC-5691

### DIFF
--- a/src/lib/containers/UserCardSmall.tsx
+++ b/src/lib/containers/UserCardSmall.tsx
@@ -36,9 +36,10 @@ export const UserCardSmall: React.FunctionComponent<UserCardSmallProps> = ({
   className,
   ...rest
 }) => {
-
   const [userBundle, setUserBundle] = useState<UserBundle | undefined>()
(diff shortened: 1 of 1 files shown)


src/lib/containers/UserCardSmall.tsx
```diff
@@ -110,6 +111,7 @@ export const UserCardSmall: React.FunctionComponent<UserCardSmallProps> = ({
 
       <a
         ref={target}
+        href={disableLink ? undefined : link}
```
jay-hodgson, line 114: On click, does it both change the current window to the link url and open it up in a new window?

jay-hodgson: (looking at the window.open(link, '_blank') in the click handler)

jay-hodgson (approved):

src/lib/containers/UserCardSmall.tsx
nickgros: Just opens the profile in a new window. `event.preventDefault()` prevents the link from opening in the new window (at least in Firefox, Chrome, and Safari)

PR #1149, opened 2021-07-29
Add new Annotation Editor to Entity Modal + other changes

nickgros:
* Adds the annotation editor to the EntityModal. In Synapse.org, the 'Annotations' button will open the EntityModal, directly opening the annotations tab (the tab buttons will not be rendered, so functionally it will just be an annotations modal). If the user has edit access, they will see an edit button that opens the editor in the same modal.
* The annotations table now uses the JSON representation of the annotations rather than the bundle's annotations
  * The data is now instantly updated in the cache when successfully updated, so no need to re-fetch the bundle
  * Reduces chances of exposing inconsistencies between JSON and Annotations APIs (e.g., you can set annotations like 'name' in the old API).
  * There were performance concerns about using the JSON API for reads, but it will soon be backed by a cache in the backend, so it shouldn't be an issue.
* Add Material UI [Skeletons](https://material-ui.com/components/skeleton/)

https://user-images.githubusercontent.com/17580037/129583270-714a4302-e22c-4bf4-bde1-0490aae06179.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@fortawesome/react-fontawesome": "^0.1.3",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
+    "@material-ui/lab": "^4.0.0-alpha.57",
     "@rjsf/core": "^3.0.0",
     "@sage-bionetworks/rjsf-core": "^3.1.3",
     "@upsetjs/react": "^1.6.2",
--- a/src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
+++ b/src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
@@ -35,6 +35,8 @@ export type SchemaDrivenAnnotationEditorProps = {
   schemaId?: string
   liveValidate?: boolean
   onSuccess?: () => void
+  /** If defined, shows a 'Cancel' button and runs this effect on click */
+  onCancel?: () => void
 }
 
 export type SchemaDrivenAnnotationEditorModalProps = {
@@ -72,6 +74,7 @@ export const SchemaDrivenAnnotationEditor: React.FunctionComponent<SchemaDrivenA
   onSuccess = () => {
     /* no-op */
   },
+  onCancel,
 }: SchemaDrivenAnnotationEditorProps) => {
   const formRef = useRef<Form<Record<string, unknown>>>(null)
 
@@ -145,9 +148,11 @@ export const SchemaDrivenAnnotationEditor: React.FunctionComponent<SchemaDrivenA
   )
 
   return (
-    <div className="bootstrap-4-backport">
+    <div className="bootstrap-4-backport AnnotationEditor">
       {isLoading ? (
-        <SynapseSpinner />
+        <div className="LoadingPlaceholder">
+          <SynapseSpinner size={30} />
+        </div>
       ) : (
         <>
           <ReactTooltip id={ANNOTATION_EDITOR_TOOLTIP_ID} />
@@ -188,7 +193,7 @@ export const SchemaDrivenAnnotationEditor: React.FunctionComponent<SchemaDrivenA
             </Alert>
           )}
           <Form
-            className="AnnotationEditor"
+            className="AnnotationEditorForm"
             liveValidate={liveValidate}
             noHtml5Validate={true}
             ArrayFieldTemplate={CustomArrayFieldTemplate}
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "@fortawesome/react-fontawesome": "^0.1.3",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
+    "@material-ui/lab": "^4.0.0-alpha.57",
```
nickgros, line 26: Adding material lab for [Skeletons](https://material-ui.com/components/skeleton/)

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -108,22 +98,16 @@ export const SchemaDrivenAnnotationEditor: React.FunctionComponent<SchemaDrivenA
...
   useEffect(() => {
-    if (json) {
-      /**
-       * To only show annotations in the form, we must remove non-annotation fields.
-       * We will need to submit these values when we create the entity, so we set them aside
-       * and will merge objects later.
-       */
-
-      setEntityJson(getStandardEntityFields(json))
-      setFormData(removeStandardEntityFields(json))
```
nickgros, line 124: Separating the data into annotation and non-annotation data is now done within the custom hook

src/lib/style/components/_annotations.scss
```diff
@@ -2,108 +2,119 @@
 
 .bootstrap-4-backport {
   .AnnotationEditor {
-    display: grid;
```
nickgros, line 5: Diff view makes this change look crazier than it is, just added a couple of classes for the loading spinner and button arrangements

src/lib/utils/hooks/SynapseAPI/useEntity.ts
```diff
@@ -12,18 +14,56 @@ import {
...
+      setAnnotations(removeStandardEntityFields(query.data))
+    }
+  }, [query.data])
+
+  return {
+    ...query,
+    entityMetadata,
+    annotations,
+  }
 }
```
nickgros, line 68: Now all subscribers to this hook (i.e. both the annotation table and the annotation editor) can immediately discern annotations from standard metadata

src/lib/containers/entity/metadata/EntityModal.tsx
```diff
@@ -35,21 +36,26 @@ export const EntityModal: React.FC<EntityModalProps> = ({
...
     >
       <Modal.Header closeButton>
-        <Modal.Title>
-          {isLoadingBundle ? <SynapseSpinner /> : entityBundle?.entity?.name}
-        </Modal.Title>
+        {entityBundle ? (
+          <Modal.Title>{entityBundle.entity!.name}</Modal.Title>
+        ) : (
+          <Skeleton>
+            <Modal.Title>Placeholder</Modal.Title>
```
jay-hodgson, line 56: From the title, it looks like this text is not actually rendered (but only used for the string length).  Is that right?

nickgros: Exactly!

jay-hodgson: Ideally I wouldn't want to see placeholder dummy text all over the project (if we adopt this UI, which I think we should since it's better than a spinner!).

src/lib/utils/hooks/SynapseAPI/useEntity.ts
```diff
@@ -12,18 +14,56 @@ import {
 import { SynapseClient } from '../..'
 import { SynapseClientError } from '../../SynapseClient'
 import { useSynapseContext } from '../../SynapseContext'
-import { EntityJson } from '../../synapseTypes'
+import { EntityJson, entityJsonKeys, EntityJsonValue } from '../../synapseTypes'
 
+export function getStandardEntityFields(json: EntityJson): EntityJson {
+  return pick(json, entityJsonKeys) as EntityJson
```
jay-hodgson, line 20: I suppose this is where entityJsonKeys is the union of all possible entity object field names.  I think this is fine, but should we be worried about the edge cases (like a FileEntity annotated with repositoryName)?

nickgros: It would cause a problem if we conditionally render content using presence of the property, rather than the concrete type (which I think we do quite a bit).

[We do show an error in the new editor if a user tries to save such a property](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/ca0ebdbdcc59d61781999db21a5eeb8d176ba0d0/src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx#L242), but they can still click through the error. The backend doesn't block it.

nickgros: We would also only see this issue if we use the JSON service, which (for now), we only use for getting/setting annotations

jay-hodgson (approved):

PR #1164, opened 2021-08-12
SWC-5690: add instructions for R and Python download (along with note…

jay-hodgson:
…s for each client)

### DIFF
--- a/src/lib/containers/table/table-top/ProgrammaticOptions.tsx
+++ b/src/lib/containers/table/table-top/ProgrammaticOptions.tsx
@@ -6,8 +6,15 @@ import {
 import { TransformSqlWithFacetsRequest } from '../../../utils/synapseTypes/Table/TransformSqlWithFacetsRequest'
 import useDeepCompareEffect from 'use-deep-compare-effect'
 import { getTransformSqlWithFacetsRequest } from '../../../utils/SynapseClient'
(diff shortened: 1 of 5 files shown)


src/lib/containers/table/table-top/ProgrammaticOptions.tsx
```diff
@@ -56,39 +63,91 @@ function ProgrammaticOptions({
...
 
+  const installationInstructions = <p>Installation instructions are available at our
+    <a
+      className="ProgrammaticOptions__docslink"
+      href="https://help.synapse.org/docs/Installing-Synapse-API-Clients.1985249668.html"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      {' '}
+      Synapse API Documentation Site
```
emmanmills, line 74: I am curious why you want to add a space character inside the anchor tag instead of outside?

jay-hodgson: Good question!  This space character was in the code previously (see the old line 84), and I didn't clean it up :) I'll take care of it, thanks @emmanmills !

jay-hodgson: Pulled out the space character, and verified the output UI looked ok.

emmanmills (approved):
Looks good!

PR #1176, opened 2021-08-18
PORTALS-1994: increase specificity

jay-hodgson:


src/lib/containers/UserCardMedium.tsx
```diff
@@ -246,7 +246,11 @@ export const UserCardMedium: React.FC<UserCardMediumProps> = ({
             tabIndex={0}
           >
             <span style={{ paddingRight: '5px', paddingBottom: '1px' }}>
-              <a className="link">
+              <a className={`link ${
+                isLarge
+                  ? 'SRC-whiteText'
```
jay-hodgson, line 251: SRC-whiteText was applied to a parent element and previously changed the color of this child anchor, but now that all links have been restyled it's no longer being respected.  This forces the color change of the anchor text.

nickgros (approved):

PR #1177, opened 2021-08-19
SWC-5680 - Add global toast notification system, push toast when conditional key is dropped from schema

nickgros:
The toast system is powered by [react-hot-toast](https://react-hot-toast.com/). I chose react-hot-toast over the most popular/feature complete option, react-toastify, for the following reasons
* react-hot-toast had all the features we needed
* react-toastify didn't cleanly handle animating toasts coming from the bottom of the screen, which is how we currently animate toasts in Synapse.org. I don't think this could be fixed with additional CSS.
* Some concerns about frequency of recent updates and documentation-completeness

In general though, the two systems work similarly and we could pretty easily switch to `react-toastify` with a small amount of effort if desired.

To use the toast system, you need to place the SynapseToastContainer in the page (for example, as a child of the `body` element). Then you can call the `displayToast` function from anywhere. The `displayToast` function just uses the FullWidthAlert component. 

I made a small change to FullWidthAlert to prevent wrapping text prematurely if there are no action buttons/links.


src/lib/containers/FullWidthAlert.tsx
```diff
@@ -56,57 +53,72 @@ function FullWidthAlert(props: FullWidthAlertProps) {
...
-        timer = setTimeout(onClose, autoCloseAfterDelayInSeconds * 1000)
-      }
-      return () => {
-        if (timer) {
-          clearTimeout(timer);
-        }
+
+  const hasActions =
+    (primaryButtonText && onPrimaryButtonClick) ||
+    (secondaryButtonText && secondaryButtonHref)
```
nickgros, line 62: Check to see if there's a primary/secondary button. If not, we apply an additional class name to the component

src/lib/containers/FullWidthAlert.tsx
```diff
@@ -56,57 +53,72 @@ function FullWidthAlert(props: FullWidthAlertProps) {
...
+    <Alert
+      variant={variant}
+      show={show}
+      dismissible={false}
+      transition={transition}
+      className={`FullWidthAlert bootstrap-4-backport ${
+        isGlobal ? 'global' : ''
+      } ${additionalAlertVariantClass}`}
+    >
+      <div className={`gridContainer ${hasActions ? '' : 'noActions'}`}>
```
nickgros, line 86: Apply the 'noActions' class name to the grid container if there's no primary/secondary button

src/lib/style/components/_global-alert.scss
```diff
@@ -1,71 +1,83 @@
...
+    grid-template-rows: auto;
+    grid-template-columns: 50px auto 50px;
+    .closeAlert {
+      grid-column: 3 / span 1;
+    }
+    .messageArea {
+      grid-column: 2 / span 1;
+      max-width: unset;
+    }
+  }
```
nickgros, line 82: If there's no primary/secondary button in the FullWidthAlert
* We only need one row
* 3 columns: icon, text, and space for a close button
* Remove the max width from the message area, since we're explicitly making this change to use all of the space

image

styleguide.setup.js
```diff
@@ -45,6 +66,11 @@ global.iconOptions = {
   'Resilience-AD': resilienceadSvg,
 }
 
+// Inject the ToastContainer so we can push toast notifications
+let toastContainerDiv = document.createElement('div')
+document.getElementsByTagName('body')[0].append(toastContainerDiv)
+ReactDOM.render(React.createElement(SynapseToastContainer), toastContainerDiv)
```
nickgros, line 72: Injecting a single SynapseToastContainer into the Styleguidist page so we can push toasts from demos

styleguide.setup.js
```diff
@@ -55,3 +81,4 @@ global.COMPUTATIONAL = COMPUTATIONAL
 global.PUBLICATION = PUBLICATION
 global.Checkbox = Checkbox
 global.ReactBootstrap = ReactBootstrap
+global.ReactQueryDevtoolsPanel = ReactQueryDevtoolsPanel
```
nickgros, line 84: Not used in this PR, but having this available makes it easier to debug the react-query cache for a single demo.

We don't add it to the page like the toast container because we don't want it available/visible by default.

We can't add it to the demo wrapper because one will render for each wrapper, when you only want one, and they won't be usable.

package.json
```diff
@@ -27,6 +27,7 @@
     "@rjsf/core": "^3.0.0",
     "@sage-bionetworks/rjsf-core": "^3.1.3",
     "@upsetjs/react": "^1.6.2",
+    "animate.css": "^4.1.1",
```
nickgros, line 30: Adding animate.css to replicate the fadeInUp/fadeOutDown animations we use for toasts in SWC. Just need to make sure this doesn't cause conflicts with animations in SWC.

src/lib/style/components/_toasts.scss
```diff
@@ -0,0 +1,21 @@
+.SynapseToastContainer {
+  inset: unset !important;
```
nickgros, line 2: I was pretty heavy-handed with the `!important`s here but figured it's not a huge deal since it's only affecting our toast messages. Let me know if you think this is a problem and I can spend some more time trying to work around it.

src/lib/containers/entity/annotations/CustomObjectFieldTemplate.tsx
```diff
@@ -11,9 +13,62 @@ import AddToList from '../../../assets/icons/AddToList'
...
+   * If a property exists in the previous version of the schema and has user data, and then is dropped from the schema due to a data update, then
+   * we need to prompt the user whether they want to undo the change, or continue and keep/remove those fields
+   *
+   * In this component, we can identify when one or more fields with user data are dropped (the field gains the additional property flag in the schema prop).
+   * We then use a function provided by the context to report which fields were lost.
+   */
+  useDeepCompareEffectNoCheck(() => {
+    if (props.schema.properties) {
+      const propertyKeys = Object.keys(props.schema.properties)
+      // Schema-defined properties are those properties in the schema without the additional property flag.
```
jay-hodgson, line 39: And this ADDITIONAL_PROPERTY_FLAG is our workaround for additional "custom" annotations (those annotations that do not conform to the schema)?

src/lib/containers/entity/annotations/CustomObjectFieldTemplate.tsx
```diff
@@ -11,9 +13,62 @@ import AddToList from '../../../assets/icons/AddToList'
...
+
+      if (previousSchemaDefinedProperties != null) {
+        // Compare the schema defined properties with the previous version to see if any were lost.
+
+        const lostProperties = Array.from(
+          previousSchemaDefinedProperties,
+        ).filter(
+          schemaDefinedProperty =>
+            !schemaDefinedProperties.has(schemaDefinedProperty) &&
+            props.formData[schemaDefinedProperty] != null, // if the data is null, then we don't need to worry about it; user data isn't lost
```
jay-hodgson, line 55: Thank you for the comments, this makes sense.

src/lib/style/components/_global-alert.scss
jay-hodgson: Excellent.

styleguide.setup.js
jay-hodgson: Nice.  How would one add this ReactQueryDevtoolsPanel to debug the react-query cache for a single demo, exactly?

jay-hodgson (approved):

src/lib/containers/entity/annotations/CustomObjectFieldTemplate.tsx
nickgros: The flag is applied and used by react-jsonschema-form internally to denote which properties are not defined by the schema. We can inspect it here since this is a RJSF template component.

So yes, this is a workaround in the sense that using this means we don't have to duplicate the work of computing the schema without conditionals (and also saves us from writing the code to do that correctly).

styleguide.setup.js
nickgros: You just need to add the component to the demo you're testing--it's not a perfect UX but it should be good enough for debugging

Could be worth adding a simple demo just to document that we can do this

image

PR #1180, opened 2021-08-20
Multiple toast fixes, add toast demo

nickgros:
* Increase the exit animation length to 1s. 0.5 seconds is too short, which can cause the notification to pop back in after the animation ends but before the element has been removed from the DOM
* Set the z-index of the toast container to 2001 (matches existing toasts in SWC)
* Added a demo for creating sample toasts
* Add an additional class to prevent primary button/close button overlap on wide screens:

Before: 
image
After:
image


src/demo/containers/playground/ToastDemo.md
```diff
@@ -0,0 +1,3 @@
+```js
+<ToastDemo />
```
jay-hodgson, line 2: Could you also include an example code snippet here that uses displayToast() (probably would need to make it globally available in the StyleGuidist setup)?

jay-hodgson (approved):

src/demo/containers/playground/ToastDemo.md
nickgros: Sure, I'll put the demo code here instead of wrapping it in a component

PR #1182, opened 2021-08-20
PORTALS-1972: Only show the No Results UI if there are no rows AND this is the first page of data (offset  is 0)

jay-hodgson:


emmanmills (approved):
Looks good!

PR #1183, opened 2021-08-23
SWC-5721: New Synapse entity tab icons, and export IconSvg so that it can be independently used

jay-hodgson:


emmanmills (approved):
Looks good!

PR #1184, opened 2021-08-23
Release 2.0.32

jay-hodgson:


jay-hodgson:
Fixes a couple of Toast message bugs, and exports Icons for use in the secondary Entity tab navigation on Synapse.org.

PR #1186, opened 2021-08-24
SWC-5731 - Annotation editor: use custom widget for enumerations

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5731)

We're using [react-select](https://github.com/JedWatson/react-select) since it lets you enter a custom value (using the `Creatable` component). It's very powerful, but we're just using a subset of its features.

Expect style changes to come after design sees this.

https://user-images.githubusercontent.com/17580037/130680418-11ecd496-2167-4a32-8ac0-7d301f41482e.mov


src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -135,8 +135,9 @@ export function AdditionalPropertiesSchemaField<T>(
   } = useListState(
     Array.isArray(formData)
       ? formData
-      : // Coerce individual value to an array
-        [formData],
+      : typeof formData === 'string'
+      ? formData.split(',').map(s => s.trim()) // split a string of comma-separated values, then trim whitespace
+      : [formData], // Coerce individual value to an array
```
nickgros, line 140: Add another branch for converting a schema-defined field to an additional property: if it's a string, split at commas.

The inverse happens in CustomDefaultTemplate.tsx, so converting between schema-defined string and custom annotation array is lossless

package.json
```diff
@@ -151,6 +153,7 @@
     "npm": "^6.9.0",
     "npm-run-all": "^4.1.3",
     "prettier": "^2.1.2",
+    "react-select-event": "^5.3.0",
```
nickgros, line 156: `@testing-library/react`-compatible utility for react-select

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
jay-hodgson: Got it.  I've struggled with stringing these ternary operators together.  I think once is fine, twice is not ideal, and greater than or equal to three times makes it too difficult to read.

src/lib/containers/entity/annotations/CustomSelectWidget.tsx
```diff
@@ -0,0 +1,104 @@
...
+    onChange,
+    onBlur,
+    onFocus,
+    placeholder,
+  } = props
+  const { enumOptions } = options as {
+    enumOptions: EnumOption[]
+  }
+  return (
+    <Creatable
```
jay-hodgson, line 74: It's like you've implemented the adapter pattern to use react-select in rjsf-core.  Nice work!

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.md
```diff
@@ -54,7 +39,16 @@ return (
         schemaId={schemaId}
         liveValidate={liveValidate}
         onSuccess={() => {
-          setShowToast(true)
+          displayToast(
```
jay-hodgson, line 42: This is really convenient.  We all need to start using this!

jay-hodgson: The signature is pretty ugly (since it's probably mirroring the signature of the old DisplayUtils in the SWC project).  I wonder if you can add another displayToast() version that takes in an object (where many parameters are optional) to avoid all of these nulls.  What do you think?

jay-hodgson (approved):

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.md
nickgros: Definitely. The only necessary params are the message and maybe the variant. I'd be ok with changing the signature to something like
```ts
type ToastOptions = {
	variant?: 'info' | 'success' | 'warning' | 'danger' // default 'info'
	title?: string // default undefined
	autoCloseInMs?: number // default 15000
	primaryButtonText?: string // default undefined
	onPrimaryButtonClick?: () => void // default undefined
	secondaryButtonText?: string // default undefined
	secondaryButtonHref?: string // default undefined
}

function displayToast(message: string, opts: SynapseToastOptions);
```

PR #1204, opened 2021-08-31
SWC-5757, SWC-5758

nickgros:
Fixes [SWC-5757](https://sagebionetworks.jira.com/browse/SWC-5757) and [SWC-5758](https://sagebionetworks.jira.com/browse/SWC-5758)

### DIFF
--- a/src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
+++ b/src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
@@ -60,6 +60,7 @@ describe('AdditionalPropertiesSchemaField unit tests', () => {
 
     it('converts strings to floats', () => {
       const data = ['abc', '3.4', '4', 'NaN']
+      expect(4.0).toEqual(4)
       expect(transformDataFromPropertyType(data, PropertyType.FLOAT)).toEqual([
         'NaN',
         3.4,
--- a/src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
+++ b/src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
@@ -19,6 +19,10 @@ describe('AdditionalPropertiesSchemaField unit tests', () => {
       const data = [1.1, 2.2, 3.3, 4.4, 5] // not all have to be float
       expect(guessPropertyType(data)).toBe(PropertyType.FLOAT)
     })
+    it('should return "float" for float properties with NaN', () => {
+      const data = [4.4, 'NaN', 53]
+      expect(guessPropertyType(data)).toBe(PropertyType.FLOAT)
+    })
     it('should return "boolean" for boolean properties', () => {
       const data = [true, false]
       expect(guessPropertyType(data)).toBe(PropertyType.BOOLEAN)
@@ -45,6 +49,25 @@ describe('AdditionalPropertiesSchemaField unit tests', () => {
       ])
     })
 
+    it('converts integers to floats', () => {
+      const data = [1, 2, 3]
+      expect(transformDataFromPropertyType(data, PropertyType.FLOAT)).toEqual([
+        '1.0',
+        '2.0',
+        '3.0',
+      ])
+    })
+
+    it('converts strings to floats', () => {
+      const data = ['abc', '3.4', '4', 'NaN']
+      expect(transformDataFromPropertyType(data, PropertyType.FLOAT)).toEqual([
+        'NaN',
+        3.4,
+        '4.0',
+        'NaN',
+      ])
+    })
+
     it('converts booleans to strings', () => {
       const data = [true, false]
       expect(transformDataFromPropertyType(data, PropertyType.STRING)).toEqual([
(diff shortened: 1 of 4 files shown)


src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -49,10 +52,21 @@ export function transformDataFromPropertyType(
...
+        Number.isNaN(Number(item)) ? undefined : Math.floor(Number(item)),
       )
+
+    case PropertyType.FLOAT:
+      return list.map(item => {
+        const asFloat = parseFloat(item)
+        if (Number.isNaN(asFloat)) {
+          return 'NaN'
+        } else if (Number.isInteger(asFloat)) {
+          return `${asFloat}.0`
```
nickgros, line 65: The backend drops precision, so float values may become integers. We add the '.0' so the backend will re-accept it as a float.

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -209,7 +222,10 @@ export function AdditionalPropertiesSchemaField<T>(
           placeholder={props.placeholder ?? ''}
           options={{}}
           formContext={props.formContext as T}
-          onBlur={props.onBlur}
+          onBlur={(id, value) => {
+            setList(transformDataFromPropertyType(list, propertyType))
+            props.onBlur(id, value)
+          }}
```
nickgros, line 228: Re-cast the value to its type on blur (e.g., if you have `7.5` in an integer field, we will convert it to `7` on blur).

src/lib/containers/entity/annotations/CustomTextWidget.tsx
```diff
@@ -0,0 +1,22 @@
+import { Widget, WidgetProps } from '@sage-bionetworks/rjsf-core'
+import React from 'react'
+
+/**
+ * This TextWidget is almost identical to the rjsf TextWidget, except we handle numbers like strings to prevent issues if
+ * the actual annotation data is not a string.
+ */
```
nickgros, line 7: Without this custom widget, non-number annotations that correspond to number fields will be auto-converted to 'undefined' without the user's knowledge.

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -73,14 +87,13 @@ export function getWidgetFromPropertyType(
   propertyType: PropertyType,
 ): AdditionalPropertyWidget {
   switch (propertyType) {
-    case PropertyType.INTEGER:
-    case PropertyType.FLOAT:
-      return AdditionalPropertyWidget.UpDownWidget
```
nickgros, line 78: Don't need to use this UpDownWidget anymore because we're using the (custom) TextWidget for all numbers and strings

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
nickgros: This is SWC-5758

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -19,13 +19,16 @@ export enum PropertyType {
...
   CheckboxWidget = 'CheckboxWidget',
 }
 
 export function guessPropertyType(list: Array<any>): PropertyType {
-  if (list.every(item => typeof item === 'number')) {
+  if (
+    list.every(
+      item => typeof item === 'number' || item === 'NaN', // "NaN" is technically a float value
+    )
+  ) {
```
nickgros, line 31: SWC-5757 - treat "NaN" as a float

src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
```diff
@@ -45,6 +49,25 @@ describe('AdditionalPropertiesSchemaField unit tests', () => {
...
+        '3.0',
+      ])
+    })
+
+    it('converts strings to floats', () => {
+      const data = ['abc', '3.4', '4', 'NaN']
+      expect(transformDataFromPropertyType(data, PropertyType.FLOAT)).toEqual([
+        'NaN',
+        3.4,
+        '4.0',
```
jay-hodgson, line 66: Why does the '3.4' input get transformed to 3.4, but '4' is a string ('4.0')?   I guess I expected it to be 4.0

src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
jay-hodgson: Like the floor truncation - it's simple and easy to predict the behavior.

jay-hodgson (approved):

src/__tests__/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.test.ts
nickgros: I switched the implementation from appending `".0"` to using `Number.toFixed`, but in either case, the output is a string. We have to use a string because the number `4.0` is automatically converted to `4`

PR #1209, opened 2021-09-01
Release 2.0.35

jay-hodgson:


jay-hodgson:
Includes the following changes...

Summary | Issue key
-- | --
Allow "showDownloadColumn" when isEntityView instead of isFileView | PORTALS-1999
Change "request access" column logic to tolerate entity views | PORTALS-1997
Pop-up user cards email address "copy" button not working | SWC-5760
Floats equal to whole integers should not be coerced to integers | SWC-5758
Editor should consider "NaN" to be a floating point value | SWC-5757
Take Download List V2 out of alpha mode | SWC-5651

PR #1215, opened 2021-09-08
PORTALS-1991 & SWC-5765: Duplicate accessors bug

emmanmills:
This is caused by not being able to add the current user when creating a new AR.

### DIFF
--- a/src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
+++ b/src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
@@ -156,7 +156,11 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
         })
       })
     } else {
-      accessorChanges.forEach(item => {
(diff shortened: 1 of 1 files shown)


src/lib/containers/access_requirement_list/managedACTAccess/RequestDataAccessStep2.tsx
```diff
@@ -166,9 +166,10 @@ const RequestDataAccessStep2: React.FC<RequestDataAccessStep2Props> = props => {
     })
     Promise.all(promises).then(profiles => {
       const profileAndAccessType:Accessor[] = profiles.map((item, i) => {
+        const accessType = accessorChanges && accessorChanges[i]?.type ? accessorChanges[i].type : AccessType.GAIN_ACCESS
         return {
           profile: item,
-          accessType: accessorChanges[i].type
```
emmanmills, line 171: This is the source of the bug. `accessorChanges` is undefined when a user creates a new AR. And the app just throws an error and the user is not added.

jay-hodgson (approved):

PR #1220, opened 2021-09-09
SWC-5761: User profile links (to projects/teams/challenges)

jay-hodgson:


jay-hodgson:
Here's my quick local integration test (to verify that it can be integrated into the SWC project).  The layout is obviously wrong, but it works :)
Screen Shot 2021-09-09 at 11 31 20 AM

src/lib/utils/typography/Typography.styles.tsx
```diff
@@ -0,0 +1,96 @@
+import { makeStyles } from "@material-ui/core";
+
+const basicFontStyle = {
+  fontFamily: "Lato",
+  fontStyle: "normal",
+  letterSpacing: "0px",
+}
+
+const useTypographyStyle = makeStyles(
```
nickgros, line 9: **Not needed for this PR**, but would like to have a demo in StyleGuidist with all of these variants and a link to the Figma. Any other ways we can close the gap between design tokens and the end result?

src/lib/utils/hooks/SynapseAPI/useGetUserProjects.ts
```diff
@@ -0,0 +1,37 @@
...
+  useInfiniteQuery,
+  UseInfiniteQueryOptions,
+} from 'react-query'
+import { SynapseClient } from '../..'
+import { SynapseClientError } from '../../SynapseClient'
+import { useSynapseContext } from '../../SynapseContext'
+import { ProjectHeaderList } from '../../synapseTypes'
+import { GetProjectsParameters } from '../../synapseTypes/GetProjectsParams'
+
+export function useGetUserProjectsInfinite(
```
nickgros, line 11: I think we already have a `useProjects` file with this hook (used in Entity Finder)

nickgros: Nvm, I see that it uses a different service. Could we put all of these hooks in the same file?

src/lib/style/components/_user-profile-links.scss
```diff
@@ -0,0 +1,28 @@
...
+  .Tabs {
+    @include SRCMixins.underline-tabs;
+    .Tab {
+      padding-left: 10px;
+    }
+  }
+
+  .TabContent {
+    min-height: calc(100vh - 100px);
+    max-height: calc(100vh - 100px);;
```
nickgros, line 21: Just curious: what are these doing?

src/lib/style/abstracts/_mixins.scss
```diff
@@ -77,10 +77,15 @@
...
-    border-bottom: 3px solid SRC.$primary-action-color-500;
+    border-bottom: 3px solid $active-color;
+    color: #22252A;
+    svg {
+      color: SRC.$primary-action-color-500;
+    }
   }
   > *[aria-selected='false'] {
-    border-bottom: 3px solid lightgray;
+    color: #878E95;
```
nickgros, line 87: Are these hard-coded colors (#22252A and #878E95) specified by design, and if so, can we assign them to named variables?

src/lib/containers/user_profile_links/UserChallenges.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { PRODUCTION_ENDPOINT_CONFIG } from '../../utils/functions/getEndpoint'
+import { useGetUserChallengesInfinite } from '../../utils/hooks/SynapseAPI/useGetUserChallenges'
+import { ChallengeWithProjectHeader } from '../../utils/synapseTypes/ChallengePagedResults'
+
+export type UserChallengesProps = {
+  userId: string
+}
+
+export default function UserChallenges({userId}:UserChallengesProps) {
+  const handleError = useErrorHandler()
```
nickgros, line 15: Do these components have an ancestor ErrorBoundary component? I didn't see one but I'm not sure.

src/lib/containers/user_profile_links/UserChallenges.tsx
```diff
@@ -0,0 +1,89 @@
...
+      )
+    : []
+
+  return (
+    <>
+      {allRows.length > 0 && (
+        <>
+          {allRows.map((item:ChallengeWithProjectHeader) => {
+            if (item && item.challenge && item.projectHeader) {
+              // another option may be to use an EntityLink
```
nickgros, line 61: I think you only meant to leave this comment for the Project component :)

src/lib/utils/typography/Typography.tsx
```diff
@@ -0,0 +1,42 @@
...
+  | 'smallText1'
+  | 'smallText2'
+  | 'smallLink'
+  | 'label'
+  | 'buttonLink'
+  | 'hintText'
+  component?: ElementType
+}
+
+const Typography: FC<ITypography> = (props) => {
```
nickgros, line 24: Looking forward to using this!

src/lib/containers/user_profile_links/UserProfileLinks.tsx
```diff
@@ -0,0 +1,78 @@
...
+        return 'dashboard'
+      case UserProfileLinksTabs.TEAMS:
+        return 'peopleOutlined'
+      case UserProfileLinksTabs.CHALLENGES:
+        return 'challengesOutlined'
+    }
+  }
+  return (
+    <div className="UserProfileLinks">
+      <Typography variant="headline2" className="title">{userProfile?.userName}&apos;s Items</Typography>
```
nickgros, line 39: Just a really, _really_ picky, minor note: this could look weird if this renders before the user profile is fetched, if it's not cached in react-query. Maybe would be worth using a skeleton here.

Since this will be embedded in SWC, I'm not sure if a cache miss is likely; it might not make a difference--that probably depends on how we're populating the user cards

src/lib/containers/user_profile_links/UserChallenges.tsx
```diff
@@ -0,0 +1,89 @@
...
+              return (
+                <p key={`user-challenge-list-item-${item.challenge.projectId}`}>
+                  <a
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href={`${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!Synapse:${item.challenge.projectId}/challenge`}
```
nickgros, line 67: Is opening in new window preferred? No opinion from me, just seems different to how we normally treat internal links.

nickgros (approved):

src/lib/containers/user_profile_links/UserChallenges.tsx
jay-hodgson: Good call, will add a SynapseErrorBoundary to the UserProfileLinks parent component

src/lib/containers/user_profile_links/UserChallenges.tsx
jay-hodgson: Good call, I think opening in the same window makes sense for these links

src/lib/style/abstracts/_mixins.scss
jay-hodgson: Taken from the design, will pull out into a named variable.

src/lib/style/components/_user-profile-links.scss
jay-hodgson: It's to render a potentially long list of projects/teams/challenges in a scroll panel, constrained to the height of the viewport (so you could scroll outside of this area and still access the footer, for example).

src/lib/utils/typography/Typography.styles.tsx
jay-hodgson: That's a good idea.

PR #1224, opened 2021-09-10
SWC-5728

nickgros:
Fixes an issue where custom annotations were coerced to strings before being converted back to their original type


src/lib/containers/entity/annotations/AdditionalPropertiesSchemaField.tsx
```diff
@@ -163,6 +151,14 @@ export function AdditionalPropertiesSchemaField<T>(
     setList,
   } = useListState(convertToArray(formData))
 
+  // The type determines which widget we show.
+  const [propertyType, setPropertyType] = useState<PropertyType>(
+    guessPropertyType(list),
```
nickgros, line 156: Use the list to initialize, rather than initializing to string

jay-hodgson (approved):

PR #1226, opened 2021-09-10
PORTALS-2012: avoid app crash (no error boundary to handle errors)

jay-hodgson:


jay-hodgson:
The end result will not be pretty, but it's a lot better than a blank page or app crash.  And because this is a pretty unusual case (access to a page that is unlinked unless logged in), this is what I recommend for now.
Screen Shot 2021-09-10 at 10 46 20 AM

nickgros (approved):

PR #1230, opened 2021-09-14
SWC-5776

nickgros:
Directly addresses [SWC-5776](https://sagebionetworks.jira.com/browse/SWC-5776), which is a subtask of [SWC-5769](https://sagebionetworks.jira.com/browse/SWC-5769)
- Type updates
  - Add Versionable, VersionableEntity, update Entity and FileEntity to utilize hierarchy
  - Add field isLatestVersion to VersionInfo
- [SWC-5776](https://sagebionetworks.jira.com/browse/SWC-5776): Add `mustSelectVersionNumber` prop to Entity Finder, which forces the user to select a version number for versionable entities
- Fix user card text color when selected in entity finder
- EntityBadgeIcons: Use IntersectionObserver to check if the component is in view before fetching any data.

Below is a video demonstrating `mustSelectVersionNumber`. Note how 'Always Latest Version' is not an option.

https://user-images.githubusercontent.com/17580037/133303437-c9e5c1d4-fa20-4691-bf3f-8fb4d53cf8dc.mov

### DIFF
--- a/src/lib/containers/EntityBadgeIcons.tsx
+++ b/src/lib/containers/EntityBadgeIcons.tsx
@@ -27,6 +27,7 @@ import {
 import { useEffect } from 'react'
 import { EntityModal, EntityModalTabs } from './entity/metadata/EntityModal'
 import useGetEntityBundle from '../utils/hooks/SynapseAPI/useEntityBundle'
+import { useInView } from 'react-intersection-observer'
 
 const isPublic = (bundle: EntityBundle): boolean => {
   return bundle.benefactorAcl!.resourceAccess.some(ra => {
@@ -100,7 +101,11 @@ export const EntityBadgeIcons: React.FunctionComponent<EntityBadgeIconsProps> = 
     ANNOTATIONS_MISSING = 'Missing',
   }
 
-  const { data: bundle } = useGetEntityBundle(entityId)
+  const { ref, inView } = useInView()
+  const { data: bundle } = useGetEntityBundle(entityId, undefined, undefined, {
+    enabled: inView,
+    staleTime: 60 * 1000, // 60 seconds
+  })
   const [showModal, setShowModal] = useState(false)
   const [schemaConformance, setSchemaConformance] = useState(
     SchemaConformanceState.NO_SCHEMA,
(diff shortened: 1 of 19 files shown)


src/lib/containers/EntityBadgeIcons.tsx
```diff
@@ -189,121 +196,121 @@ export const EntityBadgeIcons: React.FunctionComponent<EntityBadgeIconsProps> =
...
-            className={`EntityBadge__Badge ${schemaConformance}`}
-            style={canOpenModal ? { cursor: 'pointer' } : undefined}
-            onClick={canOpenModal ? () => setShowModal(true) : undefined}
-            icon={faTag}
-            data-for={ENTITY_BADGE_ICONS_TOOLTIP_ID}
-            data-tip={sanitizeHtml(annotationsHtml)}
-            data-html={true}
-            data-testid={'annotations-icon'}
+    <div ref={ref} className="EntityBadge" style={{ flexWrap, justifyContent }}>
+      {bundle && (
```
nickgros, line 201: Moved this check inside the `div` so we can attach the ref to check to see if the component comes into view.

src/lib/containers/entity_finder/details/view/DetailsViewRow.tsx
```diff
@@ -79,24 +77,35 @@ export const DetailsViewRow: React.FunctionComponent<DetailsViewRowProps> = ({
     },
   )
 
+  // TODO: Handle pagination, preferably with an IntersectionObserver
+  // We can't use IntersectionObserver via ref on a div within a <select> input
+  // It might make more sense to use react-select here
+  const { data: versionData } = useGetVersionsInfinite(entityHeader.id, {
```
nickgros, line 83: We now have to fetch versions for all versionable entities that come into view. Before we were deferring until the entity was selected. Hopefully shouldn't hurt performance too badly.

src/lib/containers/entity_finder/details/view/DetailsViewRow.tsx
```diff
@@ -79,24 +77,35 @@ export const DetailsViewRow: React.FunctionComponent<DetailsViewRowProps> = ({
     },
   )
 
+  // TODO: Handle pagination, preferably with an IntersectionObserver
+  // We can't use IntersectionObserver via ref on a div within a <select> input
+  // It might make more sense to use react-select here
```
nickgros, line 82: For `select` inputs, useInView won't work for populating options. This would be a good place to use react-select, the custom select component used in the new annotation editor, but that's a bigger change than I want to make right now.

jay-hodgson (approved):

PR #1231, opened 2021-09-14
SWC-5771: do not explicitly set the user links panel width 

jay-hodgson:
will grow based on container


jay-hodgson:
Delegating size to wrapper (where all of the responsive grid definitions live!)

emmanmills (approved):

PR #1233, opened 2021-09-16
PORTALS-2002

nickgros:
- Adds `ColumnSpecifiedLink` type, which allows you to format a SynapseTableCell with a link defined by another column. 
  - Works for both cards and tables
- Convert function `renderLabel` to component `SynapseCardLabel`
- Convert function `renderTableCell` to component `SynapseTableCell`
- Typing cleanup in various card components

Screenshot working in the AD Portal. The first two rows have an empty `link` field

image

### DIFF
--- a/src/__tests__/lib/containers/SynapseTable.test.tsx
+++ b/src/__tests__/lib/containers/SynapseTable.test.tsx
@@ -20,7 +20,7 @@ import { SynapseConstants } from '../../../lib'
 import { QueryWrapperChildProps } from '../../../lib/containers/QueryWrapper'
 import { getColumnIndiciesWithType } from '../../../lib/containers/synapse_table_functions/getColumnIndiciesWithType'
 import { getUniqueEntities } from '../../../lib/containers/synapse_table_functions/getUniqueEntities'
-import { renderTableCell } from '../../../lib/containers/synapse_table_functions/renderTableCell'
+import { SynapseTableCell } from '../../../lib/containers/synapse_table_functions/SynapseTableCell'
 
 import SynapseTable, {
   SORT_STATE,
@@ -639,7 +639,7 @@ describe('basic functionality', () => {
       it('renders an entity link', () => {
         const tableCell = shallow(
           <div>
-            {renderTableCell({
+            {SynapseTableCell({
               ...tableCellProps,
               colIndex: ENTITYID_INDEX,
               columnValue: mockEntityLinkValue,
@@ -652,7 +652,7 @@ describe('basic functionality', () => {
       it('renders a link for all authenticated users', () => {
         const tableCell = shallow(
           <div>
-            {renderTableCell({
+            {SynapseTableCell({
               ...tableCellProps,
               colIndex: USERID_INDEX,
               columnValue: mockAllAuthenticatedUsersValue,
@@ -670,7 +670,7 @@ describe('basic functionality', () => {
       it('renders a link for a team', () => {
         const tableCell = shallow(
           <div>
-            {renderTableCell({
+            {SynapseTableCell({
               ...tableCellProps,
               mapUserIdToHeader,
               colIndex: USERID_INDEX,
@@ -686,7 +686,7 @@ describe('basic functionality', () => {
       it('renders a user card link', () => {
         const tableCell = shallow(
           <div>
-            {renderTableCell({
+            {SynapseTableCell({
               ...tableCellProps,
               mapUserIdToHeader,
               colIndex: USERID_INDEX,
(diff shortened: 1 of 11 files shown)


src/lib/containers/GenericCard.tsx
```diff
@@ -207,39 +223,86 @@ export const renderLabel = (args: {
...
+      console.warn(
+        `Could not determine column index of ${labelLink.linkColumnName}`,
+      )
+      return <>{value}</>
+    } else if (row == null) {
+      // Sorry future dev, I couldn't figure out how to pass the row down from GenericCard because it's not fully typed
+      console.warn(
+        `Couldn't create link because row had value ${row}. Specifying a link via column is not currently possible in cards`,
+      )
+      return <>{value}</>
```
nickgros, line 264: Handling common failure cases when trying to call this from a card instead of table cell

src/lib/containers/GenericCard.tsx
```diff
@@ -207,39 +223,86 @@ export const renderLabel = (args: {
...
     } else {
       return <MarkdownSynapse renderInline={true} markdown={value} />
     }
   }
   const split = strList ? strList : str.split(',')
-  return split.map((el, index) => {
-    const { baseURL, URLColumnName, wrapValueWithParens } = labelLink
-    const value = wrapValueWithParens ? `(${el})` : el
-    const href = `/${baseURL}?${URLColumnName}=${value}`
+  if ('linkColumnName' in labelLink) {
```
nickgros, line 250: This is where we handle the new type of link configuration

src/__tests__/lib/containers/SynapseTable.test.tsx
```diff
@@ -685,65 +678,57 @@ describe('basic functionality', () => {
...
+            columnName={'a'}
+            selectColumns={[
+              { columnType: EntityColumnType.STRING, name: 'a', id: '' },
+            ]}
+          />,
         )
         expect(tableCell.find(MarkdownSynapse).props().markdown).toEqual(
           mockMarkdownColumnValue,
         )
       })
```
nickgros, line 711: Since we were calling a function, rather than rendering a component, we weren't actually doing shallow rendering in any of these tests! This is the only one that required changing to `mount`, since it inspects the MarkdownSynapse component

src/lib/containers/CardContainerLogic.tsx
```diff
@@ -1,32 +1,42 @@
...
- * 
+ *
  * This can be fixed by rewriting tests to not rely on the shallow renderer.
- * 
+ *
  * See here: https://github.com/enzymejs/enzyme/issues/1553
  */
 
-// TODO: this import nearly doubles the package size of SRC as a UMD build by ~400KB
-// will have to find a way to use individual lodash packages instead of the entire thing
```
nickgros, line 26: I don't think this is still true--we're importing individual exports from lodash-es and rollup is doing its job w/ tree shaking

nickgros: Also...it would be great if 400KB was still "nearly doubling" 😂

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -7,19 +7,48 @@ import UserCard from '../UserCard'
...
+  columnName: string
+  selectColumns: SelectColumn[] | undefined
+  columnModels: ColumnModel[] | undefined
+  tableEntityId?: string
+  row: Row
+}
+
 // Render table cell, supports Entity's and User Icons
-export const renderTableCell = ({
+export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
```
nickgros, line 51: This is a component, so treat it like one

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -100,23 +99,21 @@ const QueryWrapperPlotNav: React.FunctionComponent<QueryWrapperPlotNavProps> = p
       <SynapseContextConsumer>
         {context => (
           <QueryWrapper
-            {...rest}
+            {...props}
```
nickgros, line 102: Props that were destructured above weren't propagating (e.g. facetsToPlot, facetsToFilter). This bug has existed since this file was created, but I think the particular props and children that were affected have varied over time

src/lib/containers/GenericCard.tsx
```diff
@@ -207,39 +225,85 @@ export const renderLabel = (args: {
...
+      if (isEmpty(href)) {
+        return <>{value}</>
+      }
+
+      return (
+        <>
+          {split.map((el, index) => {
+            return (
+              <React.Fragment key={el}>
+                <a href={href} key={el} className={newClassName} style={style}>
```
jay-hodgson, line 273: target?  noreferrer noopener?

jay-hodgson (approved):

PR #1234, opened 2021-09-17
SWC-5786: bring back the spinner after the initial data load

jay-hodgson:


src/demo/containers/playground/WidgetDemo.tsx
```diff
@@ -100,13 +100,11 @@ export const WidgetDemo: React.FunctionComponent = () => {
       <br />
       <Checkbox
         label="Initially Checked Checkbox"
-        id="ch1"
```
jay-hodgson, line 103: Attempt to fix GH action responsible for NPM publish, GH tag, and styleguidist output.  https://github.com/Sage-Bionetworks/Synapse-React-Client/runs/3604908922?check_suite_focus=true

nickgros: This alone won't fix the automated deploy (there's other configuration we are missing), but I can take care of the rest.

src/lib/containers/user_profile_links/UserProjects.tsx
```diff
@@ -72,7 +74,10 @@ export default function UserProjects({ userId }: UserProjectsProps) {
         </>
       )}
       {!isFetching && allRows.length == 0 && <div>Empty</div>}
-      {isFetching && <SkeletonTable numRows={5} numCols={1} />}
+      {isLoading && <SkeletonTable numRows={5} numCols={1} />}
+      {isFetching && !isLoading && <div className="placeholder">
+        <SynapseSpinner size={30} />
+      </div>}
```
nickgros, line 80: Just my opinion: this spinner will pop up each time it re-fetches, and with our current config, React Query is very eager to refetch. I think for now we should omit this entirely (or we can consider changing our React Query config).

If we _need_ to tell the user that data is re/fetching, then I think we would want a more subtle indicator. We would probably want to talk to design about that.

jay-hodgson: Works for me.  Spinners have been removed.

nickgros (approved):

PR #1235, opened 2021-09-17
Fix update documentation action

nickgros:
Use an off the shelf action to get the deployment to work.


.github/workflows/npmpublish.yml
```diff
@@ -32,5 +32,10 @@ jobs:
               ref: "refs/tags/v${{ env.PACKAGE_VERSION }}",
               sha: context.sha
             })
-      - name: Update GitHub Pages
-        run: yarn deploy
+      - name: Build docs
+        run: yarn build:docs
+      - name: Deploy docs to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.5
```
nickgros, line 38: It's surprisingly a pain to get this to work in GH actions, so we're using a popular action that someone else has created

.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,41 @@
+name: "NPM Publish and GitHub tag"
+on:
+  push:
+    branches:
+      - main
```
nickgros, line 5: I just realized that this only runs when we push to the main branch. Is there any harm in pulling this this into `develop`? That would make sure that both the action change and code changes are pushed to `main` at the same time.

jay-hodgson (approved):

jay-hodgson (changes requested):
the commit history makes it look like you merged main into develop, was that intentional?

jay-hodgson (approved):

PR #1237, opened 2021-09-20
PORTALS-2014

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/PORTALS-2014)
Not sure why event propagation was preventing this dropdown from opening (portals only, not in SRC demo)


jay-hodgson:
I wonder if this has something to do with the app-wide click event listener that we set up in AppInitializer?  Just a guess, but that's one difference between the two environments.

jay-hodgson (approved):

PR #1244, opened 2021-09-28
SWC-5794: Access Requirement list step was being reset

jay-hodgson:
Not sure why this code was here, but this feature is outside of experimental mode (we should not reset the step).


jay-hodgson:
@emmanmills : Is it ok to remove [this](https://github.com/Sage-Bionetworks/Synapse-React-Client/commit/6557493227bcb045ded7463cc19823b2f23b104e) now that the Managed ACT Access Requirement support is out of alpha?

emmanmills (approved):

PR #1257, opened 2021-10-08
PORTALS-2010, PORTALS-2026

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -852,26 +852,34 @@ export default class SynapseTable extends React.Component<
           </td>,
         )
       }
-
+      const isFileEntity:boolean = mapEntityIdToHeader[rowSynapseId]?.type == 'org.sagebionetworks.repo.model.FileEntity'
```
jay-hodgson, line 855: On a row-by-row basis, check to see if it's a File entity.

PR #1261, opened 2021-10-13
revert change made for PORTALS-2026

jay-hodgson:


src/lib/containers/DirectDownload.tsx
```diff
@@ -63,8 +63,9 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = props =
 
     if (!preSignedURL) {
       console.log('Fail to get file download link')
+    } else {
+      window.open(preSignedURL)
```
jay-hodgson, line 67: But only open the presigned URL in a new window if we were able to get a presigned URL

src/lib/containers/DirectDownload.tsx
```diff
@@ -140,9 +141,11 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = props =
   const getIcon = () => {
     if (isExternalFile) {
       return (
-        <a rel="noreferrer" href={externalURL} target="_blank">
-          <Icon type="externallink" />
-        </a>
+        <button className={'btn-download-icon'}>
```
jay-hodgson, line 144: https://sagebionetworks.jira.com/browse/PORTALS-1880
Output in the same button container so these icons are always aligned.

PR #1264, opened 2021-10-15
PORTALS-2009: Hide 'download' column when not logged in

jay-hodgson:
And if showing an Entity View, make sure row IDs are added to the entity ID set ("id" column may not have been selected)

### DIFF
--- a/src/lib/containers/table/SynapseTable.tsx
+++ b/src/lib/containers/table/SynapseTable.tsx
@@ -252,6 +252,16 @@ export default class SynapseTable extends React.Component<
       mapEntityIdToHeader,
       entityIdColumnIndicies,
     )
+    // also include row entity ids if this is a view (it's possible that the ID column was not selected)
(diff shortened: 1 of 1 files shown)


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -252,6 +252,16 @@ export default class SynapseTable extends React.Component<
...
       entityIdColumnIndicies,
     )
+    // also include row entity ids if this is a view (it's possible that the ID column was not selected)
+    if (this.state.isEntityView) {
+      const { queryResult } = data
+      const { queryResults } = queryResult
+      const { rows } = queryResults
+      rows.forEach((row) => {
+        const rowSynapseId = `syn${row.rowId}`
+        distinctEntityIds.add(rowSynapseId)
```
jay-hodgson, line 262: Add all row ID entity IDs to the distinctEntityIds Set

PR #1274, opened 2021-11-01
PORTALS-2032 - Use FTS for tables if it's enabled

nickgros:
Demo:

https://user-images.githubusercontent.com/17580037/139728524-e87054cc-adec-4a9c-a7d1-a1290740ef2b.mov

SearchV2 had no tests, could take some effort to add some for this b/c of trying to bootstrap a whole QueryWrapper setup


package.json
```diff
@@ -30,7 +30,7 @@
     "@sage-bionetworks/rjsf-core": "^3.1.4",
     "@upsetjs/react": "^1.6.2",
     "animate.css": "^4.1.1",
-    "bootstrap": "^4.6.0",
+    "bootstrap": "^4.6.1",
```
nickgros, line 33: This will resolve [SWC-5752](https://sagebionetworks.jira.com/browse/SWC-5752)

package.json
```diff
@@ -71,7 +71,7 @@
     "react-plotly.js": "^2.5.1",
     "react-plotlyjs-ts": "^2.2.1",
     "react-popper": "^2.2.5",
-    "react-query": "^3.25.1",
+    "react-query": "3.31.0",
```
nickgros, line 74: The types in react-query are unstable in minor version releases so we need to pin a particular version. Ran into typescript issues in Agora because I was pulling a newer version of react-query than I expected

src/lib/containers/FullTextSearch.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const search = (event: React.SyntheticEvent<HTMLFormElement>) => {
+    event.preventDefault()
+
+    // The HTML validation doesn't trim the string, so we add an extra check.
+    // We don't auto-trim the form text ourselves because the user may still be focused on the input.
+    if (searchText.trim().length < MIN_SEARCH_QUERY_LENGTH) {
+      searchInputRef.current?.setCustomValidity(
+        `Search term must have a minimum of ${MIN_SEARCH_QUERY_LENGTH} characters`,
+      )
```
nickgros, line 40: Is requiring a min length of a trimmed string good enough? I really don't want to worry about tokenizing the input.

src/lib/containers/SearchV2.tsx
```diff
@@ -1,6 +1,3 @@
-import * as React from 'react'
```
nickgros, line 1: There is only one non-whitespace/import change in this file, in the render method

src/lib/containers/SearchV2.tsx
```diff
@@ -221,56 +246,56 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
...
-              className="SearchV2__searchbar"
-              onSubmit={this.search}
+      <div
+        className={`SearchV2 ${
+          showFacetFilter
+            ? QUERY_FILTERS_EXPANDED_CSS
+            : QUERY_FILTERS_COLLAPSED_CSS
+        }`}
+      >
+        <Collapse in={topLevelControlsState?.showSearchBar}>
```
nickgros, line 256: Using MUI's Collapse instead of a custom animation

src/lib/containers/SearchV2.tsx
```diff
@@ -221,56 +246,56 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
...
+                <FontAwesomeIcon
+                  className="SRC-primary-text-color"
+                  icon="times"
+                />
+              </button>
+            )}
+          </form>
+        </Collapse>
         <div className="SearchV2__dropdown_pos">
           <CSSTransition
```
nickgros, line 300: Not using MUI Collapse here because the dropdown is absolutely-positioned, so Collapse doesn't work

src/lib/utils/functions/TypeUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+  ObjectType = unknown,
+>(expectedConcreteType: string) {
+  return (object: ObjectType): object is TTypeChecked =>
+    !!(
+      object &&
+      typeof object === 'object' &&
+      'concreteType' in object &&
+      object['concreteType'] === expectedConcreteType
+    )
+}
```
nickgros, line 25: Thoughts on this? Trying to do runtime type checking is hard and/or verbose, this might get us one step closer.

Another option is to do a huge refactor to define/use JavaScript classes, and just make sure that everywhere we create or fetch an object, we instantiate a new object.

src/lib/utils/synapseTypes/Table/QueryFilter.ts
```diff
@@ -1,29 +1,67 @@
...
-  concreteType: string
-  columnName: string // name of the column to filter
-  operator?: ColumnSingleValueFilterOperator
-  function?: ColumnMultiValueFunction
-  values: string[]
+  concreteType:
+    | TEXT_MATCHES_QUERY_FILTER_CONCRETE_TYPE
+    | COLUMN_SINGLE_VALUE_QUERY_FILTER_CONCRETE_TYPE
+    | COLUMN_MULTI_VALUE_FUNCTION_QUERY_FILTER_CONCRETE_TYPE
 }
```
nickgros, line 52: The majority of the changes in this PR are because it's no longer true that a QueryFilter has a `values` field. Thankfully, we have static typing!

.eslintrc.json
```diff
@@ -42,7 +42,9 @@
     "@typescript-eslint/no-unnecessary-type-assertion": "warn",
     "@typescript-eslint/no-non-null-assertion": "off",
     "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
-    "@typescript-eslint/explicit-module-boundary-types": "off"
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/no-empty-function": "off",
+    "@typescript-eslint/no-extra-semi": "warn"
```
nickgros, line 47: - We use empty no-op functions all over the place, the rule isn't helpful
- Prettier may enforce adding extra semicolons where TSLint doesn't like them, so changing that to a warning

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -188,47 +199,89 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
...
-          columnName: el.columnName,
-          values:
-            el.columnName === columnName
-              ? el.values.filter(el => el !== value)
-              : el.values,
-          operator: el.operator,
-          function: el.function,
-          concreteType: el.concreteType
+    cloneLastQueryRequest.query.additionalFilters =
+      cloneLastQueryRequest.query.additionalFilters?.reduce(function (
```
nickgros, line 231: Big style question--should we use `reduce()` or `map().filter()`?

Would need to do something like

```ts
cloneLastQueryRequest.query.additionalFilters?.map(qf => {
	if (isTextMatchesQueryFilter(qf) {
		return qf
	} else if (isColumnSingleValueQueryFilter(el) || isColumnMultiValueFunctionQueryFilter(el)) {
		// transform by removing matching column/value tuple
		// ...
		return transformedQueryFilter
	}
}).filter(queryFilter => {
	if (isTextMatchesQueryFilter(qf)) {
		return true
	} else if (isColumnSingleValueQueryFilter(el) || isColumnMultiValueFunctionQueryFilter(el)) {
		return values.length > 0
	}
})
```

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -188,47 +199,89 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
...
         [RadioValuesEnum.ANY, RadioValuesEnum.ANY],
       )
     }
   }
 
-  const removeSearchQuerySelection = (columnName: string, value: string) => {
+  const removeColumnSearchQuerySelection = (
+    columnName: string,
+    value: string,
+  ) => {
```
nickgros, line 225: Now. QueryFilters may not apply to specific columns, so this function must become more specific.

nickgros: The `(column, value)` signature remains because we still split an array of QueryFilter values into individual pills in the UI.

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -188,47 +199,89 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
...
+        return filtered
+      },
+      [])
+
+    executeQueryRequest!(cloneLastQueryRequest)
+  }
+
+  const removeTextMatchesQueryFilter = (
+    queryFilter: TextMatchesQueryFilter,
+  ) => {
```
nickgros, line 272: Since the other method only handles Column-specific QueryFilters, we need a new method to handle the new TextMatchesQueryFilter

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -238,22 +291,47 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
...
-      const { columnName } = el
-      return el.values.map(value => {
-        return (
-          <SelectionCriteriaPill
-            key={value}
-            filter={{
+  const searchSelectionCriteriaPillProps:
+    | (SelectionCriteriaPillProps & {
+        key: string
+      })[]
```
nickgros, line 297: Instead of rendering an array of elements, I decided to compute an array of props and then render the elements inline (we render facet filters inline). I might revert this, I don't think I actually like this.

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -238,22 +291,47 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
...
+    | undefined = lastQueryRequest?.query.additionalFilters?.reduce(
```
nickgros, line 298: Same concern as the above `reduce`

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -54,79 +55,85 @@ export type QueryWrapperPlotNavProps = SearchParams &
...
+            <QueryWrapper
+              {...props}
+              token={context?.accessToken}
+              initQueryRequest={initQueryRequest}
+            >
+              {entity && isTableEntity(entity) && entity.isSearchEnabled ? (
+                <FullTextSearch />
+              ) : (
+                <SearchV2 {...searchConfiguration} />
+              )}
```
nickgros, line 110: This is the main change.

We could just add FTS as an option in SearchV2 so users could query both across all rows and in particular columns. Was this option considered in the design phase?

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -54,79 +55,85 @@ export type QueryWrapperPlotNavProps = SearchParams &
...
+      downloadCartPageUrl,
+    } = props
+    const sqlUsed = insertConditionsFromSearchParams(
+      sql,
+      searchParams,
+      sqlOperator,
+    )
+
+    const entityId = parseEntityIdFromSqlStatement(sqlUsed)
+    const { data: entity } = useGetEntity(entityId)
```
nickgros, line 82: Surprisingly, we don't have the entity here. I need it to determine if full text search is enabled.

src/lib/containers/widgets/facet-nav/FacetNav.tsx
```diff
@@ -163,9 +165,11 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
...
   }
 
-  const hasSelectedFacets =
-    lastQueryRequest?.query.selectedFacets !== undefined &&
-    lastQueryRequest.query.selectedFacets.length > 0
+  const hasFacetsOrFilters =
+    (lastQueryRequest?.query.selectedFacets !== undefined &&
+      lastQueryRequest.query.selectedFacets.length > 0) ||
+    (lastQueryRequest?.query.additionalFilters !== undefined &&
+      lastQueryRequest?.query.additionalFilters.length > 0)
```
nickgros, line 172: We weren't considering filters when rendering this text or the "Clear All" button.

src/lib/containers/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -1,83 +1,96 @@
...
+  | { readonly facetWithSelection: FacetWithSelection }
+  | {
+      filter: {
+        columnName?: string
+        value: string
+      }
+    }
+) & {
+  onRemove: () => void
 }
```
nickgros, line 29: Using a funky type to be a little more explicit about what we're expecting: `onRemove` with either `facetWithSelection` or `filter`, but not both.

src/lib/containers/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -1,83 +1,96 @@
...
+      const { filter } = props
+      let filterValue = filter.value
+      if (filterValue?.startsWith('%') && filterValue?.endsWith('%')) {
+        // strip '%' wildcard character when using a LIKE condition
+        filterValue = filterValue.substring(1, filterValue.length - 1)
+      }
+
+      // For full-text search, set the pill text and tooltip text
+      innerText = filterValue
+      tooltipText = `Text matches: "${filterValue}"`
```
nickgros, line 60: This is how we handle FTS, other pills are unchanged.

src/lib/containers/FullTextSearch.tsx
jay-hodgson: Yep, let's start with that!

src/lib/containers/TotalQueryResults.tsx
jay-hodgson: I think reduce is easier to understand in this case, but I don't have a strong opinion on this.
@conundrumgirl ?

jay-hodgson (approved):

src/lib/containers/TotalQueryResults.tsx
conundrumgirl: In both map and filter we have the same conditions check. So if we could get rid of this redundancy by using reduce I think it would be preferable

PR #1275, opened 2021-11-02
SWC-5838

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5838)

### DIFF
--- a/src/lib/containers/LoadingScreen.tsx
+++ b/src/lib/containers/LoadingScreen.tsx
@@ -38,11 +38,11 @@ type BlockingLoaderProps = {
 }
 export const BlockingLoader: React.FC<BlockingLoaderProps> = ({ show }) => {
   useEffect(() => {
-    document.body.style.cursor = 'wait'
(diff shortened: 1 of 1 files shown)


src/lib/containers/LoadingScreen.tsx
```diff
@@ -38,11 +38,11 @@ type BlockingLoaderProps = {
 }
 export const BlockingLoader: React.FC<BlockingLoaderProps> = ({ show }) => {
   useEffect(() => {
-    document.body.style.cursor = 'wait'
+    document.body.style.cursor = show ? 'wait' : 'default'
     return () => {
       document.body.style.cursor = 'default'
     }
-  }, [])
+  }, [show])
```
nickgros, line 45: We were changing the spinner anytime this component was rendered, even if the loader isn't actively being shown. Very confusing, since it looks like Synapse is loading when nothing is actually happening.

jay-hodgson (approved):

PR #1276, opened 2021-11-03
SWC-5839

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5839)

Fixes a bug where the Entity Finder's select all checkbox would be checked if at least one item was selected and the user fetched a list consisting entirely of unselectable entities (e.g. folders when you can only select files)


src/__tests__/lib/utils/hooks/useGetIsAllSelectedInfiniteList.test.ts
```diff
@@ -0,0 +1,297 @@
...
+      getIsAllSelectedFromInfiniteList(
+        fetchedEntities,
+        selectedEntities,
+        [EntityType.PROJECT], // can only select projects
+        false,
+        mockFetchNextPage,
+        false,
+      ),
+    ).toBe(false)
+  })
```
nickgros, line 145: This test captures the bug in the ticket

jay-hodgson (approved):
Nice set of unit tests

PR #1277, opened 2021-11-03
SWC-5845 and SWC-5840

nickgros:
Fixes [SWC-5840](https://sagebionetworks.jira.com/browse/SWC-5840) and [SWC-5845](https://sagebionetworks.jira.com/browse/SWC-5845)

SWC-5840 is a bug where the entity finder doesn't utilize the full width if the viewport is wide enough. This is fixed by always autosizing with a min width, rather than disabling autosize in certain situations.

SWC-5845 is a bug where hidden items take up space in the Entity Finder's DetailsView table. The simple fix is to exclude hidden items from the table data. Before, were hiding them with CSS, and the row virtualization technique would still create a space for the hidden row.

I discovered SWC-5845 while testing search, so also added an optimization where we will only search for entity types that can be selected. This should make searching more efficient in all cases where not all entity types can be selected.

### DIFF
--- a/src/__tests__/lib/containers/entity_finder/EntityFinder.test.tsx
+++ b/src/__tests__/lib/containers/entity_finder/EntityFinder.test.tsx
@@ -392,7 +392,7 @@ describe('EntityFinder tests', () => {
   })
 
   it('handles searching for terms', async () => {
-    renderComponent()
+    renderComponent({ selectableTypes: [EntityType.FILE] })
 
     const query = 'my search terms '
     const queryTerms = ['my', 'search', 'terms']
@@ -408,6 +408,35 @@ describe('EntityFinder tests', () => {
             type: EntityDetailsListDataConfigurationType.ENTITY_SEARCH,
             query: {
               queryTerm: queryTerms,
+              // Verify that at least one of the omitted types is excluded from the search
+              booleanQuery: expect.arrayContaining([
+                {
+                  key: 'node_type',
+                  value: 'project',
+                  not: true,
+                },
+              ]),
+            },
+          },
+        }),
+        {},
+      ),
+    )
+    await waitFor(() =>
+      expect(mockDetailsList).toBeCalledWith(
+        expect.objectContaining({
+          configuration: {
+            type: EntityDetailsListDataConfigurationType.ENTITY_SEARCH,
+            query: {
+              queryTerm: queryTerms,
+              // The lone selectable type should not have been excluded
+              booleanQuery: expect.not.arrayContaining([
+                {
+                  key: 'node_type',
+                  value: 'file',
+                  not: true,
+                },
+              ]),
             },
           },
         }),
--- a/src/lib/containers/entity_finder/EntityFinder.tsx
+++ b/src/lib/containers/entity_finder/EntityFinder.tsx
@@ -21,6 +21,7 @@ import { SYNAPSE_ENTITY_ID_REGEX } from '../../utils/functions/RegularExpression
 import { useSynapseContext } from '../../utils/SynapseContext'
 import { EntityHeader, Reference } from '../../utils/synapseTypes'
 import { EntityType } from '../../utils/synapseTypes/EntityType'
+import { KeyValue } from '../../utils/synapseTypes/Search'
 import { SynapseErrorBoundary } from '../ErrorBanner'
 import { BreadcrumbItem, Breadcrumbs, BreadcrumbsProps } from './Breadcrumbs'
 import {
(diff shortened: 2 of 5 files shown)


src/lib/containers/entity_finder/details/view/DetailsView.tsx
```diff
@@ -295,8 +301,8 @@ export const DetailsView: React.FunctionComponent<DetailsViewProps> = ({
             classPrefix="DetailsViewTable"
             data={tableData}
             height={height}
-            width={autoSizeWidth ? width : 1200}
-            rowHeight={46}
+            width={width > MIN_TABLE_WIDTH ? width : MIN_TABLE_WIDTH}
```
nickgros, line 304: This fixes SWC-5840

src/lib/containers/entity_finder/details/view/DetailsView.tsx
```diff
@@ -59,7 +62,6 @@ export type DetailsViewProps = EntityDetailsListSharedProps & {
   noResultsPlaceholder?: React.ReactElement
   /** We defer to the configuration component to determine this */
   selectAllIsChecked?: boolean
-  autoSizeWidth: boolean
```
nickgros, line 62: Dropped because it is no longer needed

src/lib/containers/entity_finder/details/view/DetailsView.tsx
```diff
@@ -68,7 +70,6 @@ export type DetailsViewProps = EntityDetailsListSharedProps & {
 export type DetailsViewRowData = (EntityHeader | ProjectHeader | Hit) & {
   entityType: EntityType
   isSelected: boolean
-  isHidden: boolean
```
nickgros, line 71: Hidden items are now excluded from RowData, so this property is no longer needed.

src/lib/containers/entity_finder/details/view/DetailsView.tsx
```diff
@@ -233,19 +233,25 @@ export const DetailsView: React.FunctionComponent<DetailsViewProps> = ({
...
+          isSelected: appearance === 'selected',
+          isDisabled: appearance === 'disabled',
+          isVersionableEntity: isVersionableEntityType(entityType),
+          currentSelectedVersion: selected.get(entity.id),
+        })
+      }
+      return entities
+    },
+    [],
+  )
```
nickgros, line 254: This fixes SWC-5845 -- hidden items are omitted.

src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -309,10 +322,11 @@ export const EntityFinder: React.FunctionComponent<EntityFinderProps> = ({
                       headerList: searchByIdResults,
                     }
                   : {
-                      type:
-                        EntityDetailsListDataConfigurationType.ENTITY_SEARCH,
+                      type: EntityDetailsListDataConfigurationType.ENTITY_SEARCH,
                       query: {
                         queryTerm: searchTerms,
+                        booleanQuery:
+                          searchForOnlyTypesBooleanQuery(selectableTypes),
```
nickgros, line 329: Only search for selectableTypes

src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -40,6 +41,20 @@ const DEFAULT_VISIBLE_TYPES = [EntityType.PROJECT, EntityType.FOLDER]
...
+  // So we will negate searching for all omitted types.
+  const allTypes = Object.values(EntityType)
+  const typesToOmit = allTypes.filter(type => !entityTypes.includes(type))
+  return typesToOmit.map(type => ({
+    key: 'node_type',
+    value: type.toString(),
+    not: true,
+  }))
+}
+
```
nickgros, line 57: Our search API doesn't have a logical "OR" option, but we can negate searching for all other types, as long as our EntityType enum is kept up-to-date.

src/__tests__/lib/containers/entity_finder/EntityFinder.test.tsx
```diff
@@ -408,6 +408,35 @@ describe('EntityFinder tests', () => {
...
+    )
+    await waitFor(() =>
+      expect(mockDetailsList).toBeCalledWith(
+        expect.objectContaining({
+          configuration: {
+            type: EntityDetailsListDataConfigurationType.ENTITY_SEARCH,
+            query: {
+              queryTerm: queryTerms,
+              // The lone selectable type should not have been excluded
+              booleanQuery: expect.not.arrayContaining([
```
jay-hodgson, line 433: I'd rather not admit how long it took me to recognize the expect._not_... in the test.

jay-hodgson (approved):

src/__tests__/lib/containers/entity_finder/EntityFinder.test.tsx
nickgros: Jest matchers definite leave a lot to be desired in terms of readability, especially compared to JUnit/Mockito!

PR #1279, opened 2021-11-03
PORTALS-2034

jay-hodgson:


jay-hodgson:
Most of these changes are to support the new NF GFF Tool Details page, where each sub-section of data is driven by a query.
Screen Shot 2021-11-03 at 4 09 11 PM

package.json
```diff
@@ -202,7 +202,8 @@
     "build:rollup": "rollup -c && cp -r src/umd dist/",
     "serve:docs": "DISABLE_ESLINT_PLUGIN=true FAST_REFRESH=false npx styleguidist server --open",
     "build:docs": "yarn run build-css && DISABLE_ESLINT_PLUGIN=true FAST_REFRESH=false npx styleguidist build",
-    "link-src": "yarn link && cd ./node_modules/react && yarn link && cd ../react-router && yarn link && cd ../react-router-dom && yarn link && cd ../../"
+    "link-src": "yarn link && cd ./node_modules/react && yarn link && cd ../react-router && yarn link && cd ../react-router-dom && yarn link && cd ../../",
+    "unlink-src": "yarn unlink && cd ./node_modules/react && yarn unlink && cd ../react-router && yarn unlink && cd ../react-router-dom && yarn unlink && cd ../../"
```
jay-hodgson, line 206: Adding the unlink-src script for this project and portals.

src/lib/containers/CardContainer.tsx
```diff
@@ -102,15 +103,7 @@ export const CardContainer = (props: CardContainerProps) => {
...
-      <>
-        <p className="SRC-no-results-title">
-          There is currently no content here.
-        </p>
-        <p className="SRC-no-results-description">
-          Information is always being updated, so check back later to see if
-          content has been added.
-        </p>
-      </>
+      <NoContentAvailable />
```
jay-hodgson, line 106: Refactored into a reusable component, then decided against reusing it (for the case when no people are found in the UserCardListRotate component).  So, it's available to you, but unused here.

src/lib/containers/UserCardListRotate.tsx
```diff
@@ -137,6 +137,11 @@ const UserCardListRotate: React.FunctionComponent<UserCardListRotateProps> = ({
   return (
     <div className="UserCardListRotate bootstrap-4-backport">
       {isLoading && loadingScreen}
+      {!isLoading && userIds.length === 0 && (
+        <p className="font-italic">
+          No one was found.
```
jay-hodgson, line 142: Needed to show something here, since the section is always shown when I use this component on the Details page.  When shown on a Home page, we should never encounter this case.

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -9,7 +9,9 @@ import {
   TOOL,
   STUDY,
   EXPERIMENTAL,
+  EXPERIMENTAL_TOOL,
```
jay-hodgson, line 12: EXPERIMENTAL_TOOL is synonymous with EXPERIMENTAL (design showed "EXPERIMENTAL TOOL" as the type)

src/lib/utils/synapseTypes/Table/SelectColumn.ts
```diff
@@ -21,5 +21,5 @@ export type SelectColumn = {
   name: string // The required display name of the column
   columnType: EntityColumnType
   // The column type determines the type of data that can be stored in a co
-  id: string // The optional ID of the select column, if this is a direct column selected
+  id?: string // The optional ID of the select column, if this is a direct column selected
```
jay-hodgson, line 24: SelectColumn.id is not returned when the column is aliased

src/lib/containers/SubsectionRowRenderer.tsx
```diff
@@ -0,0 +1,167 @@
...
+        query: {
+          sql: sqlUsed,
+          limit,
+        },
+      }
+
+      const queryResultBundle = await SynapseClient.getFullQueryTableResults(
+        request,
+        accessToken,
+      )
```
nickgros, line 68: I didn't realize we had this method and I'm not crazy about it. I'm guessing that we're pretty confident we'll never have to render a huge table in this component.

src/lib/containers/SubsectionRowRenderer.tsx
```diff
@@ -0,0 +1,167 @@
...
+      {!isLoading && rowSet && rowSet.rows.length > 0 && (
+        rowSet.headers.map((selectColumn, colIndex) => {
+          // If a link column was provided (that contain URLs), do not create a page sub-section for that column.
+          if (columnLink && selectColumn.name == columnLink.linkColumnName) {
+            return <></>
+          }
+          return <div key={`${colIndex}`} className="SubsectionRowRenderer__item" role="table">
+            {/* Is this a page subsection (default), or a page section? */}
+            {!columnNameIsSectionTitle && <h4 className="SubsectionRowRenderer__item__subsection-title" role='heading'>{selectColumn.name}</h4>}
+            {columnNameIsSectionTitle && <><h2 className="SubsectionRowRenderer__item__section-title" role='heading'>{selectColumn.name}</h2><hr /></>}
```
nickgros, line 113: Would this be a good place to use the `Typography` component?

src/__tests__/lib/containers/SubsectionRowRenderer.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        matchColumnName: 'Catalog Number',
+        isMarkdown: false,
+      },
+      friendlyValuesMap: {
+        'row2': 'Friendly Row 2 Value'
+      }
+    })
+
+    await waitFor(() => {
+      expect(screen.getAllByRole('heading').length).toBe(1)
```
nickgros, line 44: Great use of aria roles + React Testing Library

nickgros (approved):

src/lib/containers/SubsectionRowRenderer.tsx
jay-hodgson: Doesn't need to be used here, so I'll change it!  Useful for those edge cases when we need all of the data before we can show anything (typically plotting)

src/lib/containers/SubsectionRowRenderer.tsx
jay-hodgson: Good idea!

src/lib/containers/SubsectionRowRenderer.tsx
```diff
@@ -62,7 +63,7 @@ const SubsectionRowRenderer: React.FunctionComponent<SubsectionRowRendererProps>
         },
       }
 
-      const queryResultBundle = await SynapseClient.getFullQueryTableResults(
+      const queryResultBundle = await SynapseClient.getQueryTableResults(
```
nickgros, line 66: Since you're not getting the full results, optionally you could now use (useGetQueryResultBundle)[https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/62ab80316421af5d3126739e5e81b9fca5745156/src/lib/utils/hooks/SynapseAPI/useGetQueryResultBundle.ts#L7] if it works with the rest of this component.

src/lib/containers/UserCardMedium.tsx
```diff
@@ -340,4 +342,18 @@ export const UserCardMedium: React.FC<UserCardMediumProps> = ({
...
+  return <div
+    className="cardContainer SRC-userCard SRC-userCardMediumUp"
+    style={{width:'380px'}}
+  >
+    <Skeleton variant='circle' width='80px' height='80px'/>
+    <div style={{width:'250px'}}>
+      <SkeletonTable numCols={1} numRows={2}/>
+    </div>
+  </div>
+}
```
nickgros, line 355: I like this

nickgros (approved):
Looks great, approved 2x

PR #1286, opened 2021-11-08
PORTALS-2052

jay-hodgson:


src/lib/containers/Login.tsx
```diff
@@ -131,8 +133,8 @@ class Login extends React.Component<Props, State> {
     event.preventDefault()
     const redirectUrl = this.props.googleRedirectUrl
       ? this.props.googleRedirectUrl
-      : `${SynapseClient.getRootURL()}?provider=${SynapseClient.AUTH_PROVIDER}`
-    SynapseClient.oAuthUrlRequest(SynapseClient.AUTH_PROVIDER, redirectUrl)
+      : `${SynapseClient.getRootURL()}?provider=${GOOGLE_AUTH_PROVIDER}`
```
jay-hodgson, line 136: Login component still only supports Google SSO login for now.  We can add other providers (with their specific UI) later, as they become available.

src/lib/utils/SynapseClient.ts
```diff
@@ -1310,18 +1309,19 @@ This function should be called whenever the root App is initialized
...
   // in test environment the searchParams isn't defined
-  const { searchParams } = code
+  const { searchParams } = fullUrl
   if (!searchParams) {
     return
   }
-  code = searchParams.get('code')
-  if (code) {
+  const code = searchParams.get('code')
+  const provider = searchParams.get('provider')
```
jay-hodgson, line 1319: Use the value for the "provider" query param instead of hardcoding to Google.

src/lib/containers/Login.tsx
```diff
@@ -131,8 +133,8 @@ class Login extends React.Component<Props, State> {
     event.preventDefault()
     const redirectUrl = this.props.googleRedirectUrl
       ? this.props.googleRedirectUrl
```
conundrumgirl, line 135: What does googleRedirectUrl do? Would that need to change as well once we have different providers?

src/lib/containers/Login.tsx
```diff
@@ -8,6 +8,8 @@ import {
 } from '../utils/functions/getEndpoint'
 import { GoogleIcon24 } from '../assets/GoogleIcon24'
 
+export const GOOGLE_AUTH_PROVIDER = 'GOOGLE_OAUTH_2_0'
```
conundrumgirl, line 11: If we'll have multiple providers would it make sense to have an enum or a map for provides? 
e..g PROVIDERS = {
GOOGLE: 'GOOGLE_OAUTH_2_0',
....
}

conundrumgirl (approved):
looks good. Just a couple of questions

src/lib/containers/Login.tsx
jay-hodgson: Absolutely!  When Login is updated to support more than one, this should definitely be a map.  I'll make the change now to set up the next engineer for the enhancement.

src/lib/containers/Login.tsx
jay-hodgson: This is the redirect URL that we send to the OAuth provider (only Google for now).  If it's registered to our OAuth client for that provider, then they will follow it on successful login (and give us the access code).
It was poorly named prop if it's not Google.  Renaming means that this prop needs to be updated when this version of SRC is pulled into Synapse.org (otherwise it will break Google SSO).  Opened [SWC-5849](https://sagebionetworks.jira.com/browse/SWC-5849) to track.  Thanks!

conundrumgirl (approved):

PR #1295, opened 2021-11-16
PORTALS-2051: make sure all parts of the svg are rendered, no matter …

jay-hodgson:
…what other svgs are on the same page (unique element ids)


jay-hodgson:
On the resources page, this card icon was displayed properly:
Screen Shot 2021-11-16 at 11 27 59 AM

However, when rendered on the details page (where another svg with a duplicate element ID was present), only part of the icon was rendered:
Screen Shot 2021-11-16 at 11 27 52 AM

jay-hodgson:
Merging to update nfgff portals branch with new SRC, but will update if anything is found during the code review!

nickgros (commented):
Looked like everything matched up. Great sleuthing!

PR #1296, opened 2021-11-16
SWC-5784 - New design system color palettes and buttons

nickgros:
Utilizing this in the new component. I think we will likely apply these changes to Synapse before the component comes out of experimental mode.

The only visual change is that the primary palette's '500' item is no longer a super-saturated version of the primary action color. This change will be visible in a few Synapse.org components and portals, but shouldn't break anything.

### DIFF
--- a/src/lib/style/abstracts/_variables.scss
+++ b/src/lib/style/abstracts/_variables.scss
@@ -1,6 +1,28 @@
+@use 'sass:color';
+@use 'sass:map';
+
 // -----------------------------------------------------------------------------
 // This file contains all application-wide Sass variables.
 // -----------------------------------------------------------------------------
+
+// Colors defined by design: https://www.figma.com/file/0oPm5lLSUva8kyfVNMS6FA/Synapse-Style-%26-Component-Library?node-id=77%3A1232
+$colors: (
+  'gray-1000': #22252a,
+  'gray-900': #353a3f,
+  'gray-800': #4a5056,
+  'gray-700': #878e95,
+  'gray-600': #aeb5bc,
+  'gray-500': #d0d4d9,
+  'gray-400': #dfe2e6,
+  'gray-300': #eaecee,
+  'gray-200': #f1f3f5,
+  'gray-100': #fbfbfc,
+  'white': #ffffff,
+  'success': #32a330,
+  'info': #017fa5,
+  'warning': #cc9f00,
+  'error': #c13415,
+) !default;
 
 $gray-dark: #dcdcdc !default;
 $gray-light: #f9f9f9 !default;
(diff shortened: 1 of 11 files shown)


src/lib/style/abstracts/_variables.scss
```diff
@@ -1,7 +1,29 @@
...
+  'gray-400': #dfe2e6,
+  'gray-300': #eaecee,
+  'gray-200': #f1f3f5,
+  'gray-100': #fbfbfc,
+  'white': #ffffff,
+  'success': #32a330,
+  'info': #017fa5,
+  'warning': #cc9f00,
+  'error': #c13415,
+) !default;
```
nickgros, line 25: There is still a lot of work to do to use these colors rather than the variables below.

src/lib/style/abstracts/_variables.scss
```diff
@@ -51,37 +73,28 @@ $primary-action-color-active: adjust-color(
...
+$primary-color-palette: (
+  100: adjust-color($primary-action-color, $saturation: -25%, $lightness: 50%),
+  200: adjust-color($primary-action-color, $saturation: -13%, $lightness: 20%),
+  300: adjust-color($primary-action-color, $saturation: -3%, $lightness: 10%),
+  400: adjust-color($primary-action-color, $saturation: -1%, $lightness: 5%),
+  500: $primary-action-color,
+  600: adjust-color($primary-action-color, $saturation: 5%, $lightness: -4%),
+  700: adjust-color($primary-action-color, $saturation: 11%, $lightness: -8%),
+  800: adjust-color($primary-action-color, $saturation: 18%, $lightness: -10%),
+  900: adjust-color($primary-action-color, $saturation: 24%, $lightness: -18%),
```
nickgros, line 85: As before, we impute colors if a color palette isn't supplied. This will be useful for portals.

nickgros: The 500 value here has changed; before it was $primary-action-color +saturation -lightness. Now we just use the $primary-action-color.

Some icons/buttons may appear slightly dimmer. I don't think it's a huge deal, but I can check with design on this if it's a bigger issue than I think it is.

src/lib/style/bootstrap4_backports/_base-import.scss
```diff
@@ -158,4 +159,49 @@
...
+      $border: map.get(SRC.$primary-color-palette, 500),
+      $hover-background: map.get(SRC.$primary-color-palette, 600),
+      $hover-border: map.get(SRC.$primary-color-palette, 600),
+      $active-background: map.get(SRC.$primary-color-palette, 700),
+      $active-border: map.get(SRC.$primary-color-palette, 700)
+    );
+    border-radius: 0px;
+    font-weight: 900;
+    &:hover {
+      box-shadow: none; // ?
```
nickgros, line 177: Need to clarify w/ design about this but IMO a box shadow on these new buttons would look out-of-place.

jay-hodgson (approved):

PR #1300, opened 2021-11-17
SWC-5784 - Refactor SynapseTableCell

nickgros:
I haven't yet figured out how I'm going to take advantage of SynapseTableCell but we're getting closer to something reusable.

* SynapseTableCell now uses switch/case on the ColumnType rather than passing indices of specific column types.
* EntityLink now supports passing a synId
* EntityLink now supports showing the entity  without linking (I will probably rename this component later). This will be used in the new dataset editor component.
* Added useGetEntity


src/lib/containers/EntityLink.tsx
```diff
@@ -1,31 +1,66 @@
...
-        target="_blank"
-        rel="noopener noreferrer"
-        href={`${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!Synapse:${id}`}
-      >
-        <EntityTypeIcon type={type} style={{ marginRight: '5px' }} />
-        {name}
-      </a>
-    </p>
-  )
+export const EntityLink = (props: EntityLinkProps) => {
```
nickgros, line 18: Will rename this later...didn't want to get logic changes lost in the diff due to a file rename

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -678,22 +682,21 @@ export default class SynapseTable extends React.Component<
    *
    * @memberof SynapseTable
    */
-  private handlePaginationClick = (eventType: string) => (
```
nickgros, line 681: This is whitespace only, thx Prettier

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -80,18 +81,16 @@ type Info = {
   index: number
   name: string
 }
-export interface Dictionary<T> {
-  [key: string]: T
-}
```
nickgros, line 85: Dictionary<T> is the same thing as Record<string, T>, which is built-in, so I removed this.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -749,39 +752,6 @@ export default class SynapseTable extends React.Component<
...
-    )
-    const entityIdListColumnIndicies = getColumnIndiciesWithType(
-      data,
-      ColumnType.ENTITYID_LIST,
-    )
-    const otherListColumnIndicies = getColumnIndiciesWithType(
-      data,
-      ColumnType.STRING_LIST,
-      ColumnType.INTEGER_LIST,
-    )
```
nickgros, line 784: These are no longer needed since we are passing the ColumnType

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -820,28 +791,27 @@ export default class SynapseTable extends React.Component<
...
+                  <SynapseCardLabel
+                    value={columnValue}
+                    columnName={columnName}
+                    selectColumns={selectColumns}
+                    columnModels={columnModels}
+                    isHeader={false}
+                    labelLink={columnLinkConfig}
+                    rowData={row.values}
+                  />
+                )}
```
nickgros, line 804: I pulled this out of the SynapseTableCell component. My reasoning for this is that I'm guessing that using a SynapseCardLabel is exclusively for Portals use cases (which SynapseTable is for). I'd like to reuse SynapseTableCell for other purposes, like the new Dataset editor.

Many of these components could use better names.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -863,31 +833,33 @@ export default class SynapseTable extends React.Component<
           </td>,
         )
       }
-      const isFileEntity:boolean = mapEntityIdToHeader[rowSynapseId]?.type == 'org.sagebionetworks.repo.model.FileEntity'
```
nickgros, line 866: The rest is whitespace changes

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -4,70 +4,39 @@ import React from 'react'
...
-  entityColumnIndicies: number[]
-  userColumnIndicies: number[]
-  dateColumnIndicies: number[]
-  dateListColumnIndicies: number[]
-  booleanListColumnIndicies: number[]
-  fileHandleIdColumnIndicies: number[]
-  entityIdListColumnIndicies: number[]
-  otherListColumnIndicies: number[]
-  colIndex: number
+  columnType: ColumnType
```
nickgros, line 21: Instead of checking if an index is in one of many arrays, just get the ColumnType

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -4,70 +4,39 @@ import React from 'react'
...
   columnValue: string
   isBold: string
-  mapEntityIdToHeader: Dictionary<EntityHeader>
-  mapUserIdToHeader: Dictionary<any>
-  columnLinkConfig?: MarkdownLink | CardLink | ColumnSpecifiedLink
```
nickgros, line 40: We no longer need columnLinkConfig, selectColumns, columnModels, or rowData because the SynapseCardLink logic has been extracted from this component

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -77,150 +46,152 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
-  } else if (
-    userColumnIndicies.includes(colIndex) &&
-    Object.prototype.hasOwnProperty.call(mapUserIdToHeader, columnValue)
-  ) {
-    const { ownerId, userName } = mapUserIdToHeader[columnValue]
-    if (mapUserIdToHeader[columnValue].isIndividual === false) {
-      // isUserGroupHeader
-      const icon = userName === AUTHENTICATED_USERS ? 'globe-americas' : 'users'
-      if (userName === AUTHENTICATED_USERS) {
+  switch (columnType) {
```
nickgros, line 50: The cases should be in the same order as the if blocks. I like this a lot more because we can more quickly see how we are (or aren't) handling certain column types.

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -77,150 +46,152 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
+                callbackFn={noop}
+                idForToolTip={`${columnName}_${rowIndex}`}
+              >
+                <p className={isBold}> {columnValue}</p>
+              </ElementWithTooltip>
+            </div>
+          )
+        }
+      }
+      break
```
nickgros, line 194: Should I add a `default` case? Should it throw a warning?

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -77,150 +46,152 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
+            />
+          )
+        }
+      }
+      break
+    case ColumnType.STRING:
+    case ColumnType.DOUBLE:
+    case ColumnType.INTEGER:
+    case ColumnType.BOOLEAN:
+    case ColumnType.LINK:
```
nickgros, line 174: I need to figure out what a link column is. Sounds like it should be handled differently.

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -77,150 +46,152 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
+              >
+                <p className={isBold}> {columnValue}</p>
+              </ElementWithTooltip>
+            </div>
+          )
+        }
+      }
+      break
   }
+  return <></>
```
nickgros, line 196: I need a more sensible default return value/error.

src/lib/containers/EntityLink.tsx
```diff
@@ -1,31 +1,66 @@
...
+      return (
+        <p className={className}>
+          <EntityTypeIcon type={type} style={{ marginRight: '5px' }} />
+          {entity.name}
+        </p>
+      )
+    }
+  } else {
+    // Entity has not been fetched yet
+    // TODO: Handle not found, unauthorized
```
nickgros, line 63: Do we have any existing components/patterns for these cases?

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -90,137 +76,157 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
     }
+    default:
+      console.warn(
+        `ColumnType ${columnType} has unspecified handler. Rendering the column value.`,
+      )
+      return <p className={isBold}>{columnValue}</p>
   }
+  // We can reach this if we don't get a mapping of IDs to entities or principals.
+  // TODO: If we don't have a id:data mapping, we should render a component that can fetch the required data, rather than breaking from the case.
+  return <p className={isBold}>{columnValue}</p>
```
nickgros, line 231: We will only be able to get here if we don't have the entity ID or user ID stored in the ID->data mappings that were passed as props. I'm guessing that this happens when we get a 403/404.

So my guess is that all of the cases where this can happen are 
* cache miss (these probably aren't happening)
* 403 error on an entity
* 404 error on an entity/user/team

It would make more sense to me to render a component that can fetch the data (so we don't have to worry about cache misses), and explicitly handle the 403/404 cases.

Since that would be a substantial change, I didn't do it here.

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -147,19 +150,22 @@ export const SchemaDrivenAnnotationEditor: React.FunctionComponent<SchemaDrivenA
...
+      setShowSubmissionError(true)
+    },
+  })
+
+  function submitChangedEntity() {
+    mutation.mutate({
+      ...dropNullishArrayValues(formData ?? {}),
+      ...entityJson,
+    } as EntityJson)
+  }
```
nickgros, line 169: Instead of calling  `mutation.mutate()`, call `submitChangedEntity`	which calls `mutation.mutate(newJson)`

src/lib/utils/hooks/SynapseAPI/useEntity.ts
```diff
@@ -134,19 +138,20 @@ export function useGetJson(
...
+      return SynapseClient.updateEntityJson(entityId, json, accessToken)
+    },
     {
       ...options,
       onSuccess: async (data, variables, ctx) => {
+        const entityId = data.id
+
         await queryClient.invalidateQueries([accessToken, 'entity', entityId], {
           exact: false,
         })
```
nickgros, line 157: Removing the mutation key because it doesn't seem to be very useful (and it's causing me to get new typescript errors which are breaking the build). Also, now we should call `mutation.mutate` with new data, instead of passing the new data into the hook fn.

src/lib/containers/EntityLink.tsx
jay-hodgson: As you know, we have an ErrorBanner, which is not a good fit for a table cell entity load error.  I'd recommend the UI proposed for files that are ineligible for package download, found [here](https://www.figma.com/file/usEyd7hdEd0uYyN04yt0WO/branch/PnQZRd4YeLhvFj6IVSaVQb/Data-Lists-for-Download-R1?node-id=2282%3A443865).  Where the error would be displayed in the tooltip.
Screen Shot 2021-11-18 at 10 25 12 AM

jay-hodgson (approved):
Nice work, thanks Nick!

jay-hodgson (approved):

PR #1316, opened 2021-12-03
SWC-5861

nickgros:
Tests for dataset items editor.
No tests for specific renderers, but I've spent too much time on this. I'll come back to it later.

### DIFF
--- a/src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
+++ b/src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
@@ -42,8 +42,6 @@ function referenceToDatasetItem(reference: Reference): DatasetItem {
   }
 }
 // The Entity Finder is complicated to use and would require setting up a lot of API mocks, so we'll just mock the component.
-// const EntityFinderModal = require('../../../../../lib/containers/entity_finder/EntityFinderModal')
-// EntityFinderModal.EntityFinderModal = jest.fn(() => 'null')
 jest.mock(
   '../../../../../lib/containers/entity_finder/EntityFinderModal',
   () => {
@@ -55,7 +53,7 @@ jest.mock(
 const mockEntityFinder = EntityFinderModal as unknown as jest.Mock<
   (props: Partial<EntityFinderModalProps>) => JSX.Element
 >
-
+const mockEntityFinderButtonText = 'Add Items From Entity Finder'
 function mockEntityFinderToAddItems(items: Array<Reference>) {
   mockEntityFinder.mockImplementation(({ show, onConfirm }) => {
     return (
@@ -100,7 +98,6 @@ async function selectIndividualItem(id: string) {
 
 async function removeItem(id: string) {
   await selectIndividualItem(id)
-
   clickRemove()
 }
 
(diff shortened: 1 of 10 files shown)


src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+        }),
+      ),
+    )
+  })
+
+  /**
+   * TODO: This test passes on its own but fails in the suite.
+   * We're calling resetHandlers and clearAllMocks, so I have no idea why it isn't working.
+   */
+  it.skip('Handles an error on Save by displaying a toast and not calling the callback', async () => {
```
nickgros, line 477: I spent a ton of time trying to figure this out...the test works on its own and the behavior is correct when I take these steps in the browser. It just fails in the test suite. Mocks are reset after each test, so I have no idea why this isn't working.

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -400,7 +400,7 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
 
         <Button
           variant="outline"
-          disabled={selectedIds.size === DATASET_MAX_ITEM_LIMIT}
+          disabled={datasetToUpdate == null}
           onClick={() => setShowEntityFinder(true)}
         >
           Add Items
```
nickgros, line 406: Removing the max number of items requirement, and just restricting adding items until the dataset is loaded.

I don't mind removing it here because this button isn't a particularly helpful place to enforce the requirement anyways. The user could still end up getting the error from the server. We should enforce the requirement in the UI with a conditional warning or in the Entity Finder.

Additionally, the limit is changing for Views, so it may change for datasets as well.

src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+              {mockEntityFinderButtonText}
+            </button>
+          </div>
+        ) : null}
+      </>
+    )
+  })
+}
+
+function addItemsViaEntityFinder() {
```
jay-hodgson, line 73: It seems like this function is just testing your mock entity finder implementation (and the ability to find the Add Items button on the screen).

jay-hodgson: Wait, is this essentially just a utility function to invoke the onConfirm() function given to the entity finder modal

nickgros: It is basically a utility function. I could add a comment or remove the `expect`s to make that more clear

src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
jay-hodgson: This is unfortunate, and something I've also encountered and can't explain.  @conundrumgirl : Any ideas?

jay-hodgson (approved):

PR #1324, opened 2021-12-06
Release 2.0.63

jay-hodgson:
SWC-5878:  Minor bug fix for Download List and general improvement to the UX when list is updated


jay-hodgson:
Looks like we may have an unstable test:
FAIL src/__tests__/lib/containers/entity/annotations/SchemaDrivenAnnotationsEditor.test.tsx

PR #1328, opened 2021-12-08
SWC-5750

nickgros:
Summary of changes:
* Use esbuild instead of rollup for the UMD build. Reduces this step from ~40s to ~4s. The TypeScript compilation step takes ~40s, so this is now ~45s in total instead of ~80s.
* ~Upgrade to Yarn 3~
	* ~The motivation for this was increased consistency using `yarn link`. In newer versions of yarn, `link` works differently to enforce the presence/resolution of dependencies. Without this, we had an intermittent issue where a dependency required to build SRC's SCSS (which we build in Synapse.org and Portals) could be missing.~
	* ~This unfortunately breaks our `link` commands in Portals.~
		* ~Now, from portals, you call `yarn link ../path/to/Synapse-React-Client/` to link the package and `yarn unlink synapse-react-client` to unlink it. No need to run any commands in the `Synapse-React-Client` folder.~
		* ~This also modifies `package.json`, so care should be taken to not commit the relevant change without calling `unlink`.~
* ~Upgrade dependencies to match more recent versions in SWC/portals. `yarn link` now throws an error if there's a dependency mismatch that isn't clarified in package.json's `resolutions`. We'll upgrade these packages for now so we don't have to specify resolutions.~

I'm getting many console errors in Synapse.org similar to this when I use the new build:
image

I'm not sure why this is happening. Apparently, a proptypes assertion fails for many Material-UI components using React.forwardRef. The components appear just fine for me, and the warnings don't appear in production mode, so I'm inclined to ignore them for now. I tried upgrading to Material-UI v5 but but ran into other issues, mainly that Material UI requires React 17 and *forces* you to use the new JSX transform, which esbuild doesn't currently support without first using babel.


.yarnrc.yml
```diff
@@ -0,0 +1,2 @@
+yarnPath: .yarn/releases/yarn-3.1.1.cjs
```
nickgros, line 1: [Yarn docs instruct us](https://yarnpkg.com/getting-started/migration#step-by-step) to commit its cjs module, not a mistake.

esbuild.config.mjs
```diff
@@ -0,0 +1,114 @@
...
+  entryPoints: ['src/lib/rollup.index.ts'],
+  bundle: true,
+  platform: 'browser',
+  target: 'es2015',
+  globalName: 'SRC',
+  plugins: [
+    sassPlugin.sassPlugin({
+      includePaths: ['node_modules'],
+      outfile: './src/umd/synapse-react-client.production.styles.css',
+    }),
```
nickgros, line 48: Build Sass

esbuild.config.mjs
```diff
@@ -0,0 +1,114 @@
...
+    svgrPlugin({ ref: true }),
```
nickgros, line 49: Handle SVGs

esbuild.config.mjs
```diff
@@ -0,0 +1,114 @@
...
+    metafile: true,
+    outfile: './src/umd/synapse-react-client.production.min.js',
+  })
+  .then(result => {
+    const metafile = result.metafile
+    const outputMb = Object.entries(metafile.outputs).forEach(
+      ([key, value]) => {
+        console.log(`${key}: ${(value.bytes / 1024 / 1024).toFixed(2)} MB`)
+      },
+    )
```
nickgros, line 111: Output the size of the file in MB

package.json
```diff
@@ -180,30 +174,30 @@
     ]
   },
   "scripts": {
+    "clean:build": "rimraf dist",
+    "clean:umd": "rimraf dist/umd",
```
nickgros, line 178: We should clean the dist results when we build, just to be safe.

package.json
```diff
@@ -180,30 +174,30 @@
...
+    "start-deprecated-demo": "npm-run-all -p watch:css start-js",
     "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!(lodash-es|jest*)/)\" , --env=jsdom --setupFilesAfterEnv ./src/setupTests.js --resetMocks false",
     "test:coverage": "yarn test --coverage",
     "test:noPrompt": "CI=true yarn test",
     "eject": "react-scripts eject",
     "type-check": "tsc --noEmit",
     "type-check:watch": "yarn run type-check --watch",
-    "build": "yarn run build-css && yarn run build:js && yarn run build:copy-assets && yarn build:rollup",
+    "build": "yarn clean:build && yarn run build:css && yarn run build:js && yarn run build:copy-assets && yarn build:esbuild",
+    "build:css": "sass src/lib/style/main.scss src/lib/style/main.css --load-path=node_modules && sass src/demo/style:src/demo/style --load-path=node_modules",
```
nickgros, line 195: Dropping '.' from sass load-path (unnecessary) and adding `node_modules`, which is more common. IIRC the only reason we didn't do this before was because of a rollup plugin issue.

src/lib/style/components/_markdown.scss
```diff
@@ -1,7 +1,6 @@
 .markdowntable {
   border: 1px solid #dddddd;
   border-collapse: separate;
-  *border-collapse: collapsed;
```
nickgros, line 4: I'm assuming this rule never worked.

src/lib/containers/AccountLevelBadge.tsx
```diff
@@ -2,9 +2,9 @@ import { UserBundle } from '../utils/synapseTypes'
 import * as React from 'react'
 import { useState, useEffect } from 'react'
 import { SynapseClient, SynapseConstants } from '../utils'
-import { ReactComponent as Registered } from '../assets/icons/account-registered.svg'
-import { ReactComponent as Certified } from '../assets/icons/account-certified.svg'
-import { ReactComponent as Validated } from '../assets/icons/account-validated.svg'
+import Registered from '../assets/icons/account-registered.svg'
+import Certified from '../assets/icons/account-certified.svg'
+import Validated from '../assets/icons/account-validated.svg'
```
nickgros, line 7: This style of imports can't be handled by esbuild without bringing in webpack or babel (which undermines the benefits of using esbuild).

At some point, we should turn these SVGs into react components or call them via IconSvg

jay-hodgson:
Is there a significant change in size for the umd build output?
So this will break the convenience script command to link SRC in the portals project, right?  Any other changes to portals?

nickgros:
> Is there a significant change in size for the umd build output?

Not really. According to [unpkg](https://unpkg.com/browse/synapse-react-client@2.0.62/dist/umd/), our Rollup build is 1.83 MB. esbuild outputs a 1.77 MB build.

> So this will break the convenience script command to link SRC in the portals project, right? Any other changes to portals?

I changed my mind on upgrading to Yarn 3--I cannot find a way to make linking work in the same way it works in Yarn 1. SWC and Portals will need SCSS updates, but that should be it. I'm drafting those PRs:
https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4724
https://github.com/Sage-Bionetworks/portals/pull/709/files

PR #1329, opened 2021-12-08
SWC-5888

jay-hodgson:


jay-hodgson:
Screen Shot 2021-12-08 at 12 13 40 PM

src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -126,96 +126,98 @@ export const DownloadCartPage:React.FunctionComponent<DownloadCartPageProps> = (
         <div>
           {data.numberOfFilesAvailableForDownload > 0 && 
             <div className="DownloadListTabContent">
-              <div className="subSectionOverview">
-                <div>
-                  <div className="headlineWithHelp">
-                    <Typography variant={'headline3'}>
-                      <IconSvg options={{ icon: 'packagableFile' }} /> Web Download (.ZIP Packages)
+              <div className="subSectionOverviewContainer">
```
jay-hodgson, line 129: Added a new wrapper div (subSectionOverviewContainer) that has the background color.  subSectionOverview still has the display: grid and is now in a bootstrap container (fixes the ticket).

src/lib/style/components/_download-cart-page.scss
```diff
@@ -92,6 +92,12 @@
       display: flex;
       column-gap: 10px;
       >* {
+        &:first-child {
```
jay-hodgson, line 95: The first item (leftmost) should not have padding on the left (it's now in a container), and the last item (rightmost) should not have padding on the right.

nickgros (approved):

PR #1330, opened 2021-12-09
SWC-5884

nickgros:


src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -430,28 +437,30 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
-              onRowsRendered={rebuildTooltip}
-              onScroll={rebuildTooltip}
-            ></BaseTable>
+            <AutoResizer height={TABLE_HEIGHT} onResize={rebuildTooltip}>
+              {({ height, width }: { height: number; width: number }) => (
+                <BaseTable
+                  classPrefix="DatasetEditorTable"
+                  data={tableData}
+                  height={height}
+                  width={width > totalColumnWidth ? width : totalColumnWidth}
```
nickgros, line 446: If the container width is greater than the sum of all column widths, use the container width.

src/lib/style/bootstrap4_backports/_overrides.scss
```diff
@@ -112,11 +112,5 @@ html#synapse\.org {
         padding-bottom: 0px;
       }
     }
-    .modal-footer {
-      & > * {
-        margin-left: 0px;
-        margin-right: 0px;
-      }
-    }
```
nickgros, line 120: This was a rule that was relevant for an individual modal design (which currently only appears in alpha mode, so it will be updated before any user sees it). It shouldn't be applied to all modals (buttons in the footer need space between them).

src/lib/style/components/_fluid-modal.scss
```diff
@@ -28,11 +28,6 @@
         overflow-y: auto;
       }
       .modal-footer {
-        min-height: 45px;
-        max-height: 60px;
-        height: 6vh;
-        margin-bottom: 15px;
-
```
nickgros, line 35: The default footer height for Modals is sensible enough, we don't need to tweak it and create new problems.

jay-hodgson:
Dataset Items Editor tests › Allows selecting versions and updates the selected version when a new one is picked
Unable to find role="listbox"
      452 |     expect(mockFileReference.targetVersionNumber).not.toEqual(1)
      453 |
   > 454 |     userEvent.selectOptions(await screen.findByRole('listbox'), '1')
      455 |
      456 |     await clickSave()
      457 |

jay-hodgson (approved):

PR #1331, opened 2021-12-09
SWC-5867

jay-hodgson:


jay-hodgson:
HelpPopover:
Screen Shot 2021-12-09 at 3 06 10 PM

New action:
Screen Shot 2021-12-09 at 2 42 47 PM

Direct download programmatic instructions:
Screen Shot 2021-12-09 at 2 43 08 PM

src/lib/containers/HelpPopover.tsx
```diff
@@ -0,0 +1,42 @@
...
+  
+  const actionButtonConfig = helpUrl ? {
+    content: <>More info</>,
+    closePopoverOnClick: true,
+    onClick: () => window.open(helpUrl, '_blank'),
+    variant: 'primary'
+  } : undefined
+  return (
+    <>
+      <MarkdownPopover
```
jay-hodgson, line 28: I see this being implemented in a few ways across the codebase, like to standardize and make reusable.

src/lib/containers/ProgrammaticInstructionsModal.tsx
```diff
@@ -0,0 +1,155 @@
+import React, { useState } from 'react'
```
jay-hodgson, line 1: Refactored.  Pulled code from ProgrammaticOptions, which was a modal used in the context of table download.
Then changed ProgrammaticOptions to use this ProgrammaticInstructionsModal, and then renamed ProgrammaticOptions to ProgrammaticTableDownload!

src/lib/containers/download_list_v2/DirectProgrammaticDownload.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from 'react'
```
jay-hodgson, line 1: Similar to DirectDownload, but this takes care of the button (code icon) and modal (uses ProgrammaticInstructionsModal)

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -98,7 +99,8 @@ export default function DownloadListTable(props: DownloadListTableProps) {
         'success',
         {title}
       )
-      // refetching the statistics will update the download list, so no need to update the file list here.
+      // refetching the statistics will update the download list, but refresh the list immediately since this will take time
+      refetch()
```
jay-hodgson, line 103: I noticed that it takes a second for the change to propagate back down to this component.  Let's refresh the list immediately, and we'll end up doing an additional background refetch after the stats change.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -299,19 +301,12 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
-                          {/* <span className="programmaticAccessItem"
-                            data-for={`${item.fileEntityId}-programmatic-instructions-tooltip`}
-                            data-tip="Programmatic download options">
-                              <ReactTooltip
-                                delayShow={TOOLTIP_DELAY_SHOW}
-                                place="left"
-                                type="dark"
-                                effect="solid"
-                                id={`${item.fileEntityId}-programmatic-instructions-tooltip`}
+                          <span className="programmaticAccessItem">
```
jay-hodgson, line 304: New action in the Download Cart file table!

src/lib/containers/entity_finder/details/view/DetailsView.tsx
```diff
@@ -450,21 +449,10 @@ export const DetailsView: React.FunctionComponent<DetailsViewProps> = ({
                   return (
                     <>
                       Version
-                      <MarkdownPopover
```
jay-hodgson, line 453: Used the DetailsView as the basis for the initial implementation for HelpPopover

src/lib/containers/table/table-top/ProgrammaticTableDownload.tsx
```diff
@@ -0,0 +1,86 @@
...
+  queryResultBundle: QueryResultBundle
+  onHide: () => void
+}
+
+function ProgrammaticOptions({
+  queryBundleRequest,
+  queryResultBundle,
+  onHide,
+}: ProgrammaticOptionsProps) {
+  const [generatedSql, setGeneratedSql] = useState('')
```
jay-hodgson, line 22: Now this component only needs to be responsible for figuring out the code for each programmatic client.

src/lib/rollup.index.ts
```diff
@@ -73,14 +75,16 @@ const SynapseComponents = {
...
   SynapseToastContainer,
   displayToast,
   IconSvg,
   UserProfileLinks,
   PlotlyWrapper,
   DatasetItemsEditor,
   EntityTypeIcon,
+  HelpPopover,
+  ProgrammaticTableDownload,
+  DirectProgrammaticDownload,
```
jay-hodgson, line 87: Need to merge with esbuild changes.

src/lib/containers/ProgrammaticInstructionsModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+  rCode,
+  cliNotes,
+  cliCode,
+  helpMarkdown,
+  helpUrl,
+}: ProgrammaticInstructionsModalProps) => {
+  let defaultTab
+  if (cliCode) {
+    defaultTab = ProgrammaticOptionsTabs.COMMAND_LINE
+  } else if (rCode) {
```
nickgros, line 48:
```suggestion
  } else if (pythonCode) {
```

src/lib/containers/download_list_v2/DirectProgrammaticDownload.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function DirectProgrammaticDownload({
+  entityId,
+  version
+}: DirectProgrammaticDownloadProps) {
+  const [isShowingModal, setIsShowingModal] = useState<boolean>(false)
+
+  const rCode = `library(synapser)
+synLogin('username','password')
```
nickgros, line 19: Doesn't need to be resolved in this PR, but we should modify these instructions to support Personal Access Tokens instead of a password

rollup.config.js
```diff
@@ -13,7 +13,7 @@ import sass from 'sass'
 import analyze from 'rollup-plugin-analyzer'
 
 const extensions = ['.js', '.jsx', '.ts', '.tsx']
-const maxBytes = 5e6 // if > 5MB total, then error
+const maxBytes = 6e6 // if > 5MB total, then error
 const onAnalysis = ({ bundleSize }) => {
   if (bundleSize < maxBytes) return
   console.log(`Bundle size exceeds ${maxBytes} bytes: ${bundleSize} bytes`)
```
nickgros, line 19: I'll make a ticket to add this step to the new esbuild bundle. I'm wondering how the analysis gets this number, though, since after tree-shaking/minification we're typically under 2MB for JS and 0.5 MB for CSS

src/lib/containers/download_list_v2/DownloadListTable.tsx
nickgros: Could also try invalidating the cache key. Not sure if that will put you into a hard loading state, which is probably worse than waiting for stale data to update. I think what you have is probably the right thing to do.

src/lib/containers/download_list_v2/DirectProgrammaticDownload.tsx
jay-hodgson: Good point

jay-hodgson:
Demo: Screen Shot 2021-12-10 at 9 45 12 AM

src/lib/containers/ProgrammaticInstructionsModal.tsx
```diff
@@ -73,10 +73,8 @@ export const ProgrammaticInstructionsModal = ({
       size="lg"
     >
       <Modal.Header closeButton>
-        <Modal.Title>
-          <Typography variant={'headline1'}>{title}</Typography>
-          {helpMarkdown && <HelpPopover markdownText={helpMarkdown} helpUrl={helpUrl} />}
-        </Modal.Title>
```
nickgros: I think `Modal.Title` is important for getting the correct sizing and aligning the close button. Any particular reason you removed it?

nickgros (approved):

src/lib/containers/ProgrammaticInstructionsModal.tsx
jay-hodgson: Good point.  The Modal Title style was being applied to the HelpPopover (and child contents), meaning the popover help text was huge.  I'll try wrapping the HelpPopover text in body1 Typography instead

jay-hodgson:
Screen Shot 2021-12-10 at 10 50 15 AM

src/lib/containers/MarkdownPopover.tsx
```diff
@@ -71,7 +72,9 @@ export const MarkdownPopover: React.FunctionComponent<MarkdownPopoverProps> = ({
         ref={setPopperElement}
         {...attributes.popper}
       >
-        <MarkdownSynapse {...contentProps} />
+        <Typography variant="body1">
+          <MarkdownSynapse {...contentProps} />
+        </Typography>
```
nickgros, line 77: We'll probably want to do this the other way around in the future but this works for now!

PR #1334, opened 2021-12-10
SWC-5895 and SWC-5896

jay-hodgson:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synapse-react-client",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
nickgros, line 3:
```suggestion
  "version": "2.1.0",
```

I didn't release 2.1.0, so this is unnecessary (unless you did)

nickgros (approved):

PR #1341, opened 2021-12-14
SWC-5900

nickgros:
Changed prop name will cause breaking API change (there is a corresponding PR in SWC, portals will catch this at type-checking) but passing a string is still permitted.

### DIFF
--- a/src/lib/containers/FullWidthAlert.tsx
+++ b/src/lib/containers/FullWidthAlert.tsx
@@ -16,7 +16,7 @@ export interface FullWidthAlertProps extends AlertProps {
   primaryButtonText?: string
   onPrimaryButtonClick?: () => void
   secondaryButtonText?: string
-  secondaryButtonHref?: string
+  onSecondaryButtonClick?: (() => void) | string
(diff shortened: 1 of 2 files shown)


src/lib/containers/FullWidthAlert.tsx
```diff
@@ -94,12 +94,24 @@ function FullWidthAlert(props: FullWidthAlertProps) {
...
         </span>
-        {secondaryButtonText && secondaryButtonHref && (
+        {secondaryButtonText && onSecondaryButtonClick && (
           <a
             className="secondaryButton"
             target="_blank"
             rel="noopener noreferrer"
-            href={secondaryButtonHref}
+            onClick={
+              typeof onSecondaryButtonClick === 'function'
```
jay-hodgson, line 103: Looking at this prop name, I would assume that it's always a function.  Maybe a comment on the prop to inform the caller that they can pass either a function or a URL string that will be used for the href?

jay-hodgson (approved):

jay-hodgson (approved):

PR #1342, opened 2021-12-14
PORTALS-1928: get the user preview image

jay-hodgson:


nickgros (approved):
Looks great. I didn't even realize we had profile picture previews! I'm assuming this will reduce the amount of data the client downloads, in addition to the much more simple request code?

PR #1345, opened 2022-01-05
SWC-5907: workaround for direct download on Safari

jay-hodgson:


jay-hodgson:
Chrome (and FF) UX remain unchanged.  Note that this is what >90% of the users will experience:

https://user-images.githubusercontent.com/1864447/148146652-ebce4502-0a14-4637-8488-49454e3b3d95.mov

This fix allows functionality for Safari users (<10% of the user base), but it's less than ideal:

https://user-images.githubusercontent.com/1864447/148146783-956672ca-2ddf-4e7a-a76a-2452cb40265e.mov

jay-hodgson:
Other things I tried can be found in the associated ticket, https://sagebionetworks.jira.com/browse/SWC-5907

src/lib/containers/DirectDownload.tsx
```diff
@@ -52,27 +52,43 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = props =
...
-      if (associatedObjectType === FileHandleAssociateType.TableEntity) {
-        const files = await getTableEntityFileHandle(true)
-        preSignedURL = files.requestedFiles[0].preSignedURL!
-      } else {
-        const file = await getFileResult(fileEntity!, accessToken, false, true)
-        preSignedURL = file.preSignedURL
+      let preSignedURL
+      // SWC-5907: opening in the file must be strictly done in the same click event process (Safari only).
+      // https://stackoverflow.com/questions/6628949/window-open-popup-getting-blocked-during-click-event
+      const isSafari = navigator.userAgent.indexOf("Safari") != -1 && navigator.userAgent.indexOf("Chrome") == -1
```
nickgros, line 58: I figure user agents are pretty stable but I wonder if we would be better served using an dependency here. Or if this should be pulled out and moved to utils

src/lib/containers/DirectDownload.tsx
```diff
@@ -52,27 +52,43 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = props =
...
-      if (onClickCallback) {
-        onClickCallback()
+      if (!preSignedURL) {
+        safariDownloadWindow?.close()
+        console.log('Fail to get file download link')
+      } else {
+        if (isSafari && safariDownloadWindow) {
+          safariDownloadWindow.location = preSignedURL
+          setTimeout(() => {
+            safariDownloadWindow!.close()
```
nickgros, line 83: I think you'll want a null check here for `safariDownloadWindow` in case the download window is closed between when this timer starts and 10s elapses.

src/lib/containers/DirectDownload.tsx
```diff
@@ -52,27 +52,43 @@ const DirectDownload: React.FunctionComponent<DirectFileDownloadProps> = props =
...
-      console.log('Fail to get file download link')
-    } else {
-      window.open(preSignedURL)
-      if (onClickCallback) {
-        onClickCallback()
+      if (!preSignedURL) {
+        safariDownloadWindow?.close()
+        console.log('Fail to get file download link')
+      } else {
+        if (isSafari && safariDownloadWindow) {
```
nickgros, line 80: I'm wondering if there could possibly be a race condition where we get here and the window hasn't opened yet, but probably not possible since JS is single-threaded.

nickgros (approved):
Left some comments, I definitely agree that this is the best approach given the options!

src/lib/containers/DirectDownload.tsx
jay-hodgson: That may be a good idea.  I'll look into it.

src/lib/containers/DirectDownload.tsx
jay-hodgson: You're right in that the creation (and URL load) of the Window are asynchronous.  But window.open() actually returns a WindowProxy that has all features immediately available.
https://developer.mozilla.org/en-US/docs/Web/API/Window/open

src/lib/containers/DirectDownload.tsx
jay-hodgson: Definitely

nickgros (approved):

PR #1346, opened 2022-01-06
PORTALS-2081: fix SynapseTable rerendering issue (resulting in scroll…

jay-hodgson:
… problem)

### DIFF
--- a/src/lib/containers/table/SynapseTable.tsx
+++ b/src/lib/containers/table/SynapseTable.tsx
@@ -1,4 +1,4 @@
-import { cloneDeep } from 'lodash-es'
+import { cloneDeep, eq } from 'lodash-es'
 import * as React from 'react'
 import { Button, Modal } from 'react-bootstrap'
 import { lexer } from 'sql-parser'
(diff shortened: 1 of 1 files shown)


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -168,15 +168,25 @@ export default class SynapseTable extends React.Component<
...
+    const shouldComponentUpdate = isPropsChange || isStateChange
+    if (shouldComponentUpdate) {
+      this.disableResize()
+    }
+    return shouldComponentUpdate
   }
   componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps) {
-    this.getEntityHeadersInData(false)
+    // PORTALS-2081: if the data changed, then get the new entity headers
+    if (!eq(prevProps.data, this.props.data)) {
```
jay-hodgson, line 187: This single line fixes PORTALS-2081, make sure the data has changed before getting the new entity headers

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -168,15 +168,25 @@ export default class SynapseTable extends React.Component<
   shouldComponentUpdate(
     nextProps: QueryWrapperChildProps & SynapseTableProps,
     nextState: Readonly<SynapseTableState>,
-    nextContext: any,
   ): boolean {
-    this.disableResize()
-    return super.shouldComponentUpdate
-      ? super.shouldComponentUpdate(nextProps, nextState, nextContext)
-      : true
+    // ignore isFetching state variables when checking for change in state
```
jay-hodgson, line 172: This code is just bonus, since I found that super.shouldComponentUpdate was always undefined (so it was always returning true).  Now I'm giving it a try.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -218,7 +228,12 @@ export default class SynapseTable extends React.Component<
         this.resizer = new ColumnResizer(this.tableElement, RESIZER_OPTIONS)
       }
     } else {
-      this.resizer.reset(RESIZER_OPTIONS)
+      // TODO: use event driven solution instead of timeout.
```
jay-hodgson, line 231: Initially the column resizer grip positions were... off.  I think it's due to the SynapseTableCell refactor (which I totally support!), but I'm sure there's a better way than this (so I left a note)

nickgros (approved):

PR #1353, opened 2022-01-19
Changes to support SWC-5872

nickgros:
- Change `FullWidthAlert`/`displayToast` signatures
	- Add disable, tooltip options
	- Now uses a new `AlertButtonConfig` type
- Add "tertiary" button style that matches our `<a>` tags stylistically

https://user-images.githubusercontent.com/17580037/150217703-6868275e-1fdf-4cc0-a235-bad1bc40a4dc.mov

Before/After using border-bottom vs text-decoration-offset:

Before:

https://user-images.githubusercontent.com/17580037/150364431-cb938f38-e5ac-4f31-a49f-7de1ea067710.mov

After:

https://user-images.githubusercontent.com/17580037/150364436-3aacc998-a09d-4932-aa8d-776fce99339c.mov

### DIFF
--- a/src/lib/containers/FullWidthAlert.tsx
+++ b/src/lib/containers/FullWidthAlert.tsx
@@ -8,15 +8,31 @@ import {
   Clear,
 } from '@material-ui/icons'
 import Typography from '../utils/typography/Typography'
+import ReactTooltip from 'react-tooltip'
+import { rebuildTooltip } from '../utils/functions/TooltipUtils'
+
+export type AlertButtonConfig = {
+  text: string
+  isDisabled?: boolean
+  tooltipText?: string
+} & (
+  | // "onClick" or "href", but not both
+  {
+      onClick?:
+        | ((e?: React.MouseEvent<HTMLElement, MouseEvent>) => void)
+        | (() => void)
+    }
+  | { href?: string }
+)
+
+const FULL_WIDTH_ALERT_TOOLTIP_ID = 'FullWidthAlertTooltip'
 
 export interface FullWidthAlertProps extends AlertProps {
   variant: string
   title?: string
   description?: string
-  primaryButtonText?: string
-  onPrimaryButtonClick?: () => void
-  secondaryButtonText?: string
-  onSecondaryButtonClickOrHref?: (() => void) | string
+  primaryButtonConfig?: AlertButtonConfig
+  secondaryButtonConfig?: AlertButtonConfig
   onClose?: () => void
   autoCloseAfterDelayInSeconds?: number
   isGlobal?: boolean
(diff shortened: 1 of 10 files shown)


src/lib/containers/FullWidthAlert.tsx
```diff
@@ -37,6 +53,41 @@ function getIcon(variant?: string) {
...
 
+function ButtonFromConfig(props: {
+  config?: AlertButtonConfig
+  className: string
+  variant: string
+}) {
+  const { config, variant, className } = props
+  if (config) {
+    return (
+      <span // See https://github.com/wwayne/react-tooltip/issues/304
```
nickgros, line 64: Have to wrap the button in a span to show the tooltip when the button is disabled.

src/lib/style/bootstrap4_backports/_base-import.scss
```diff
@@ -204,4 +204,28 @@
...
+      $active-background: rgba(255, 255, 255, 0),
+      $active-border: map.get(SRC.$primary-color-palette, 700)
+    );
+    border: none;
+    font-weight: 900;
+    white-space: normal; // Since this appears more like a link, line breaks are ok
+    &:hover {
+      box-shadow: none;
+      text-decoration: underline;
+      text-underline-offset: 4px;
```
nickgros, line 223: @jay-hodgson 

Using an underline with `text-underline-offset` instead of `border-bottom` (which our anchor style uses). Not sure if you looked into this before. The trade-off is that this doesn't work in IE but does properly handle text-wrapping.

Should the anchor styling use this, or should this button use `border-bottom`?

jay-hodgson: I didn't try using text-underline-offset, but this is an acceptable trade-off. I think this is a question for design then (on wrap, underline the group of text, or a line under all characters?).
If we change the solution, please be sure to verify other cases (like tabs UI!).

jay-hodgson (approved):

src/lib/style/abstracts/_mixins.scss
```diff
@@ -142,13 +142,17 @@ $link-underline-color-default: adjust-color(
...
   letter-spacing: 0.5px;
-  border-bottom: 1px solid $link-underline-color-default;
+  text-decoration: underline;
+  text-underline-offset: 4px;
+  text-decoration-thickness: 1px;
+  text-decoration-color: $link-underline-color-default;
   &:hover {
-    border-bottom: 2px solid map.get(SRC.$primary-color-palette, 700);
-    margin-bottom: -1px;
+    text-decoration: underline;
```
jay-hodgson, line 152: Looks like this block is repeated, could you refactor and share (so we don't update one instance later)?

jay-hodgson (approved):

PR #1355, opened 2022-01-25
SWC-5929

nickgros:
[SWC-5929](https://sagebionetworks.jira.com/browse/SWC-5929)

- Update toast on adding items to Dataset
- Add toast on removing items to Dataset
- Only show the default onSave toast when the onSave prop isn't specified

The ticket also describes a new toast shown onSave. The toast includes a button to edit the Dataset Schema, which can't be done in SRC. That toast will be added in SWC.

### DIFF
--- a/src/lib/containers/table/datasets/DatasetItemsEditor.tsx
+++ b/src/lib/containers/table/datasets/DatasetItemsEditor.tsx
@@ -56,6 +56,8 @@ export type DatasetItemsEditorTableData = DatasetItem & {
 const ROW_HEIGHT = 42
 const TABLE_HEIGHT = 350
 
+const SAVE_THE_DATASET_TO_CONTINUE = 'Save the Dataset to continue.'
+
 export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
   const { entityId, onSave, onClose } = props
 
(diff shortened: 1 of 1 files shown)


src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -104,13 +106,15 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
 
   const mutation = useUpdateEntity<Dataset>({
     onSuccess: () => {
-      displayToast(
-        'Create a Version of this Dataset to freeze it in its current state',
-        'success',
-        { title: 'Dataset Saved' },
-      )
       if (onSave) {
         onSave()
```
nickgros, line 110: We will specify the new onSave toast in SWC, which will be shown when the `onSave` callback is invoked.

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -143,22 +147,55 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
     }
   })
 
-  function addItemsToDataset(newItems: Reference[]) {
+  function addItemsToDataset(itemsToAdd: Reference[]) {
```
nickgros, line 150: Just renaming the parameter, `itemsToAdd` is a little more accurate. The items are not necessarily new.

src/lib/containers/FullWidthAlert.tsx
```diff
@@ -59,7 +59,7 @@ function ButtonFromConfig(props: {
   variant: string
 }) {
   const { config, variant, className } = props
-  if (config) {
+  if (config && ('onClick' in config || 'href' in config)) {
```
nickgros, line 62: A button with no text will be created if the config is defined but has neither `onClick` nor `href`.

Due to a mistake I made, SWC does this a lot on staging (the config object is always created), so we will probably want to patch staging's version of SRC with a build including this change.

jay-hodgson (approved):

jay-hodgson:
Logic looks sound.  This is a complex, but useful, toast message!

PR #1356, opened 2022-01-25
SWC-5935

nickgros:
[SWC-5935](https://sagebionetworks.jira.com/browse/SWC-5935)
- Fix a bug in FullWidthAlert

### DIFF
--- a/src/lib/containers/FullWidthAlert.tsx
+++ b/src/lib/containers/FullWidthAlert.tsx
@@ -59,7 +59,7 @@ function ButtonFromConfig(props: {
   variant: string
 }) {
   const { config, variant, className } = props
-  if (config) {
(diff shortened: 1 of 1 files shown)


src/lib/containers/FullWidthAlert.tsx
```diff
@@ -59,7 +59,7 @@ function ButtonFromConfig(props: {
   variant: string
 }) {
   const { config, variant, className } = props
-  if (config) {
+  if (config && ('onClick' in config || 'href' in config)) {
```
nickgros, line 62: In SWC (currently staging only), I made the mistake of always creating a configuration object, even if there's no onClick/href action. One way to fix this is to ensure at least one of those is present.

jay-hodgson (approved):

PR #1363, opened 2022-01-27
PORTALS-2093

jay-hodgson:
Disallow adding duplicate TextMatchesQueryFilter to the additionalFilters array.  Also, let's reset the search string text after adding the filter.

### DIFF
--- a/src/lib/containers/FullTextSearch.tsx
+++ b/src/lib/containers/FullTextSearch.tsx
@@ -60,10 +60,20 @@ export function FullTextSearch(props: FullTextSearchProps) {
           'org.sagebionetworks.repo.model.table.TextMatchesQueryFilter',
         searchExpression: searchText,
       }
+      // PORTALS-2093: does this additional filter already exist?
+      const found = additionalFilters.find(filter =>
+        filter.concreteType == textMatchesQueryFilter.concreteType &&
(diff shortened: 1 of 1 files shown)


nickgros (approved):

src/lib/containers/FullTextSearch.tsx
```diff
@@ -60,10 +60,20 @@ export function FullTextSearch(props: FullTextSearchProps) {
           'org.sagebionetworks.repo.model.table.TextMatchesQueryFilter',
         searchExpression: searchText,
       }
+      // PORTALS-2093: does this additional filter already exist?
+      const found = additionalFilters.find(filter =>
+        filter.concreteType == textMatchesQueryFilter.concreteType &&
+        (filter as TextMatchesQueryFilter).searchExpression == textMatchesQueryFilter.searchExpression
+      )
+      if (found) {
+        return
```
nickgros, line 69: So basically, ignore the event entirely, and don't modify/clear the input? That makes sense to me. Maybe worth providing some feedback like above using `setCustomValidity` with something like `Search term ${term} is already present`

jay-hodgson: Yes, ignore if already present.  Like the idea of adding additional feedback.   For now, since this is an edge case, going to merge in order to resolve the bug and can improve the UX later if prioritized.

PR #1365, opened 2022-01-28
PORTALS-2090: Align card metadata

jay-hodgson:


src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.md
```diff
@@ -2,40 +2,47 @@ Configured to show cards:
 ```jsx
 <QueryWrapperPlotNav
   rgbIndex={1}
-  name='Initiatives'
-  sql='SELECT * from syn24189696'
+  name='Tools'
```
jay-hodgson, line 5: Changed to look at NF GFF tool cards

src/lib/style/components/_cards.scss
```diff
@@ -214,6 +214,9 @@ $text-color-lighter: #898989;
   .SRC-cardMetadata {
     display: block;
     padding: $baseline-grid;
+    >* {
+      margin-left: 15px;
```
jay-hodgson, line 218: The relevant change.  Children of SRC-cardMetadata (currently a single table element) should have a margin-left of 15px so that it aligns with card info above

PR #1367, opened 2022-01-31
SWC-5943: work towards standardizing the base font size as 16px inste…

jay-hodgson:
…ad of 14px

### DIFF
--- a/src/lib/containers/download_list_v2/DownloadCartPage.tsx
+++ b/src/lib/containers/download_list_v2/DownloadCartPage.tsx
@@ -125,7 +125,7 @@ export const DownloadCartPage:React.FunctionComponent<DownloadCartPageProps> = (
                       You Have Files Which Require Actions Before Downloading
                     </Typography>
                   </div>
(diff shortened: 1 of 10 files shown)


src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -82,8 +82,10 @@ export const DownloadCartPage:React.FunctionComponent<DownloadCartPageProps> = (
               Clear Your Download Cart
             </a>
           </div>
-          <p className="description">You may find your added files in the tabs below. Any files which require actions before download can be found in the Access Actions Required tab,
```
nickgros, line 85: If `description` corresponds to a specific CSS class here, you may be able to delete it.

nickgros (approved):

src/lib/containers/download_list_v2/DownloadCartPage.tsx
jay-hodgson: Added `description` back!

PR #1371, opened 2022-02-03
PORTALS-2095

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -182,9 +182,9 @@ export default class SynapseTable extends React.Component<
     }
     return shouldComponentUpdate
   }
-  componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps) {
+  componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps, prevState: Readonly<SynapseTableState>) {
     // PORTALS-2081: if the data changed, then get the new entity headers
-    if (!eq(prevProps.data, this.props.data)) {
+    if (!eq(prevProps.data, this.props.data) || !eq(prevState.isEntityView, this.state.isEntityView)) {
```
jay-hodgson, line 187: Technically, I think this would be !XOR, so !(prevState.isEntityView ^ this.state.isEntityView), but I think using lodash eq is more readable.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -845,10 +847,9 @@ export default class SynapseTable extends React.Component<
       if (isShowingDownloadColumn) {
         // SWC-5790: If this is a FileEntity, the download icon should just go to entity page
         rowContent.unshift(
-          <td className="SRC_noBorderTop direct-download">
+          <td key={`direct-download-${rowSynapseId}`} className="SRC_noBorderTop direct-download">
```
jay-hodgson, line 850: This key definition was at the wrong level, causing a warning in the console.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -182,9 +182,9 @@ export default class SynapseTable extends React.Component<
     }
     return shouldComponentUpdate
   }
-  componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps) {
+  componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps, prevState: Readonly<SynapseTableState>) {
```
nickgros, line 185: Like the Readonly<T> here

src/lib/containers/table/SynapseTable.tsx
nickgros: There are so many questions that come to mind about how we could probably overhaul this so it's easy to follow. It seems weird to me that we have to check both the data and the type, but I'm guessing it has to do with the fact that isEntityView is determined in state asynchronously.

So this works for now!

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -38,6 +38,9 @@ type SynapseTableCellProps = {
   columnModels?: ColumnModel[]
   tableEntityId?: string
   rowData: string[]
+  isEntityView: boolean
+  rowId: number
+  rowVersionNumber: number
```
nickgros, line 43: These should be optional, right?

nickgros (approved):

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
nickgros: Actually, we'll always have those, but they may not be useful, so they need not be optional. Nvm!

src/lib/containers/table/SynapseTable.tsx
jay-hodgson: That's exactly right, it's because isEntityView is stored in state and determined asynchronously

PR #1372, opened 2022-02-03
SWC-5944:  do not remove external file from cart on click/download.

jay-hodgson:


src/__tests__/lib/containers/SynapseTable.test.tsx
```diff
@@ -618,6 +625,20 @@ describe('basic functionality', () => {
         )
         expect(tableCell.find(EntityLink)).toHaveLength(1)
       })
+      it('PORTALS-2095: renders an entity link for name column in EntityView', () => {
```
jay-hodgson, line 628: Apparently I didn't include this change in the PORTALS-2095 PR.

nickgros (approved):

PR #1380, opened 2022-02-08
PORTALS-2094

jay-hodgson:


jay-hodgson:
Changes based on this updated design:
Screen Shot 2022-02-08 at 2 46 10 PM

Screen Shot 2022-02-08 at 2 44 51 PM
Screen Shot 2022-02-08 at 2 45 04 PM

src/lib/assets/mui_components/ContentCopy.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
```
jay-hodgson, line 1: ContentCopy was not part of [@material-ui/icons](https://www.npmjs.com/package/@material-ui/icons), but it is in [@mui/icons-material](https://www.npmjs.com/package/@mui/icons-material)

jay-hodgson: Added as a custom icon, but maybe we should move to this other icon set?

src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -214,9 +214,13 @@ export const DownloadCartPage:React.FunctionComponent<DownloadCartPageProps> = (
...
                     </Typography>
                     <span>
-                      <a className="highlight-link disabled"
+                      <a className="highlight-link"
+                        rel="noreferrer"
                         data-for="downloadProgrammaticallyTooltipId"
-                        data-tip="This feature is coming soon.<br />You can still download individual<br />files programmatically.">
+                        data-tip="This feature is coming soon.<br />Click here to learn how to download<br />programmatically using other methods"
+                        target="_blank"
+                        href="https://help.synapse.org/docs/API-Clients.1985446128.html"
```
jay-hodgson, line 222: Dependent on https://sagebionetworks.jira.com/browse/DOCS-101 being fixed.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -137,13 +153,46 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+              place="right"
+              type="dark"
+              effect="solid"
+              id='copy-syn-ids-tooltip'
+            />
+            <button
+              data-testid="copySynIdsButton"
+              onClick={() => {
+                // trigger loading all pages of the download list table, and then copy all IDs to the clipboard
+                setCopyingAllSynapseIDs(true)
```
jay-hodgson, line 174: Step 1: User triggers copyingAllSynapseIDs=true

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -68,19 +70,33 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+      setCopyingAllSynapseIDs(false)
+    }
+
     if (
       status === 'success' &&
       !isFetchingNextPage &&
       hasNextPage &&
       fetchNextPage &&
-      inView
+      (inView || copyingAllSynapseIDs)
```
jay-hodgson, line 92: Step 2:  If copyingAllSynapseIDs=true, fetch all pages of data.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -68,19 +70,33 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
     ) {
       fetchNextPage()
+    } else if (!hasNextPage && copyingAllSynapseIDs) {
```
jay-hodgson, line 95: Step 3:  If we have all pages of data, and we're coyingAllSynapseIDs, then concat all of the Synapse IDs from the data and copy them to the clipboard.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -137,13 +153,46 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
+    )
+  }
   const availableFiltersArray: AvailableFilter[] = [
     undefined,
     'eligibleForPackaging',
     'ineligibleForPackaging',
   ]
   return (
     <>
+      <BlockingLoader show={copyingAllSynapseIDs} />
```
jay-hodgson, line 195: Show a full page blocking loader if we're gathering the data for the copy.

src/lib/assets/mui_components/ContentCopy.tsx
nickgros: The @mui/* packages are part of MUI v5. I'm not sure if the v5 icon pack is compatible with MUI v4 (which we're currently using as @material-ui/*)

src/lib/containers/download_list_v2/DownloadCartPage.tsx
nickgros: While this placeholder link is not ideal, it's better than what we have currently.

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -187,6 +236,7 @@ export default function DownloadListTable(props: DownloadListTableProps) {
                 </th>
                 <th>
                   SynID
+                  <span>{showInteractiveCopyIdsIcon()}</span>
                   <span>{showInteractiveSortIcon('synId')}</span>
```
nickgros, line 240: Just a style note, I would prefer these to be JSX elements rather than invoked functions

src/lib/containers/download_list_v2/DownloadListTable.tsx
```diff
@@ -68,19 +70,33 @@ export default function DownloadListTable(props: DownloadListTableProps) {
     }
   }, [isError, newError, handleError])
 
+  const allRows = data?.pages.flatMap(page => page.page) ?? []
+
   useEffect(() => {
+    const copyAllSynapseIDs = () => {
+      const synIDs = allRows.map((item: DownloadListItemResult)=>{
+        return item.fileEntityId
```
nickgros, line 78: I think it's crucial to also copy the version number (especially if the file isn't the latest version, but I think we should probably just always have it), or else that info will be lost when the user copies and pastes this list. The version number should likely also be shown in the table in the synId or its own column.

nickgros (changes requested):
Logic looks good, I think we just need the version numbers

src/__tests__/lib/containers/download_list_v2/DownloadListTableV2.test.tsx
```diff
@@ -88,16 +85,46 @@ describe('DownloadListTableV2 tests', () => {
...
+      };
+      Object.assign(navigator, {
+        clipboard: mockClipboard,
+      })
+
+      renderComponent()
+      
+      const copySynIDsButton = await screen.findByTestId('copySynIdsButton')
+      userEvent.click(copySynIDsButton)
+      await screen.findAllByText('file2.txt')
```
nickgros, line 125: Not clear to me what this line is checking. Is it testing that a second page was fetched? That the blocking loader is gone? Not sure it would actually work in the blocking loader case, RTL can probably always see that text if it's already there.

jay-hodgson: Unclear to me too, removed!

nickgros (approved):

PR #1381, opened 2022-02-09
PORTALS-2064: change layout of the header card

jay-hodgson:


jay-hodgson:
Normal card rendering:
Screen Shot 2022-02-09 at 10 20 19 AM
Same card rendered as a Header card:
Screen Shot 2022-02-09 at 10 21 09 AM

src/lib/containers/HeaderCard.tsx
```diff
@@ -68,49 +68,58 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
...
-                secondaryLabelLimit={secondaryLabelLimit}
-                values={values}
-              />
-            )}
+      <div className="container-fluid container-full-width">
+        <div className="row">
+          <div className="col-md-offset-1 col-md-10">
+            <div className="SRC-portalCardMain">
+              {icon}
+              <div style={{ width: '100%' }}>
```
nickgros, line 76: Reason for inline styles? I have no preference whether they are CSS or inline if it's not reusable (inline makes more sense to me, honestly), but we haven't been doing this in the past.

src/lib/containers/HeaderCard.tsx
```diff
@@ -68,49 +68,58 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
...
+                        <MarkdownSynapse markdown={description} />
+                      ) : (
+                        description
+                      )}
+                    </span>
+                  )}
+                </div>
+                <div
+                  style={{
+                    borderTop: '1px solid rgba(26, 28, 41, 0.2)',
```
nickgros, line 103: Even if intentionally using inline styles, not sure if color should be inline, but maybe this is a reasonable special case

src/lib/containers/HeaderCard.tsx
```diff
@@ -68,49 +68,58 @@ const HeaderCard: React.FunctionComponent<HeaderCardProps> = ({
...
+                <div className="SRC-cardContent" style={{ marginLeft: '15px' }}>
+                  <div className="SRC-type">{type}</div>
+                  <div>
+                    <h3 className="SRC-boldText" style={{ margin: 'none' }}>
+                      {href ? (
+                        <a target={target} href={href} className="highlight-link">
+                          {title}
+                        </a>
+                      ) : (
+                        <span> {title} </span>
```
nickgros, line 86: Should any of these use `<Typography>`?

jay-hodgson: Yes, did not update in this PR.

src/lib/containers/HeaderCard.tsx
jay-hodgson: Agreed that color should not be inline, did not update in this PR.

nickgros (approved):

PR #1394, opened 2022-02-16
SWC-5957: support case when facets are not defined, or we know that...

jay-hodgson:
the sql changes the resulting data schema (so the facets are no longer applicable and should not be shown)

### DIFF
--- a/src/lib/containers/QueryWrapper.tsx
+++ b/src/lib/containers/QueryWrapper.tsx
@@ -12,6 +12,7 @@ import {
 import { cloneDeep } from 'lodash-es'
 import { SynapseClientError } from '../utils/SynapseClient'
 import { DEFAULT_PAGE_SIZE } from '../utils/SynapseConstants'
+import { isFacetAvailableAndSupported } from '../utils/functions/sqlFunctions'
 
 /**
  * TODO: SWC-5612 - Replace token prop with SynapseContext.accessToken
@@ -80,6 +81,7 @@ export type QueryWrapperState = {
   isColumnSelected: string[]
   selectedRowIndices?: number[]
   error: SynapseClientError | undefined
+  isFaceted: boolean
 }
 
 /*
@@ -154,6 +156,7 @@ export default class QueryWrapper extends React.Component<
     this.updateParentState = this.updateParentState.bind(this)
     this.getInitQueryRequest = this.getInitQueryRequest.bind(this)
     const showFacetVisualization = props.defaultShowFacetVisualization ?? true
+
     this.state = {
       data: undefined,
       isLoading: true,
@@ -168,6 +171,7 @@ export default class QueryWrapper extends React.Component<
       isAllFilterSelectedForFacet: {},
       loadNowStarted: false,
       lastQueryRequest: cloneDeep(this.props.initQueryRequest!),
+      isFaceted: true,
       topLevelControlsState: {
         showColumnFilter: true,
         showFacetFilter: true,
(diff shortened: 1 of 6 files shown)


src/lib/containers/QueryWrapper.tsx
```diff
@@ -168,6 +171,7 @@ export default class QueryWrapper extends React.Component<
       isAllFilterSelectedForFacet: {},
       loadNowStarted: false,
       lastQueryRequest: cloneDeep(this.props.initQueryRequest!),
+      isFaceted: true,
```
jay-hodgson, line 174: components initially assume facets are available and sql results will align with facets

src/lib/containers/QueryWrapper.tsx
```diff
@@ -277,13 +281,21 @@ export default class QueryWrapper extends React.Component<
       this.updateParentState,
     )
       .then((data: QueryResultBundle) => {
+        const isFacetAvailable = isFacetAvailableAndSupported(clonedQueryRequest.query.sql, data.facets)
```
jay-hodgson, line 284: When props change and query is re-executed, we update the isFaceted state.
Also update the top level control state (ability to show facet filters and visualizations)

src/lib/containers/QueryWrapper.tsx
```diff
@@ -375,6 +387,7 @@ export default class QueryWrapper extends React.Component<
             }
           })
         }
+        const isFacetAvailable = isFacetAvailableAndSupported(this.props.initQueryRequest.query.sql, data.facets)
```
jay-hodgson, line 390: On initialization, respect the default showFacetFilter and showFacetVisualization value.  But if configured to show, then only show if facets are actually available.

src/lib/containers/QueryWrapper.tsx
```diff
@@ -458,7 +478,7 @@ export default class QueryWrapper extends React.Component<
     }
     const loadingCusrorClass = isLoading ? 'SRC-logo-cursor' : ''
     return (
-      <div className={`SRC-wrapper ${loadingCusrorClass}`}>
+      <div className={`SRC-wrapper ${loadingCusrorClass} ${isFaceted ? 'has-facets' : ''}`}>
```
jay-hodgson, line 481: Layout changes depending on if facet filters are being shown (reclaims facet filter panel section for the table)

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -131,14 +133,19 @@ const QueryWrapperPlotNav: React.FunctionComponent<QueryWrapperPlotNavProps> =
...
+                      hideFacetFilterControl={!isFaceted}
+                      hideVisualizationsControl={!isFaceted}
                     />
-                    <QueryFilter {...queryWrapperChildProps} {...props} />
-                    <QueryFilterToggleButton {...queryWrapperChildProps} />
+                    {isFaceted && <>
+                      <QueryFilter {...queryWrapperChildProps} {...props} />
+                      <QueryFilterToggleButton {...queryWrapperChildProps} />
+                      </>}
                     <FacetNav
```
jay-hodgson, line 143: This is a little strange that we still show the FacetNav, but it also contains the query count, and shows the conditions that have been applied.

jay-hodgson:
Realization: The facet values are still relevant to the sql when aggregate functions are used, but the counts are not.  I think I just want to hide the counts in this case

PR #1395, opened 2022-02-16
SWC-5957

jay-hodgson:


src/lib/containers/QueryWrapper.tsx
```diff
@@ -178,6 +182,7 @@ export default class QueryWrapper extends React.Component<
       },
       isColumnSelected: [],
       selectedRowIndices: [],
+      isFacetsAvailable: true,
```
jay-hodgson, line 185: components initially assume facets are available

src/lib/containers/QueryWrapper.tsx
```diff
@@ -277,13 +282,20 @@ export default class QueryWrapper extends React.Component<
       this.updateParentState,
     )
       .then((data: QueryResultBundle) => {
+        const isFaceted = isFacetAvailable(data.facets)
```
jay-hodgson, line 285: When props change and query is re-executed, we update the isFaceted state.
Also update the top level control state (ability to show facet filters and visualizations)

src/lib/containers/QueryWrapper.tsx
```diff
@@ -385,7 +398,14 @@ export default class QueryWrapper extends React.Component<
             data?.selectColumns
               ?.slice(0, this.props.visibleColumnCount ?? Infinity)
               .map(el => el.name) ?? [],
+          isFacetsAvailable: isFaceted,
```
jay-hodgson, line 401: On initialization, respect the default showFacetFilter and showFacetVisualization value. But if configured to show, then only show if facets are actually available.

src/lib/containers/QueryWrapper.tsx
```diff
@@ -447,6 +467,7 @@ export default class QueryWrapper extends React.Component<
       topLevelControlsState: this.state.topLevelControlsState,
       isColumnSelected: this.state.isColumnSelected,
       selectedRowIndices: this.state.selectedRowIndices,
+      isFacetsAvailable: this.state.isFacetsAvailable,
```
jay-hodgson, line 470: isFacetsAvailable state is passed down to QueryWrapper children (for convenience)

src/lib/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -4,14 +4,22 @@
   @media (min-width: map-get(SRC.$breakpoints, 'medium')) {
     .SRC-wrapper {
       display: grid;
-      grid-template-columns: 60px 10px 25% 10px 75%;
-      padding-right: 120px;
```
jay-hodgson, line 8: original layout only applies if there are facets (see .has-facets below).

jay-hodgson:
Simple config:
Screen Shot 2022-02-16 at 1 00 41 PM
When the table schema has a faceted column:
Screen Shot 2022-02-16 at 12 57 58 PM

And then the schema was modified to not facet any columns:
Screen Shot 2022-02-16 at 12 58 46 PM
Screen Shot 2022-02-16 at 1 00 00 PM

nickgros (approved):
This is a great start to cleaning this up and getting it ready to use in SWC

PR #1396, opened 2022-02-17
SWC-5955: Add a way for the user to directly edit the underlying sql

jay-hodgson:
Includes a few bug fixes that I discovered along the way.


src/lib/assets/themed_icons/SqlEditorIcon.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react'
```
jay-hodgson, line 1: Old way of adding an icon (to align with all of the other table toolbar icons)

src/lib/containers/QueryWrapper.tsx
```diff
@@ -262,6 +263,7 @@ export default class QueryWrapper extends React.Component<
       isLoading: true,
       lastQueryRequest: clonedQueryRequest,
       selectedRowIndices: [], // reset selected row indices any time the query is re-run
+      error: undefined,
```
jay-hodgson, line 266: The error was not being reset when the query was re-executed.  We may not have known this because errors were lost in some cases!

src/lib/containers/QueryWrapper.tsx
```diff
@@ -301,7 +303,9 @@ export default class QueryWrapper extends React.Component<
       })
       .catch(error => {
         console.error('Failed to get data ', error)
-        this.setState(error)
+        this.setState({
+          error,
```
jay-hodgson, line 307: Specifically, this error was not shown in the UI (this is the case where the query was re-executed and encountered an error in the service call).

src/lib/containers/SqlEditor.tsx
```diff
@@ -0,0 +1,84 @@
...
+      setSql(defaultSql)
+      inputRef.current?.focus()
+    }
+  }, [showSqlEditor, getLastQueryRequest])
+
+  const search = (event: React.SyntheticEvent<HTMLFormElement>) => {
+    event.preventDefault()
+    const { executeQueryRequest, getLastQueryRequest } = props
+    const lastQueryRequestDeepClone = getLastQueryRequest!()
+    lastQueryRequestDeepClone.query.sql = sql
```
jay-hodgson, line 43: Replace the sql with whatever is in the textfield.

src/lib/containers/SqlEditor.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export function SqlEditor(props: SqlEditorProps) {
+  const [sql, setSql] = useState('')
+  const inputRef = useRef<HTMLInputElement>(null)
+  const showSqlEditor = props.topLevelControlsState?.showSqlEditor
+  const showFacetFilter = props.topLevelControlsState?.showFacetFilter
+  const {getLastQueryRequest} = props
+  useEffect(() => {
+    if (showSqlEditor) {
```
jay-hodgson, line 32: On show, reset to the sql that was run to generate the table (and facet values)

jay-hodgson: And gain focus

src/lib/containers/SqlEditor.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  const handleChange = (event: React.FormEvent<HTMLInputElement>) => {
+    inputRef.current?.setCustomValidity('')
+    setSql(event.currentTarget.value)
+  }
+
+  return (
+    <div
+      className={`SearchV2 ${
```
jay-hodgson, line 55: SQLEditor adopts same styling as the search boxes.

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.md
```diff
@@ -73,5 +73,6 @@ Configured to show a table:
     sqlOperator='='
     sql='SELECT name, assay, study, dataType, tissue, consortium FROM syn11346063.14'
     downloadCartPageUrl='#/Other%20Components?id=downloadcartpage'
+    hideSqlEditorControl={false}
```
jay-hodgson, line 76: hideSqlEditorControl defaults to true for now, so you would need to explicitly set hideSqlEditorControl to false in order to see the new icon and access the SqlEditor.

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -107,8 +110,18 @@ const QueryWrapperPlotNav: React.FunctionComponent<QueryWrapperPlotNavProps> =
             >
               {queryWrapperChildProps => {
                 const isFaceted = queryWrapperChildProps.isFacetsAvailable
+                const showFacetFilter = queryWrapperChildProps.topLevelControlsState?.showFacetFilter
                 return (
                   <>
+                    <div
```
jay-hodgson, line 116: When the error was shown (on initial execution of the query), it may be really difficult to find since it was below the long list of facets on the left!  This moves the error banner to directly below the top level controls (which is done via css grid, but I also moved it in the DOM to make it easier to find!). Also needs to respect the facet filter collapse state

jay-hodgson:
New control (not yet approved by design, but they will need to review all of this in Experimental Mode when we replace the Synapse.org Table/View renderer):
Screen Shot 2022-02-16 at 4 39 39 PM

The SQL editor:
Screen Shot 2022-02-16 at 4 39 48 PM

An error on query execution (previously not shown in the UI):
Screen Shot 2022-02-16 at 4 40 02 PM

Components all respect the collapse state of the filter panel:
Screen Shot 2022-02-16 at 4 40 12 PM

And it works, you can edit the sql to modify the results!
Screen Shot 2022-02-16 at 4 42 28 PM

src/lib/containers/QueryWrapper.tsx
nickgros: Good catch!

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
nickgros: So now it should also be obvious if you try to submit malformed SQL. Great!

src/lib/containers/SqlEditor.tsx
nickgros: Suggest making the CSS class name more generic, like `QueryWrapperTextInput`, so it's clear that the class isn't tightly coupled with the SearchV2 component

nickgros:
I would want to see a test at the QueryWrapper level that 
* Finds the SqlEditor button
* Enters text into the editor
* Verifies that the entered query is submitted

But I see that we only have Enzyme tests, and if we're planning on a major refactor to function components soon, it doesn't make sense to me to keep using Enzyme. 

So my suggestion is either set up @testing-library/react and write this test, or just write a stub test with a description and create a ticket for testing backfill and put it in the story--since this isn't going to be instantly available, and the setup for this component is very complex, I'm ok with putting it off for a bit if we can come back to it before putting it in SWC.

nickgros (approved):
Looks great, just noted a suggested CSS class name change and want some action to make sure this is tested at some point, but those tests don't necessarily have to be in this PR.

src/lib/containers/SqlEditor.tsx
jay-hodgson: Great idea, will do

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
jay-hodgson: Exactly

PR #1400, opened 2022-02-23
Fix bundle for QueryWrapperPlotNav

nickgros:
Collection of fixes to export QueryWrapperPlotNav

### DIFF
--- a/src/lib/containers/TotalQueryResults.tsx
+++ b/src/lib/containers/TotalQueryResults.tsx
@@ -2,7 +2,7 @@ import { cloneDeep } from 'lodash-es'
 import React, { FunctionComponent, useState } from 'react'
 import { Button } from 'react-bootstrap'
 import useDeepCompareEffect from 'use-deep-compare-effect'
-import { SynapseClient, SynapseConstants } from '../'
+import { SynapseClient, SynapseConstants } from '../utils'
 import {
   getStoredEntityHeaders,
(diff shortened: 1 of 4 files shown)


src/lib/containers/TotalQueryResults.tsx
```diff
@@ -2,7 +2,7 @@ import { cloneDeep } from 'lodash-es'
 import React, { FunctionComponent, useState } from 'react'
 import { Button } from 'react-bootstrap'
 import useDeepCompareEffect from 'use-deep-compare-effect'
-import { SynapseClient, SynapseConstants } from '../'
+import { SynapseClient, SynapseConstants } from '../utils'
```
nickgros, line 5: The previous import resolves to `src/lib/index.ts` which pulls in modules and SCSS for the webpack build that we don't want in the UMD build

src/lib/containers/table/SearchResultsNotFound.tsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react'
-import NoSearchResults from '../../assets/icons/NoSearchResults.svg'
+import { ReactComponent as NoSearchResults } from '../../assets/icons/NoSearchResults.svg'
```
nickgros, line 2: Need to import SVGs in a way our ESBuild + plugins can handle

PR #1401, opened 2022-02-23
SWC-5956

jay-hodgson:


src/lib/containers/QueryWrapper.tsx
```diff
@@ -37,6 +37,8 @@ export type QueryWrapperProps = {
   showBarChart?: boolean
   componentIndex?: number //used for deep linking
   shouldDeepLink?: boolean
+  onQueryChange?: (newQueryJson: string) => void
```
jay-hodgson, line 40: If deep linking, provide way to inform caller that the Query has been updated (so you can update the URL).

src/lib/containers/QueryWrapper.tsx
```diff
@@ -37,6 +37,8 @@ export type QueryWrapperProps = {
...
+  onQueryResultBundleChange?: (newQueryResultBundleJson: string) => void
```
jay-hodgson, line 41: Provide way for caller to listen for QueryResultBundle changes (SWC uses this extensively to figure out state of other UI widgets).

src/lib/containers/SqlEditor.tsx
```diff
@@ -41,7 +41,7 @@ export function SqlEditor(props: SqlEditorProps) {
     const { executeQueryRequest, getLastQueryRequest } = props
     const lastQueryRequestDeepClone = getLastQueryRequest!()
     lastQueryRequestDeepClone.query.sql = sql
-
+    lastQueryRequestDeepClone.query.offset = 0
```
jay-hodgson, line 44: When updating the sql manually, reset the offset.

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -2,7 +2,7 @@ import { cloneDeep } from 'lodash-es'
 import React, { FunctionComponent, useState } from 'react'
 import { Button } from 'react-bootstrap'
 import useDeepCompareEffect from 'use-deep-compare-effect'
-import { SynapseClient, SynapseConstants } from '../'
```
jay-hodgson, line 5: Was breaking the export of QueryWrapperPlotNav

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -28,6 +28,12 @@ type OwnProps = {
   sql: string
   limit?: number
   shouldDeepLink?: boolean
+  /** If onQueryChange is set, the callback will be invoked when the Query changes */
+  onQueryChange?: (newQueryJson: string) => void
+  /** If onQueryResultBundleChange is set, the callback will be invoked when the QueryResultBundle changes */
+  onQueryResultBundleChange?: (newQueryResultBundleJson: string) => void
+  /** If initQueryJson is set, it will be the Query used in the initial QueryBundleRequest */
+  initQueryJson?: string
```
jay-hodgson, line 36: Provide way to initialize the Query (set by SWC based to either the default Query, or it uses the encoded Query from the URL)

src/lib/containers/table/SearchResultsNotFound.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react'
-import NoSearchResults from '../../assets/icons/NoSearchResults.svg'
+import { ReactComponent as NoSearchResults } from '../../assets/icons/NoSearchResults.svg'
```
jay-hodgson, line 2: Also broke QueryWrapperPlotNav export

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -184,9 +185,13 @@ export default class SynapseTable extends React.Component<
   }
   componentDidUpdate(prevProps: QueryWrapperChildProps & SynapseTableProps, prevState: Readonly<SynapseTableState>) {
     // PORTALS-2081: if the data changed, then get the new entity headers
-    if (!eq(prevProps.data, this.props.data) || !eq(prevState.isEntityView, this.state.isEntityView)) {
+    if (!eq(prevProps.data, this.props.data)) {
       this.getEntityHeadersInData(false)
+    } else if (!eq(prevState.isEntityView, this.state.isEntityView)) {
+      // if we determined that this is an entity view, force refresh the entity headers (for Views we need to get the rowIds!)
+      this.getEntityHeadersInData(true)
```
jay-hodgson, line 192: If we determined that this is an EntityView, then we need to get EntityHeaders for the row IDs as well, so force it!

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -523,10 +529,11 @@ export default class SynapseTable extends React.Component<
     const isShowingAccessColumn: boolean | undefined =
       showAccessColumn && this.state.isEntityView
     const isLoggedIn = !!this.context.accessToken
+    const containsGroupBy = isGroupBy(lastQueryRequest.query.sql)
```
jay-hodgson, line 532: Do not show the special download columns if the sql contains a "group by", as this changes what a row represents.

src/lib/containers/QueryWrapper.tsx
```diff
@@ -393,6 +403,9 @@ export default class QueryWrapper extends React.Component<
           })
         }
         const isFaceted = isFacetAvailable(data.facets)
+        if (this.props.onQueryResultBundleChange) {
+          this.props.onQueryResultBundleChange(JSON.stringify(data))
+        }
```
nickgros, line 408: This is quite the callback. Can't wait for us to have the chance to refactor this. Looks good

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
nickgros: I think I understand, this is a string because otherwise you would need a JSInterop compatible Query object, and at that point we're replicating models that we already get from the backend.

src/lib/containers/table/SynapseTable.tsx
nickgros: Should it be this?
```ts
     } else if (this.state.isEntityView && !eq(prevState.isEntityView, this.state.isEntityView)) {
```

Seems like if we go from View -> TableEntity we shouldn't do this, but maybe I don't fully understand this.

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -523,10 +529,11 @@ export default class SynapseTable extends React.Component<
...
     const isShowingAddToV2DownloadListColumn: boolean =
-      this.state.isFileView && !this.props.hideDownload && isLoggedIn
+      this.state.isFileView && !this.props.hideDownload && isLoggedIn && !containsGroupBy
```
nickgros, line 534: I don't think `isFileView` will work for Datasets as they don't have a viewTypeMask (yet)

nickgros (approved):
Looks great!

PR #1402, opened 2022-02-24
PORTALS-2109

yjyoo773:


src/lib/containers/ModalDownload.tsx
```diff
@@ -90,6 +90,7 @@ export default class ModalDownload extends React.Component<
       writeHeader,
       includeRowIdAndRowVersion,
       csvTableDescriptor: { separator },
+      additionalFilters: queryRequest.query.additionalFilters
```
nickgros, line 93: Typically I'd want to see a test that verifies that this parameter gets included but the existing ModalDownload tests are not in good shape, so no need right now.

nickgros (approved):

PR #1408, opened 2022-02-28
SWC-5994

nickgros:
Bump font size down on Help Icon and add left margin in one place


src/lib/style/base/_core.scss
```diff
@@ -1233,39 +1233,52 @@ hr ~ .SRC-wrapper {
 // Other links should have the normal secondary link style.
 // You can force normal link style by adding the "secondary-links" css class a container of your anchor.
 // Or force the highlighted primary link by adding "highlight-link" css class name to an anchor.
-main, #rootPanel {
```
nickgros, line 1236: All caused by Prettier. Only new change is at the bottom of this diff.

We should add autoformatting to a pre-commit task so this stops happening!

jay-hodgson (approved):

PR #1409, opened 2022-03-01
SWC-5809 - Add Jira issue collector to Error Boundary

nickgros:
If our ErrorBoundary catches a 500 error, we should show our Jira issue collector.

https://user-images.githubusercontent.com/17580037/156203837-248882a5-fd1f-4977-a56c-f52891a141b9.mov

### DIFF
--- a/src/__tests__/lib/containers/entity/metadata/EntityModal.test.tsx
+++ b/src/__tests__/lib/containers/entity/metadata/EntityModal.test.tsx
@@ -26,6 +26,20 @@ const defaultProps: EntityModalProps = {
   entityId: MOCK_FILE_ENTITY_ID,
 }
 
+jest.mock(
+  '../../../../../lib/containers/entity/annotations/SchemaDrivenAnnotationEditor',
+  () => ({
+    SchemaDrivenAnnotationEditor: jest.fn(() => <p>Mock Annotation Editor</p>),
+  }),
+)
+
+jest.mock(
+  '../../../../../lib/containers/entity/metadata/AnnotationsTable',
+  () => ({
+    AnnotationsTable: jest.fn(() => <p>Mock Annotations Table</p>),
+  }),
+)
+
 function renderComponent(
   propOverrides?: Partial<EntityModalProps>,
   wrapperProps?: SynapseContextType,
@@ -53,10 +67,7 @@ describe('EntityModal tests', () => {
     )[0]
     userEvent.click(annotationsTab)
 
-    await screen.findByText(
-      Object.keys(mockFileEntityBundle.annotations!.annotations)[0],
-    )
-    // tests for specific data are in AnnotationsTable.test.tsx
+    await screen.findByText('Mock Annotations Table')
   })
 
   it('Shows the edit button when the user has edit permissions', async () => {
@@ -117,6 +128,6 @@ describe('EntityModal tests', () => {
     userEvent.click(editButton)
 
     // Text that appears in the editor component if there's a schema
-    await screen.findByText('requires scientific annotations', { exact: false })
+    await screen.findByText('Mock Annotation Editor', { exact: false })
   })
 })
--- a/src/lib/containers/ErrorBanner.tsx
+++ b/src/lib/containers/ErrorBanner.tsx
@@ -9,6 +9,7 @@ import { isError, isSynapseClientError } from '../utils/ErrorUtils'
 import { SynapseClientError } from '../utils/SynapseClient'
 import { useSynapseContext } from '../utils/SynapseContext'
 import { Optional } from '../utils/types/Optional'
+import { useJiraIssueCollector } from './JiraIssueCollector'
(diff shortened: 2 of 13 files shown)


src/lib/containers/JiraIssueCollector.tsx
```diff
@@ -0,0 +1,164 @@
...
+  principalId: string
+  userDisplayName?: string
+  userEmailAddress?: string
+}
+
+/**
+ * Shows the Jira issue collector when invoked.
+ * @param config
+ */
+export function showJiraIssueCollector(config: JiraIssueCollectorConfig): void {
```
nickgros, line 42: Copied from SWC with minor changes

src/lib/containers/JiraIssueCollector.tsx
```diff
@@ -0,0 +1,164 @@
...
+    useGetCurrentUserProfile({
+      enabled: !!accessToken,
+    })
+  const { data: notificationEmail, isLoading: emailIsLoading } =
+    useGetNotificationEmail({
+      enabled: !!accessToken,
+    })
+
+  useEffect(() => {
+    if (show && accessToken) {
```
nickgros, line 129: Should we show the issue collector for anonymous users?

src/lib/utils/SynapseContext.tsx
```diff
@@ -33,19 +39,20 @@ export type SynapseContextProviderProps = {
...
+export const SynapseContextProvider: React.FunctionComponent<SynapseContextProviderProps> =
+  ({ children, synapseContext, queryClient }) => {
+    return (
+      <SynapseContext.Provider value={synapseContext}>
+        <QueryClientProvider client={queryClient ?? defaultQueryClient}>
+          {synapseContext?.withErrorBoundary ? (
+            <SynapseErrorBoundary>{children}</SynapseErrorBoundary>
+          ) : (
+            children
+          )}
```
nickgros, line 51: I think the main reason we haven't yet gotten much value out of error boundaries is because we treat SRC like a component library, and error boundaries solve problems for complete applications. Placement of error boundaries and the design of fallback components can be done more deliberately in a hierarchical app. Since we don't have many complex hierarchies, it often doesn't make sense to put error boundaries in SRC components, but it may make sense for these components to be surrounded within an error boundary.

As an alternative to adding an error boundary wrapper to every component, I've added the ability to add the error boundary where we provide the SynapseContext. In SWC, we wrap nearly every component in individual `SynapseContextProvider`s, so it seems valuable to place an error boundary here.

The error boundary is optional and not added by default because this makes no sense in React apps like Portals, where we want this context provider to be placed as high up in the component tree as possible. In that case, it makes no sense to wrap an entire app in an error boundary, since the whole app would visibly break any time an uncaught error is thrown.

src/lib/containers/ErrorBanner.tsx
```diff
@@ -23,6 +24,14 @@ export const ClientError = (props: { error: SynapseClientError }) => {
...
   const accessDenied = error.status === 403 && accessToken
 
+  useJiraIssueCollector({
+    show: error.status >= 500,
+    issueCollector: 'SWC',
+    issueSummary: '',
+    issueDescription: error.reason,
+    issuePriority: '3',
+  })
+
```
nickgros, line 34: When we get a client error with status >= 500, show the issue collector

src/lib/utils/hooks/SynapseAPI/useUserBundle.ts
```diff
@@ -3,14 +3,38 @@ import { SynapseClient } from '../..'
...
+  )
+}
 
 export function useGetUserProfile(
   principalId: string,
   options?: UseQueryOptions<UserProfile, SynapseClientError>,
 ) {
   const { accessToken } = useSynapseContext()
-  const queryKey = [accessToken, 'user', principalId, 'profile']
+  const queryKey = ['user', principalId, 'profile']
```
nickgros, line 37: In general, we should drop the `accessToken` from react-query keys. It makes debugging the cache much harder, and the cache should be cleared when the user logs out anyways.

src/__tests__/lib/containers/entity/metadata/EntityModal.test.tsx
```diff
@@ -26,6 +26,20 @@ const defaultProps: EntityModalProps = {
   entityId: MOCK_FILE_ENTITY_ID,
 }
 
+jest.mock(
```
nickgros, line 29: These tests were unstable on my machine, so just fixing them here with mocks

src/lib/containers/JiraIssueCollector.tsx
jay-hodgson: Well, I'd prefer to hear from Synapse users so we can follow up for more information (if necessary).  On the other hand, we'd probably rather get too many reports than too few.

jay-hodgson (approved):

PR #1413, opened 2022-03-02
SWC-6009

nickgros:
In the Entity Finder search results, we only show entity types that can be selected. This will change the text of the search button in the Entity Finder if only one type can be chosen.

For example, if only Docker Repos can be chosen, the search button will look like this:

image


package.json
```diff
@@ -35,7 +35,6 @@
     "column-resizer": "^1.3.5",
     "downshift": "^6.1.2",
     "gh-pages": "^2.0.0",
-    "i": "^0.3.6",
```
nickgros, line 38: This library does the same thing as pluralize but is less popular and I'm 99% sure we don't use it. Maybe it was accidentally installed?

jay-hodgson: Seems likely

jay-hodgson (approved):

PR #1417, opened 2022-03-02
PORTALS-2102

yjyoo773:


nickgros (approved):
All modals have a close button or an internal button that will hide the modal, so LGTM

PR #1420, opened 2022-03-04
SWC-6030

jay-hodgson:


src/lib/containers/QueryWrapper.tsx
```diff
@@ -296,9 +298,14 @@ export default class QueryWrapper extends React.Component<
         const hasMoreData =
           data.queryResult.queryResults.rows.length ===
           clonedQueryRequest.query.limit
+        const isColumnSelected = resetVisibleColumns ?
+            data?.selectColumns
+              ?.slice(0, this.props.visibleColumnCount ?? Infinity)
+              .map(el => el.name) ?? [] : this.state.isColumnSelected
```
nickgros, line 304: If we reset columns, then the new select columns becomes all of the columns up to the visible column count (if defined, otherwise all columns). If we don't have data (yet), there are no select columns. If we don't reset the columns, don't change the selected columns in state.

Did I get that right? It would be nice to simplify this, unless we're refactoring this soon.

nickgros (approved):
I can live with this and like the fix, but consider splitting the new complicated line into a few lines or separate static function

src/lib/containers/QueryWrapper.tsx
jay-hodgson: Yes, you got it right.  I'll pull this out to a shared static function and clean it up.

PR #1424, opened 2022-03-10
PORTALS-2116

jay-hodgson:


jay-hodgson:
- Change svg imports into material-ui SvgIcon components so that they can be used in the Sage Account web app.
- Add new Synapse Account registration service calls to SynapseClient.
- Add many of the types associated to these service calls.
- Fix detectSSOCode to support Google-based Synapse account creation, and binding an ORCiD identity to your Synapse account

src/lib/containers/Login.tsx
```diff
@@ -146,6 +148,7 @@ class Login extends React.Component<Props, State> {
       })
   }
   public render() {
+    const registerAccountUrl = this.props.registerAccountUrl ?? `${getEndpoint(BackendDestinationEnum.PORTAL_ENDPOINT,)}#!RegisterAccount:0`
```
conundrumgirl, line 151: is trailing comma a convention? I saw it in the code that this replaces. Just curious

src/lib/utils/synapseTypes/Principal/PrincipalServices.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// http://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/principal/AccountSetupInfo.html
+export type AccountSetupInfo = {
+  username: string
+  password: string
+  firstName?: string
+  lastName?: string
+  emailValidationSignedToken?: EmailValidationSignedToken
+}
+
```
conundrumgirl, line 43: I don't know if it makes sense, unless we add other similar type, but may be have a base type of 
```AccountBase = {
  firstName?: string
  lastName?: string
userName?: string
}`` and have AccountSetupInfo and NewUser extend it. Also, I think the capitalization of useName vs username should be consistent.

conundrumgirl (approved):
one small comment

src/lib/utils/synapseTypes/Principal/PrincipalServices.ts
jay-hodgson: Yes, but these are not my class type definitions, they are defined by the backend.
http://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/principal/AccountSetupInfo.html
http://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/auth/NewUser.html
While I would love for the capitalization to match, it does not match in the API, so <shrug>

src/lib/containers/Login.tsx
jay-hodgson: Not sure why that comma was there, but it was a copy/paste error (so I fixed it here, and at the source)

PR #1430, opened 2022-03-17
SWC-6050 - Handle MaterializedView more thoroughly in EntityTypeUtils.ts

nickgros:
* Handle MaterializedView in various utilities
* Add tests so we catch these cases next time we add an entity type

### DIFF
--- a/src/__tests__/lib/utils/functions/EntityTypeUtils.test.ts
+++ b/src/__tests__/lib/utils/functions/EntityTypeUtils.test.ts
@@ -1,0 +1,110 @@
+import {
+  convertToEntityType,
+  entityTypeToFriendlyName,
+  isContainerType,
+  isVersionableEntityType,
+} from '../../../../lib/utils/functions/EntityTypeUtils'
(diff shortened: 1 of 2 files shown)


src/__tests__/lib/utils/functions/EntityTypeUtils.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    type: EntityType.MATERIALIZED_VIEW,
+    concreteType: 'org.sagebionetworks.repo.model.table.MaterializedView',
+    friendlyName: 'Materialized View',
+  },
+]
+
+/**
+ * The purpose of these tests is more to ensure that we're handling all entity types as new ones are added.
+ * A key assumption is that that the EntityType enum is the "source of truth", so these tests will check that we handle all of the entity types in our test list.
+ */
```
nickgros, line 65: IOW, I really only care that we will see tests fail when we add a new entity type.

jay-hodgson (approved):

PR #1449, opened 2022-04-05
PORTALS-2152

nickgros:
Escape backticks, which won't work in command line.


src/lib/containers/table/table-top/ProgrammaticTableDownload.tsx
```diff
@@ -45,40 +45,70 @@ function ProgrammaticOptions({
...
       }
     }
     getData()
   }, [queryBundleRequest, queryResultBundle])
 
+  // Replace quotation marks with escaped quotations. For CLI, also escape backticks.
+  const commandLineSql = useMemo(
+    () => generatedSql.replace(/(["`])/g, '\\$1'),
+    [generatedSql],
+  )
```
nickgros, line 63: This is the main logic change. For command line instructions, also escape backticks.

conundrumgirl (approved):

PR #1460, opened 2022-04-12
PORTALS-2135

yjyoo773:


jay-hodgson (approved):

src/lib/utils/SynapseClient.ts
```diff
@@ -3306,3 +3306,31 @@ export const changePasswordWithCurrentPassword = (
...
+    '/auth/v1/user/changePassword',
+    newPassword,
+    undefined,
+    undefined,
+    BackendDestinationEnum.REPO_ENDPOINT
+  )
+}
+
+// http://rest-docs.synapse.org/rest/POST/user/password/reset.html
+export const resetPassword = (
```
jay-hodgson, line 3324: Could you change so the callback endpoint is also a param.  Then Our Sage Account web app would set this to the passwordReset route.  Since it's a page reload after the round trip, we may want to just use a URL parameter instead of the hash, but I'm ok with either.

PR #1466, opened 2022-04-14
PORTALS-2120

jay-hodgson:


jay-hodgson:
UserIdList in Synapse table cell:
Screen Shot 2022-04-13 at 4 03 05 PM

UserIdList in facet filters:
Screen Shot 2022-04-13 at 4 03 32 PM

Link type column:
Screen Shot 2022-04-13 at 4 03 12 PM

Friendly names used in "chicklets":
Screen Shot 2022-04-13 at 4 03 52 PM

Friendly names used in plots:
Screen Shot 2022-04-13 at 4 04 01 PM

Evaluation Queue name is renderered:
Screen Shot 2022-04-13 at 4 18 04 PM

src/__tests__/lib/containers/EvaluationIdRenderer.test.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react'
```
jay-hodgson, line 1: Add tests for the 2 new renderers

src/lib/containers/CardContainer.tsx
```diff
@@ -108,7 +108,7 @@ export const CardContainer = (props: CardContainerProps) => {
     // Hard coding ownerId as a column name containing the user profile ownerId
     // for each row, grab the column with the ownerId
     const userIdColumnIndex = data.queryResult.queryResults.headers.findIndex(
-      el => el.columnType === 'USERID',
+      el => el.columnType === ColumnType.USERID,
```
jay-hodgson, line 111: Use enum

src/lib/containers/EvaluationIdRenderer.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useEffect, useState } from 'react'
```
jay-hodgson, line 1: When in view, given an evaluation Id render the evaluation queue name.

src/lib/containers/UserIdList.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useEffect, useState } from 'react'
```
jay-hodgson, line 1: When in view, given an array of user profile IDs, render small user cards

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -218,11 +227,20 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
         }
       }
       break
+    case ColumnType.LINK:
```
jay-hodgson, line 230: And autolink the column value if the column type is LINK

src/lib/containers/EvaluationIdRenderer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { ref, inView } = useInView()
+  let mounted: boolean = true
+
+  useEffect(() => {
+    if (mounted && inView) {
+      getEvaluationQueue()
+    }
+    return () => {
+      mounted = false
+    }
```
nickgros, line 23: I keep seeing this pattern, but does it actually do anything? The return function only runs when the component unmounts, but is it even possible for the effect to run after unmount?

If we need to cancel the request if an unmount happens before the request resolves, then I think we would need to send a signal via AbortController

src/lib/containers/UserIdList.tsx
```diff
@@ -0,0 +1,47 @@
...
+          key={el.ownerId}
+          userProfile={el}
+          className='SRC-marginFifteen'
+        />)
+        setProfileCards(list)
+      })
+      .catch(e => {
+        console.log('UserIdList: Error getting user profiles', e)
+      })
+  }
```
nickgros, line 42: This works, but `UserCard` has its own logic to fetch profiles so you could just do this. Up to you.

```suggestion
    const list = userIds.map(id => <UserCard
          key={el.ownerId}
          ownerId={id}
          size={'SMALL USER CARD'}
          className='SRC-marginFifteen'
        />)
    setProfileCards(list)
```

src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -22,8 +22,8 @@ library.add(faArrowLeft)
 export type EnumFacetFilterProps = {
   facetValues: FacetColumnResultValueCount[]
   columnModel: SelectColumn
-  onChange: Function
-  onClear: Function
+  onChange: (facetNamesMap: Record<string, string>)=>void
+  onClear: ()=>void
```
nickgros, line 26: 👍

nickgros (approved):
Question about a pattern and an alternative suggestion for user cards, but in general looks good

PR #1467, opened 2022-04-15
SWC-6075

nickgros:
Don't link to disaggregated query by default


src/lib/containers/QueryWrapper.tsx
```diff
@@ -101,16 +101,6 @@ export type QueryWrapperProps = {
   lockedFacet?: LockedFacet
 }
 
-export type TopLevelControlsState = {
```
nickgros, line 104: Defined in more than one place

src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -82,173 +82,170 @@ export type QueryWrapperPlotNavProps = SearchParams &
   Operator &
   OwnProps
 
-const QueryWrapperPlotNav: React.FunctionComponent<QueryWrapperPlotNavProps> =
-  props => {
```
nickgros, line 86: Just changed `props` to `(props: QueryWrapperPlotNavProps)` because that causes props to show up in Styleguidist

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -107,6 +108,11 @@ export type SynapseTableProps = {
   columnLinks?: LabelLinkConfig
   hideDownload?: boolean
   isRowSelectionVisible?: boolean
+  /**
+   * If true, columns created by the COUNT function will render with a link to Synapse that opens the disaggregated query results filtered by the conditions of that row.
+   * Note that this is very brittle and only supports one column at a time. See SWC-6075 for more information. Default false.
+   */
+  linkCountToDisaggregatedQuery?: boolean
```
nickgros, line 115: New prop, defaults to false

src/lib/utils/functions/sqlFunctions.ts
```diff
@@ -84,14 +84,17 @@ export const getWhereInsertIndex = (tokens: string[][]): number => {
...
+ * This will construct a sql query by adding the conditions in searchParams
+ * to the WHERE clause, preserving all other clauses.
+ * If the searchParams are not defined, this will simply return the given sql.
+ */
 export const insertConditionsFromSearchParams = (
   sql: string,
   searchParams?: KeyValue,
   operator: SQLOperator = 'LIKE',
 ) => {
+  // TODO: Replace SQL manipulation with QueryFilters. See PORTALS-2157
```
nickgros, line 97: This is the only other place where we're using sql-parser. I think it's how we get the locked facet from a DetailsPage. I think we could probably add a QueryFilter rather than manipulating SQL, but that investigation/work are tracked in a new ticket.

jay-hodgson (approved):

PR #1469, opened 2022-04-19
SWC-6093

jay-hodgson:


jay-hodgson:
Confirmed expected UI style with Duminda, merging

PR #1470, opened 2022-04-19
Replace access icons with new MUI icons

nickgros:
Tried to re-align the icons with circles and couldn't get it consistent in both SWC and portals. So let's just use the new icons.

image

image

### DIFF
--- a/src/lib/containers/HasAccess.tsx
+++ b/src/lib/containers/HasAccess.tsx
@@ -16,13 +16,7 @@
  */
 
 import { IconProp, library } from '@fortawesome/fontawesome-svg-core'
-import {
-  faCircle,
-  faDatabase,
-  faLink,
-  faUnlockAlt,
-  faLock,
-} from '@fortawesome/free-solid-svg-icons'
+import { faCircle, faDatabase, faLink } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import * as React from 'react'
 import ReactTooltip from 'react-tooltip'
@@ -50,8 +44,9 @@ import AccessRequirementList, {
 } from './access_requirement_list/AccessRequirementList'
 import { SRC_SIGN_IN_CLASS } from '../utils/SynapseConstants'
 import { SynapseContext } from '../utils/SynapseContext'
+import IconSvg from './IconSvg'
+import { ThemeContext } from '../utils/hooks/useTheme'
 
-library.add(faUnlockAlt)
 library.add(faDatabase)
 library.add(faCircle)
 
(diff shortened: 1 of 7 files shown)


src/lib/utils/hooks/useTheme.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useContext } from 'react'
```
nickgros, line 1: IconSvg plays nicely with styles in JS, but not CSS classes, so it's more natural to determine the color via a theme object, rather than supplying a class name.

Thoughts on this approach, @jay-hodgson @conundrumgirl @yjyoo773 ? Is this going to be a pain to keep in line with SCSS? Should we fully commit to one tech, if so, which one?

A caller (e.g. a portal) could simply wrap the app in 
```tsx
const myPortalTheme = {
	colors: {
		success: "green"
	}
}

<ThemeProvider theme={myPortalTheme>
	// the rest of the app
</ThemeProvider>
```

src/lib/containers/HasAccess.tsx
```diff
@@ -321,13 +316,47 @@ export default class HasAccess extends React.Component<
     )
   }
 
+  renderOpenAccessIcon() {
+    return (
+      <ThemeContext.Consumer>
```
nickgros, line 321: In a function component, we could just do

```tsx
const theme = useTheme()

return <IconSvg
              options={{
                icon: 'accessOpen',
                color: theme.colors.success,
              }}
            />
```

So this won't typically be this much code

src/lib/utils/hooks/useTheme.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export type ThemeProviderProps = {
+  theme?: ThemeContextType
+}
+
+export const ThemeProvider: React.FunctionComponent<ThemeProviderProps> = ({
+  theme = defaultTheme,
+  children,
+}) => {
+  return <ThemeContext.Provider value={theme}>{children}</ThemeContext.Provider>
```
nickgros, line 38: If we want to do this, we would probably want to merge the provided partial theme object with a default theme, rather than completely replace.

src/lib/utils/hooks/useTheme.tsx
conundrumgirl: This is how material-ui is doing it. They provide a number of ways of customization including styled components and regular css classes. It appears to play well together.

src/lib/utils/hooks/useTheme.tsx
conundrumgirl: I agree. We should do the deep merge to make sure we get all the props

conundrumgirl (approved):

src/lib/containers/HasAccess.tsx
conundrumgirl: The reason for not wrapping the entire app in theme context is to avoid conflicts?

src/lib/containers/HasAccess.tsx
```diff
@@ -321,13 +316,47 @@ export default class HasAccess extends React.Component<
...
+                color: theme.colors.success,
+              }}
+            />
+          )
+        }}
+      </ThemeContext.Consumer>
+    )
+  }
+
+  renderLockedIcon() {
```
conundrumgirl, line 336: would it make sense to create more generic component that takes color type ('success', 'warning' etc) and icon?

nickgros: At the very least, I can definitely combine the two new methods that I wrote. It's a little weirder to combine with the existing code since we're using FontAwesome for the other icons

conundrumgirl: Yes - just the new ones.

src/lib/utils/hooks/useTheme.tsx
jay-hodgson: I support this direction if others are on board, since it's a good experience when using MUI (and it seems natural with our use of context and hooks).
Would it be possible to customize each portal via modifying the theme (I think they're scss variable based at the moment)?

jay-hodgson (approved):

src/lib/utils/hooks/useTheme.tsx
nickgros: @jay-hodgson yep! We would just need to export the ThemeContext.Provider or add the ThemeContext to our existing SynapseContext component, and then a portal could supply their own theme overrides.

PR #1473, opened 2022-04-25
SWC-6049: make file size optional in renderer code

jay-hodgson:
do not supply the byte count if greaterThan is true from the table/view stats response (we do not want to show this minimum to the user)


jay-hodgson:
Under 100 files, should work the same as it used to:
Screen Shot 2022-04-25 at 11 11 58 AM
But when greaterThan is true (>100 files), only show the file count:
Screen Shot 2022-04-25 at 11 12 23 AM

src/lib/containers/download_list/DownloadDetails.tsx
```diff
@@ -51,28 +52,33 @@ export default function DownloadDetails(props: DownloadDetailsProps) {
   }, [accessToken])
 
   const timeEstimateInSeconds =
-    isLoading || downloadSpeed === 0 ? 0 : numBytes / downloadSpeed
+    isLoading || downloadSpeed === 0 || !numBytes ? 0 : numBytes / downloadSpeed
+  const isTimeEstimateLoading = timeEstimateInSeconds === 0
```
jay-hodgson, line 56: Tried to maintain the same logic that was already in this component ( they seem to use timeEstimateInSeconds=0 to indicate that it should be inactive)

src/lib/containers/download_list/DownloadDetails.tsx
```diff
@@ -51,28 +52,33 @@ export default function DownloadDetails(props: DownloadDetailsProps) {
...
-        {isInactive ? (
+        <FontAwesomeIcon className={fileCountIconClass} icon="file" />
+        {isZeroFiles ? (
           <SkeletonInlineBlock width={50} />
         ) : (
           <> {numFiles}&nbsp;files </>
         )}
       </span>
-      <span
+      {numBytes && <span
```
jay-hodgson, line 78: only show the number of bytes if set (and greater than 0, since 0 is also falsy)

src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -248,9 +247,10 @@ export const DownloadConfirmation: React.FunctionComponent<DownloadConfirmationP
 
     useDeepCompareEffect(() => {
       if (queryResultBundle && downloadListStatistics) {
+        const isGreaterThanReportedValue = queryResultBundle.sumFileSizes?.greaterThan
```
jay-hodgson, line 250: If greaterThan is true, then use an undefined file size sum (indeterminate)

src/lib/containers/download_list/DownloadDetails.tsx
```diff
@@ -31,13 +31,14 @@ type State = {
  * @returns
  */
 export default function DownloadDetails(props: DownloadDetailsProps) {
+  const { numFiles, numBytes } = props
+  const { accessToken } = useSynapseContext()
+
   const [state, setState] = useState<State>({
-    isLoading: true,
+    isLoading: !!numBytes, // loading=true (to figure out the estimated download time) iff we were given a byte count
```
nickgros, line 38: Just dropping old code from the comment, but keeping the useful part
```suggestion
    isLoading: !!numBytes, // figure out the estimated download time iff we were given a byte count
```

jay-hodgson:
Included message change for https://sagebionetworks.jira.com/browse/SWC-6082
Screen Shot 2022-04-25 at 1 41 23 PM

src/lib/containers/download_list/DownloadDetails.tsx
nickgros: Can be hard to follow, but it at least keeps this change simple

nickgros (approved):
LGTM!

PR #1478, opened 2022-04-26
PORTALS-2163: show search bar by default if FTS is enabled

jay-hodgson:


src/lib/containers/FullTextSearch.tsx
```diff
@@ -25,12 +25,6 @@ export function FullTextSearch() {
   const [searchText, setSearchText] = useState('')
   const searchInputRef = useRef<HTMLInputElement>(null)
 
-  useEffect(() => {
-    if (showSearchBar) {
-      searchInputRef.current?.focus()
```
jay-hodgson, line 30: Since this is now shown by default, don't auto-focus

.github/workflows/build-test.yml
```diff
@@ -1,31 +1,31 @@
...
+#     steps:
+#     - uses: actions/checkout@v2
+#     - name: Use Node.js ${{ matrix.node-version }}
+#       uses: actions/setup-node@v2
+#       with:
+#         node-version: ${{ matrix.node-version }}
+#         cache: 'yarn'
+#     - run: yarn install --frozen-lockfile
+#     - run: yarn test
+#     - run: yarn build
```
jay-hodgson, line 31: Using travis - these failed checks cause too much noise

PR #1479, opened 2022-04-27
PORTALS-2159: use EntityLink instead of just collecting the entity names

jay-hodgson:


jay-hodgson:
Screen Shot 2022-04-26 at 4 58 57 PM

nickgros (approved):

PR #1484, opened 2022-04-27
SWC-6088 - QueryWrapper: Capture async job status, correctly capture errors

nickgros:
Fixes [SWC-6088](https://sagebionetworks.jira.com/browse/SWC-6088)


src/lib/containers/JiraIssueCollector.tsx
```diff
@@ -143,7 +143,7 @@ export function useJiraIssueCollector(config: UseJiraIssueCollectorConfig) {
       } else {
         console.warn('Fetching profile data to display Jira popup')
       }
-    } else {
+    } else if (show && !accessToken) {
```
nickgros, line 146: just noticed this was being logged if !show, i.e. not a 500 error. We should only log that we're not showing the Jira popup if (show && !accessToken)

src/lib/containers/QueryWrapper.tsx
```diff
@@ -340,7 +347,7 @@ export function QueryWrapper(props: QueryWrapperProps) {
     entity,
     executeQueryRequest,
     isFacetsAvailable,
-    asyncJobStatus: infiniteData?.pages[0],
+    asyncJobStatus: currentAsyncStatus,
```
nickgros, line 350: This didn't work because `infiniteData?.pages[0]` was undefined until the request for the first page completed. The "PROCESSING" state was never reported.

src/lib/utils/SynapseClient.ts
```diff
@@ -496,15 +499,15 @@ export const getAsyncResultFromJobId = async <TRequest, TResponse>(
...
+    setCurrentAsyncStatus?.(response)
   }
 
   if (response.jobState === 'FAILED') {
     /**
      * While we technically already have the failure reason in the response, the HTTP response doesn't give a helpful error code (e.g. 403)
      * that we can use for an error banner. We can get the HTTP code if we fetch the response body directly.
      */
-    doGet<TResponse>(
+    await doGet<TResponse>(
```
nickgros, line 510: This `await` is critical for capturing the error; the failed response was returned before this request would throw an error.

src/lib/utils/SynapseClient.ts
```diff
@@ -3338,19 +3347,19 @@ export const changePasswordWithToken = (
...
-export const resetPassword = (
-  email: string,
-) => {
-  const endpoint = window.location.href  + '?passwordResetToken='
-  const url = `/auth/v1/user/password/reset?passwordResetEndpoint=${encodeURIComponent(endpoint)}`
+export const resetPassword = (email: string) => {
+  const endpoint = window.location.href + '?passwordResetToken='
+  const url = `/auth/v1/user/password/reset?passwordResetEndpoint=${encodeURIComponent(
+    endpoint,
+  )}`
```
nickgros, line 3359: prettier

src/lib/containers/QueryWrapper.tsx
jay-hodgson: So you added a callback in the SynapseClient function, looks good to me

src/__tests__/lib/utils/SynapseClient.test.ts
```diff
@@ -0,0 +1,230 @@
...
+        ).rejects.toEqual(
+          expect.objectContaining({ status: errorStatus, ...errorObject }),
+        )
+        expect(setCurrentAsyncStatus).toBeCalledTimes(1)
+        expect(setCurrentAsyncStatus).toBeCalledWith(failedJob)
+      })
+    })
+  })
+
+  describe('SynapseClient unit tests', () => {
```
nickgros, line 173: These tests were untouched

src/__tests__/lib/utils/SynapseClient.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  AsynchronousJobStatus,
+  PaginatedResults,
+} from '../../../lib/utils/synapseTypes'
+import { SynapseClient } from '../../../lib/utils'
+import { FunctionReturningPaginatedResults } from '../../../lib/utils/SynapseClient'
+import { rest, server } from '../../../mocks/msw/server'
+import { ASYNCHRONOUS_JOB_TOKEN } from '../../../lib/utils/APIConstants'
+
+describe('SynapseClient tests', () => {
+  describe('SynapseClient integration tests', () => {
```
nickgros, line 15: Use MSW to have integration tests without relying on the service to be up/configured in a certain way

src/__tests__/lib/utils/SynapseClient.test.ts
```diff
@@ -0,0 +1,230 @@
+import {
```
nickgros, line 1: Changed file extension from .tsx to .ts because it wasn't using JSX/React

src/__tests__/lib/utils/SynapseClient.test.tsx
```diff
@@ -1,208 +0,0 @@
...
-import { BackendDestinationEnum } from '../../../lib/utils/functions/getEndpoint'
-import {
-  BatchFileRequest,
-  FileHandleAssociateType,
-  PaginatedResults,
-} from '../../../lib/utils/synapseTypes/'
-import { SynapseClient } from '../../../lib/utils/'
-import { FunctionReturningPaginatedResults } from '../../../lib/utils/SynapseClient'
-
-describe('it works at integration level testing', () => {
```
nickgros, line 11: Dropped all of these since they relied on 
* Synapse to be up (at least that's a pretty safe assumption 🙂)
* Synapse to be configured a certain way such that certain users and entities exist and are publicly viewable

I don't think we need a service test.

PR #1489, opened 2022-04-28
SWC-5875

nickgros:
Properly handle versions of entity metadata in the experimental mode modal, including disabling the annotations editor when inspecting an old version.

Doesn't yet show annotations of a specific version, tracked in [SWC-6105](https://sagebionetworks.jira.com/browse/SWC-6105)

e.g. draft version of dataset:
image
image

numbered version:
image
image


src/__tests__/lib/containers/entity/metadata/EntityModal.test.tsx
```diff
@@ -108,6 +108,28 @@ describe.skip('EntityModal tests', () => {
     ).not.toBeInTheDocument()
   })
 
+  it('Does not show the edit button for annotations when looking at an older version', async () => {
```
nickgros, line 111: All of these tests are skipped because they are unstable, but this was working locally.

src/lib/utils/functions/EntityTypeUtils.ts
```diff
@@ -137,15 +153,35 @@ export function isVersionableEntityType(type: EntityType): boolean {
...
+  Entity
+>(MATERIALIZED_VIEW_CONCRETE_TYPE_VALUE)
+
+export const isDataset = isTypeViaConcreteTypeFactory<Dataset, Entity>(
+  DATASET_CONCRETE_TYPE_VALUE,
+)
+
+export const isEntityView = isTypeViaConcreteTypeFactory<EntityView, Entity>(
+  ENTITY_VIEW_CONCRETE_TYPE_VALUE,
+)
```
nickgros, line 184: This makes me want to use classes

src/lib/utils/functions/TypeUtils.ts
```diff
@@ -13,13 +10,13 @@
...
->(expectedConcreteType: string) {
+  ObjectType extends { concreteType: string } = { concreteType: string },
+>(...expectedConcreteTypes: TTypeChecked['concreteType'][]) {
   return (object: ObjectType): object is TTypeChecked =>
     !!(
       object &&
       typeof object === 'object' &&
       'concreteType' in object &&
-      object['concreteType'] === expectedConcreteType
+      expectedConcreteTypes.includes(object['concreteType'])
```
nickgros, line 20: Supports more than one concreteType, so we can check interfaces.

src/lib/utils/synapseTypes/Entity/Entity.ts
```diff
@@ -56,9 +56,13 @@ export interface Versionable {
   versionNumber?: number
 }
 
-type VERSIONABLE_ENTITY_CONCRETE_TYPE =
-  | FILE_ENTITY_CONCRETE_TYPE
-  | TABLE_CONCRETE_TYPE
+export const VERSIONABLE_ENTITY_CONCRETE_TYPE_VALUES = [
+  FILE_ENTITY_CONCRETE_TYPE_VALUE,
+  ...TABLE_CONCRETE_TYPE_VALUES,
+] as const
```
nickgros, line 62: `as const` ensures that we get the specific string values when we call `typeof VALUES[number]`

src/lib/utils/functions/EntityTypeUtils.ts
```diff
@@ -162,3 +198,23 @@ export function hasFilesInView(entityView: EntityView) {
...
+  return isVersionableEntityType(convertToEntityType(entity.concreteType))
+}
+
+export function getVersionDisplay(entity: Entity): string {
+  if (!isVersionableEntity(entity)) {
+    console.warn("Entity isn't versionable:", entity)
+    return ''
+  }
+  if (!entity.isLatestVersion || !isTable(entity)) {
+    return entity.versionNumber!.toString()
```
jay-hodgson, line 214: Is this the expected behavior?  If this is the current version of a file, they want it to show the version number rather than "Current"?

nickgros: It looks like what we do today for files is `${versionNumber} (Current)`, so I'll update to match

src/lib/utils/functions/EntityTypeUtils.ts
jay-hodgson: Indeed

jay-hodgson (approved):

PR #1495, opened 2022-05-03
SWC-6008 - update dataset editor copy

nickgros:
change "Items" to "Files"


src/lib/containers/table/datasets/DatasetDisplayConstants.ts
```diff
@@ -0,0 +1,7 @@
+export const ADD_FILES = 'Add Files'
```
jay-hodgson, line 1: Not sure I want to adopt pulling out all display constants into their own file unless we need to (i18n language support).  Not against this idea, but if these were defined in DatasetItemsEditor.tsx that would be ok too

nickgros: Ok, I will move them back to DatasetItemsEditor.

PR #1496, opened 2022-05-03
SynapseContextProvider may have children

nickgros:


src/lib/utils/SynapseContext.tsx
```diff
@@ -34,6 +34,7 @@ export type SynapseContextProviderProps = {
   synapseContext?: SynapseContextType
   queryClient?: QueryClient
   theme?: SynapseTheme
+  children?: React.ReactElement
```
nickgros, line 37:
```suggestion
  children?: React.ReactNode
```

jay-hodgson (approved):

PR #1502, opened 2022-05-05
Rewrite Enzyme tests

nickgros:
Replacing Enzyme tests with RTL tests will
* Give us better confidence in testing behavior rather than implementation
* Unblock us from upgrading to >= React 17

This PR avoids touching actual components whenever possible. Any component changes typically facilitate good testing/accessibility practices, like adding appropriate ARIA roles to elements.

### DIFF
--- a/src/__tests__/lib/containers/TOC.test.tsx
+++ b/src/__tests__/lib/containers/TOC.test.tsx
@@ -1,9 +1,7 @@
+import { render, screen } from '@testing-library/react'
 import * as React from 'react'
-import { mount } from 'enzyme'
 import MarkdownSynapse from '../../../lib/containers/MarkdownSynapse'
-import { delay } from '../../../lib/utils/SynapseClient'
-import { _TIME_DELAY } from '../../../lib/utils/SynapseConstants'
-import { SynapseTestContext } from '../../../mocks/MockSynapseContext'
+import { createWrapper } from '../../../lib/testutils/TestingLibraryUtils'
 describe('renders without crashing', () => {
   let SynapseClient: any
   beforeAll(() => {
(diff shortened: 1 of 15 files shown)


src/lib/containers/widgets/RadioGroup.tsx
```diff
@@ -49,6 +49,9 @@ const RadioOption: React.FunctionComponent<RadioOptionProps> = (
       <input
         id={uniqueId}
         type="radio"
+        onChange={() => {
+          // no-op -- change is handled by the div
+        }}
```
nickgros, line 54: Silences warning about having an input field with `value` but no `onChange`

src/lib/template_style/_form.scss
```diff
@@ -103,8 +103,8 @@
   }
   input[type='checkbox']:focus + label::before,
   input[type='checkbox']:focus + span::before,
-  input[type='radio']:focus:checked + label::before,
-  input[type='radio']:focus:checked + span::before,
+  input[type='radio']:focus + label::before,
+  input[type='radio']:focus + span::before,
```
nickgros, line 107: Small style issue, radio should always have this drop shadow, not just when checked.

src/__tests__/lib/containers/TOC.test.tsx
```diff
@@ -19,32 +17,31 @@ describe('renders without crashing', () => {
...
+    render(<MarkdownSynapse ownerId={mockOwnerId} wikiId={mockWikiId} />, {
+      wrapper: createWrapper(),
+    })
+
+    // Render a link in the TOC that points to the corresponding heading element
+    const tocLink = await screen.findByRole<HTMLAnchorElement>('link')
+    const heading = await screen.findByRole<HTMLHeadingElement>('heading')
+    expect(tocLink).toHaveClass('link toc-indent1')
+
+    // TODO: Test that the link points to the header
```
nickgros, line 30: Not a new missing test, this was missing before. The widget uses some sophisticated scrolling library, so we can't just check the `href` attribute of the element.

jay-hodgson (approved):

PR #1503, opened 2022-05-06
Fix broken SVGs

nickgros:
For example, see https://nf.synapse.org/Explore/Datasets -- there's no `'dataset'` icon on the cards.

I was also having trouble getting our `yarn link` utilities to work to verify that the change would fix the issue in portals, so I made some changes that would facilitate using Yarn v3 in portals if we wanted to do so. Yarn v3 resolved the linking issue for me.


src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -140,15 +143,12 @@ const Icon: React.FunctionComponent<IconProps> = ({
   // or if the type of card has a corresponding icon, e.g. 'Publication'
   const Icon = iconSet[value] || iconSet[type]
   // TODO: get rid of dataset icon class, none of the icons should be special cased
-  const datasetIconClass =
-    value === DATASET || type === DATASET ? 'SRC-datasetIcon' : ''
-  if (typeof Icon == 'function') {
```
nickgros, line 145: At some point, `import { ReactComponent as xyz } from './file.svg'` ended up becoming an object rather than a function

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -140,15 +143,12 @@ const Icon: React.FunctionComponent<IconProps> = ({
...
-    return (
-      <span className={cssClass}>
-        <Icon />
-      </span>
-    )
-  }
-  return <img src={Icon} className={`iconImg ${datasetIconClass}`} />
+  const datasetCustomStyle =
+    value === DATASET || type === DATASET ? { height: '55px' } : {}
```
nickgros, line 147: AFAIK this datasets icon only appears in CardContainer, so I just tweaked this to make it look right.

package.json
```diff
@@ -33,6 +33,7 @@
     "@upsetjs/react": "^1.6.2",
     "animate.css": "^4.1.1",
     "bootstrap": "^4.6.1",
+    "classnames": "^2.3.1",
```
nickgros, line 36: We're using this despite it not being listed as a dependency. See Typography.tsx

package.json
```diff
@@ -97,6 +96,10 @@
     "universal-cookie": "^4.0.3",
     "use-deep-compare-effect": "^1.3.1"
   },
+  "peerDependencies": {
+    "react": "^16.14.0",
+    "react-dom": "^16.14.0"
+  },
```
nickgros, line 102: We don't actually want to supply copies of React and ReactDOM to dependents (like portals). They should be peerDependencies.

src/lib/containers/home_page/featured-data/FeaturedDataTabs.tsx
```diff
@@ -18,60 +18,65 @@ export type FeaturedDataTabsProps = {
...
-      )}
-    </div>
-  )
-}
+                <button
+                  className="SRC-centerAndJustifyContent"
+                  onClick={() => setSelectedTabIndex(index)}
+                >
+                  {config.icon && <Icon type={config.icon}></Icon>}
+                  <span>{config.title}</span>
```
nickgros, line 47: This is the only content change. Wrap the title in a span so we can make the button a flex container containing the icon and the title

src/lib/containers/home_page/featured-data/FeaturedDataTabs.tsx
```diff
@@ -18,60 +18,65 @@ export type FeaturedDataTabsProps = {
   sql: string
 }
 
-const FeaturedDataTabs: React.FunctionComponent<FeaturedDataTabsProps> = props => {
```
nickgros, line 21: Mostly whitespace changes, with a small alignment fix. Not sure if the regression occurred in the past or within my changes, but since the icons have been broken, the potential alignment issue wasn't visible.

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -140,15 +143,22 @@ const Icon: React.FunctionComponent<IconProps> = ({
...
-    )
+  const datasetCustomStyle =
+    value === DATASET || type === DATASET ? { height: '55px' } : {}
+
+  const className = `iconImg ${cssClass ?? ''}`
+
+  if (Icon == null) {
+    console.warn('Icon type not found:', type)
+    return <></>
+  }
```
nickgros, line 154: E.g. the Programs component might ask for an Icon we don't have. This prevents a crash

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -140,15 +143,22 @@ const Icon: React.FunctionComponent<IconProps> = ({
...
+  if (typeof Icon === 'string') {
+    return <img src={Icon} className={className} />
   }
-  return <img src={Icon} className={`iconImg ${datasetIconClass}`} />
+  return (
+    <span>
+      <Icon className={className} style={datasetCustomStyle} />
```
nickgros, line 160: React can call createElement on the item regardless of it's an object or a function. I would bet when our build tools changed, the previous dependency created function components, while the new one creates class components.

src/lib/style/main.scss
```diff
@@ -13,7 +13,7 @@
 
 // 2. Vendors
 @use 'vendors/normalize';
-@use 'animate.css/animate.css';
+@use 'animate.css/animate';
```
nickgros, line 16: For some reason, the old code didn't work with `sass-loader` (which is used by webpack). The compiled CSS shouldn't be affected by this.

In the portals repo, I've been looking into using `sass-loader` removing the CSS compilation step, so this would allow us to simplify things over there.

jay-hodgson (approved):

PR #1505, opened 2022-05-11
Add pre-commit code formatting

nickgros:
Code formatting is enforced on pre-commit. To accomplish this, I just ran Option 1 [here](https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged). I also ran `npx prettier --write .` to format all files after adding the precommit hook.

It's worth looking at [`0b67383` (#1505)](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1505/commits/0b6738319691439243e3d7cb03939638a98c011f) where I added the hook and https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1505/files/b47923e23f9acd60eabf16446024f4d35454c040..0f1d893221eea74f329b22c54e222de70210ef70 where I added .git-blame-ignore-revs and included more file extensions in the lint rule. [`b47923e` (#1505)](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1505/commits/b47923e23f9acd60eabf16446024f4d35454c040) only contains changes written by prettier.


nickgros:
> I am a bit confused with the purpose for .git-blame-ignore-revs just because it's something I haven't seen. I read through the stackoverflow link posted but still don't know what it's for.

@yjyoo773 Sometimes it can be helpful to look at the git blame to get context about a particular code fragment. For example I'll often find a commit where a code was changed, then check the PR it was included in to get a Jira ticket, and then the Jira ticket would hopefully give the explanation I'm looking for.

The file + configuration will tell git to ignore this change when looking at the blame, because code formatting changes don't provide valuable semantic context that you would typically want to find.

PR #1507, opened 2022-05-11
SWC-6101

yjyoo773:


src/lib/assets/icons/reply.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jay-hodgson, line 1: If an icon is available in the Material UI icon set, we should use it instead.  I think this is ReplyTwoTone?  Today we access these via the IconSvg component.

src/lib/containers/DiscussionSearchResult.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState, useEffect } from 'react'
```
jay-hodgson, line 1: Is this MessageTwoTone or ChatTwoTone?

src/lib/assets/icons/thread.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jay-hodgson, line 1: Is this MessageTwoTone or ChatTwoTone?

src/lib/containers/DiscussionSearchResult.tsx
```diff
@@ -0,0 +1,87 @@
...
+const DiscussionSearchResult = (props: DiscussionSearchResultProps) => {
+    const { threadId, replyId } = props
+    const { accessToken } = useSynapseContext()
+    const [threadBundle, setThreadBundle] = useState<DiscussionThreadBundle>()
+    const [messageUrl, setMessageUrl] = useState<string>('')
+    const [replyBundle, setReplyBundle] = useState<DiscussionReplyBundle>()
+    const [replyAuthor, setReplyAuthor] = useState<UserProfile>()
+
+    const getThreadOrReply = async () => {
+        let someUrl
```
jay-hodgson, line 40: someUrl renamed to messageUrl?

src/lib/containers/DiscussionSearchResult.tsx
```diff
@@ -0,0 +1,87 @@
...
+        }
+        setMessageUrl(await getMessage(someUrl.messageUrl))
+        setThreadBundle(thread)
+    }
+
+    useEffect(() => {
+        getThreadOrReply()
+    }, [])
+
+    const getUrl = (threadId: string, replyId: string, projectId: string) => {
```
jay-hodgson, line 59: I think replyId should be optional (and only set if the replyId is set in the search result), right?  If unset, it would be better to not include it in the URL query param in this URL.

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+        setSearchResult(searchResponse)
+        if (matchList) {
+            setMatchList([...matchList, ...searchResponse.matches])
+        }
+    }
+
+    return (
+        <div className="bootstrap-4-backport ForumSearch">
+            <div>
+                <FontAwesomeIcon
```
jay-hodgson, line 84: Use IconSvg (searchTwoTone) if possible

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+                        setSearchInput(event.target.value)
+                    }}
+                    onKeyDown={(event: any) => {
+                        if (event.key === 'Enter') {
+                            onSearch()
+                        }
+                    }}
+                />
+                {searchInput && (
+                    <FontAwesomeIcon
```
jay-hodgson, line 105: Use IconSvg (clear?) if possible

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+                isLoading ?
+                    <div style={{ textAlign: 'center' }}>
+                        <SynapseSpinner size={40} />
+                    </div> :
+                    matchList &&
+                    <>
+                        {props.projectId && (
+                            <Typography variant='body2'>Results for '{searchInput}' in {entity?.name}</Typography>
+                        )}
+                        {matchList.map(entity =>
```
jay-hodgson, line 130: rename entity -> match

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+                    />
+                )}
+            </div>
+            {
+                noSearchResult && <NoSearchResultComponent />
+            }
+            {
+                isLoading ?
+                    <div style={{ textAlign: 'center' }}>
+                        <SynapseSpinner size={40} />
```
jay-hodgson, line 123: Instead of a spinner, it would be better to have placeholder Skeleton results.  To do this, I would define and export a "LoadingDiscussionSearchResult" in DiscussionSearchResult, and show a few of those here instead.  See UserCardMedium.LoadingUserCardMedium for example.

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+                        {matchList.map(entity =>
+                            <div key={`${entity.forumId}-${entity.threadId}-${entity.replyId}`}>
+                                <DiscussionReply threadId={entity.threadId} replyId={entity.replyId} />
+                            </div>
+                        )}
+                    </>
+            }
+            {
+                searchResult?.nextPageToken &&
+                <div className='text-center'>
```
jay-hodgson, line 139: Not sure if centering is correct, but I think Button variant should be "primary"

src/lib/containers/ForumSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+import DiscussionReply from './DiscussionSearchResult'
+import { SynapseSpinner } from './LoadingScreen'
+import { Entity } from '../utils/synapseTypes'
+import Typography from '../utils/typography/Typography'
+import NoSearchResults from '../assets/icons/NoSearchResults'
+import { Button } from 'react-bootstrap'
+
+export type ForumSearchProps = {
+    forumId: string,
+    projectId?: string,
```
jay-hodgson, line 16: More comments throughout.
If projectId is set, will show that these search results are associated with the given project.

src/lib/containers/ForumSearch.tsx
yjyoo773: Should the button not be centered?

src/lib/containers/DiscussionSearchResult.tsx
```diff
@@ -35,51 +36,69 @@ const DiscussionSearchResult = (props: DiscussionSearchResultProps) => {
...
+        setIsLoading(false)
     }
 
     useEffect(() => {
         getThreadOrReply()
     }, [])
 
-    const getUrl = (threadId: string, replyId: string, projectId: string) => {
+    const getUrl = (threadId: string, projectId: string, replyId?: string,) => {
         return `https://synapse.org/#!Synapse:${projectId}/discussion/threadId=${threadId}&replyId=${replyId}`
```
jay-hodgson, line 64: Please only include the query param replyId if replyId is set (do not include replyId=undefined)

src/lib/containers/DiscussionSearchResult.tsx
```diff
@@ -35,51 +36,69 @@ const DiscussionSearchResult = (props: DiscussionSearchResultProps) => {
...
+                        </div>
+                    </Col>
+                </Row>
+            }
+        </div>
+    )
+}
+
+export const LoadingDiscussionSearchResult: React.FunctionComponent = () => {
+    return (
```
jay-hodgson, line 96: This should mirror the UI structure of a Discussion Search Result.  What does this look like today?

src/lib/containers/ForumSearch.tsx
jay-hodgson: I would double check other instances of paged results, but I don't think this is typically centered (but I might be wrong)

src/lib/containers/DiscussionSearchResult.tsx
jay-hodgson: Also, expected to see a list of these shown while search results are loading (instead of a spinner)

yjyoo773:
https://user-images.githubusercontent.com/60489495/168181775-9a82b008-3b27-46e3-a4b9-59ba09647774.mov

jay-hodgson (approved):

PR #1509, opened 2022-05-13
SWC-6122: Type refinement + API prep

nickgros:
Based on #1508


src/lib/utils/SynapseConstants.ts
```diff
@@ -7,6 +7,9 @@ export const AUTHENTICATED_PRINCIPAL_ID = 273948
 export const PUBLIC_PRINCIPAL_ID = 273949
 /** The single user representing all unauthenticated users */
 export const ANONYMOUS_PRINCIPAL_ID = 273950
+/** The Synapse Access and Compliance team */
+export const ACT_TEAM_ID = 464532
```
nickgros, line 11: Not sure if should hard-code, but this will let us link to the ACT team when they aren't in an ACL

jay-hodgson (approved):

PR #1516, opened 2022-05-18
PORTALS-2084: modify the current history entry (use replaceState) ins…

jay-hodgson:
…tead of creating a new one (pushState)


yjyoo773 (approved):
LGTM

PR #1520, opened 2022-05-18
SWC-5798, SWC-5799 - Entity Modal / Annotation Editor improvements

nickgros:
[SWC-5798](https://sagebionetworks.jira.com/browse/SWC-5798):
* Don't show "Open" button if that entity page is already open
 
[SWC-5799](https://sagebionetworks.jira.com/browse/SWC-5799):
* Move the "Edit" action to the right side of the modal by making it a secondary action, rather than tertiary action, in the FluidModal component
* Remove tertiaryActions since we aren't using it (and shouldn't use it per Synapse style guidelines)


src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -98,11 +96,7 @@ export const SchemaDrivenAnnotationEditor = (
 
   const { entityMetadata: entityJson, annotations } = useGetJson(entityId!, {
     enabled: !!entityId && !formData, // once we have data, don't refetch. it would overwrite the user's entries
-    onError: e => {
-      if (e.status !== 404) {
-        handleError(e)
-      }
```
nickgros, line 104: 404 here should error

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -116,13 +110,8 @@ export const SchemaDrivenAnnotationEditor = (
     entityId!,
     {
       enabled: !!entityId,
-      retry: false,
       refetchOnWindowFocus: false,
-      onError: e => {
-        if (e.status !== 404) {
-          handleError(e)
-        }
```
nickgros, line 124: 404 here should not error, added `allowNotFoundError` to the corresponding SynapseClient method.

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -136,11 +125,7 @@ export const SchemaDrivenAnnotationEditor = (
         delete schema.$id
         return schema
       },
-      onError: e => {
-        if (e.status !== 404) {
-          handleError(e)
-        }
```
nickgros, line 142: 404 should error here

src/lib/containers/entity/annotations/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -197,7 +182,7 @@ export const SchemaDrivenAnnotationEditor = (
               </b>
             </Alert>
           )}
-          {entityJson && (!formData || isEmpty(formData)) && !schema && (
+          {entityJson && (!formData || isEmpty(formData)) && schema === null && (
```
nickgros, line 185: `!schema` may be true if there is no schema, as it is undefined. `schema === null` is only true when we know that there is no schema.

src/lib/containers/entity/metadata/EntityModal.tsx
```diff
@@ -70,24 +70,25 @@ export const EntityModal: React.FC<EntityModalProps> = ({
...
-      copy: `Open ${entityTypeToFriendlyName(entityBundle.entityType!)}`,
-      onClick: () =>
-        window.open(
-          `${getEndpoint(
-            BackendDestinationEnum.PORTAL_ENDPOINT,
-          )}#!Synapse:${entityId}`,
-          '_blank',
-          'noopener',
-        ),
+    if (!window.location.href.includes(entityId)) {
```
nickgros, line 79: Render a button that links to the entity page in Synapse. I think for now, checking if the current URL including the synID is probably good enough.

src/lib/utils/SynapseClient.ts
```diff
@@ -3126,11 +3126,13 @@ export const updateSubmissionStatus = (
  */
 //
 export const getSchemaBinding = (entityId: string, accessToken?: string) => {
-  return doGet<JsonSchemaObjectBinding>(
-    ENTITY_SCHEMA_BINDING(entityId),
-    accessToken,
-    undefined,
-    BackendDestinationEnum.REPO_ENDPOINT,
+  return allowNotFoundError(() =>
```
nickgros, line 3129: If getSchemaBinding yields a 404, then there is no schema. Not an error.

jay-hodgson (approved):

PR #1521, opened 2022-05-19
SWC-6118

yjyoo773:


src/lib/containers/IconSvg.tsx
```diff
@@ -233,6 +234,8 @@ const getIcon = (options: IconSvgOptions) => {
       return <ReplyTwoTone style={muiSvgStyle}></ReplyTwoTone>
     case 'chatTwoTone':
       return <ChatTwoTone style={muiSvgStyle}></ChatTwoTone>
+    case 'accessManagement':
+      return <AccessManagement></AccessManagement>
```
jay-hodgson, line 238: Add support for fill color and customSvgStyle like many other custom svg icons in here?

src/lib/containers/SynapseNavDrawer.tsx
```diff
@@ -250,6 +256,14 @@ export const SynapseNavDrawer: React.FunctionComponent<
                   iconName: 'delete',
                   onClickGoToUrl: '/#!Trash:0',
                 })}
+              {isInSynapseExperimentalMode() &&
+                (currentUserBundle?.isACTMember ||
+                  currentUserBundle?.isARReviewer) &&
```
jay-hodgson, line 261: Notes for isARReviewer state that it is always true for an ACT member, so I don't think you need to ask for USER_BUNDLE_MASK_IS_ACT_MEMBER or reference it here.

PR #1522, opened 2022-05-20
SWC-6114: ability to set a static tooltip for a linked secondary column label

jay-hodgson:


jay-hodgson:
Screen Shot 2022-05-20 at 2 24 49 PM

src/lib/containers/GenericCard.tsx
```diff
@@ -240,39 +242,61 @@ export const SynapseCardLabel: React.FC<SynapseCardLabelProps> = props => {
         </>
       )
     } else {
-      return <MarkdownSynapse renderInline={true} markdown={value} />
+      labelContent = <MarkdownSynapse renderInline={true} markdown={value} />
```
jay-hodgson, line 245: This is a terrible diff for this file. 
Took existing logic, but instead of returning the JSX.Element I store it as labelContent.
Had to add "else" clauses (since function no longer returns in these cases).

src/lib/containers/GenericCard.tsx
```diff
@@ -284,27 +308,19 @@ export const SynapseCardLabel: React.FC<SynapseCardLabelProps> = props => {
...
-                {el}
-              </a>
-              {index < split.length - 1 && (
-                <span style={{ marginRight: 4 }}>, </span>
-              )}
-            </React.Fragment>
-          )
-        })}
-      </>
+  if (labelLink.tooltipText) {
```
jay-hodgson, line 313: And this is the new code.  If tooltipText is set, then wrap the labelContent in a ReactTooltip.  Otherwise, just return the labelContent

yjyoo773 (approved):
LGTM!

PR #1535, opened 2022-05-24
SWC-6123 - Access Requirement Search Box

nickgros:
Searching ARs:

https://user-images.githubusercontent.com/17580037/170099659-7f4719fb-5803-4c58-89cb-3e29bc844695.mov

Initial AR:

https://user-images.githubusercontent.com/17580037/170121232-1269dbfa-85b5-4492-b0e9-6820439bf032.mov

Enter AR ID:

https://user-images.githubusercontent.com/17580037/170121352-6fc39481-e7d0-4a19-ba32-a80b4763bd5c.mov

### DIFF
--- a/src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
+++ b/src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
@@ -27,7 +27,7 @@ describe('Submission Page tests', () => {
 
     // Configure MSW
     server.use(
-      // Return a mocked access requirement
+      // Return mocked access requirement search results
       rest.post(
         `${getEndpoint(
           BackendDestinationEnum.REPO_ENDPOINT,
--- a/src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
+++ b/src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
@@ -1,0 +1,100 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import React from 'react'
+import selectEvent from 'react-select-event'
+import AccessRequirementSearchBox from '../../../../lib/containers/dataaccess/AccessRequirementSearchBox'
+import { createWrapper } from '../../../../lib/testutils/TestingLibraryUtils'
+import { ACCESS_REQUIREMENT_SEARCH } from '../../../../lib/utils/APIConstants'
+import {
+  BackendDestinationEnum,
+  getEndpoint,
+} from '../../../../lib/utils/functions/getEndpoint'
+import { mockSearchResults } from '../../../../mocks/mockAccessRequirements'
+import { rest, server } from '../../../../mocks/msw/server'
+
+const mockOnChange = jest.fn()
+const onServiceRecievedRequest = jest.fn()
+
+function renderComponent() {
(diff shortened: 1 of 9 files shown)


src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+  })
+}
+
+describe('Submission Page tests', () => {
+  beforeAll(() => {
+    server.listen()
+
+    // Configure MSW
+    server.use(
+      // Return a mocked access requirement
```
nickgros, line 30:
```suggestion
      // Return mocked access requirement search results
```

src/lib/containers/dataaccess/AccessRequirementSearchBox.tsx
```diff
@@ -0,0 +1,41 @@
...
+      })) ?? []
+
+    return options
+  }
+
+  return (
+    <AsyncSelect
+      defaultOptions
+      cacheOptions
+      isClearable
```
nickgros, line 34: Truthy defaultOptions just results in the request firing before typing, which is fine. It will pre-populate the options.

Truthy cacheOptions will cache the results based on the input query. Since name is the only filter we're using for this component, the cache is useful + valid.

src/lib/containers/dataaccess/AccessRequirementSearchBox.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import AsyncSelect from 'react-select/async'
+import { SynapseClient } from '../../utils'
+import { useSynapseContext } from '../../utils/SynapseContext'
+
+export type AccessRequirementSearchBoxProps = {
+  onChange: (accessRequirementId?: string | undefined) => void
```
jay-hodgson, line 7: Expected the ability to set the initial AR ID as well (send ACT to a page that has this filter applied via the page token, it should be visible in the search box)

src/lib/containers/dataaccess/AccessRequirementSearchBox.tsx
jay-hodgson: Hopefully we do not need to support multiple pages of results for ARs

src/__tests__/lib/containers/dataaccess/AccessRequirementSearchBox.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+    expect(mockOnChange).not.toHaveBeenCalled()
+
+    // Select the first access requirement (the label is $NAME ($ID))
+    await selectEvent.select(input, 'Access Requirement 1 (123)')
+    // The prop should have fired upon selecting this option
+    await waitFor(() =>
+      expect(mockOnChange).toBeCalledWith(mockSearchResults.results[0].id),
+    )
+  })
+})
```
jay-hodgson, line 100: Nice set of tests and use of msw

src/lib/containers/dataaccess/AccessRequirementSearchBox.tsx
nickgros: Not for this component, at this time

src/lib/containers/dataaccess/AccessRequirementSearchBox.tsx
nickgros: Good catch

jay-hodgson (approved):

PR #1537, opened 2022-05-25
SWC-6125

yjyoo773:


yjyoo773:
Screen Shot 2022-05-25 at 11 31 33 AM
The Load More button would normally not show in this case but rendered to show the design and location of it.

src/lib/utils/SynapseClient.ts
```diff
@@ -3626,3 +3622,24 @@ export const forumSearch = (
...
+
+/**
+ * Search through the history of access approvals filtering by accessor/submitter
+ * and optional by access requirement id. The caller must be a member of the ACT.
+ * https://rest-docs.synapse.org/rest/POST/accessApproval/search.html
+ * @param accessApprovalSearchRequest
+ * @param accessToken
+ */
+
+export const accessApprovalSearch = (
```
nickgros, line 3634: Small nit, prefer name for functions to be a verb over a noun
```suggestion
export const searchAccessApprovals = (
```

src/lib/utils/synapseTypes/AccessApproval.ts
```diff
@@ -17,3 +41,15 @@ export interface AccessApproval {
...
+export interface AccessApprovalSearchRequest {
+  accessorId: string // 	Filter by the id of the principal that is an accessor in the approval. Note that the submitter is always part of the accessors. This field is required.
+  accessRequirementId?: string // Filter by the id of the access requirement of the approvals.
+  sort?: AccessApprovalSearchSort[] // Identifies both the field and direction for a single sort operation.
+  nextPageToken?: string // A token used to get the next page of a request.
+}
+
+export interface AccessApprovalSearchResponse {
+  results: AccessApprovalSearchResult[] // The matching Access Approval information corresponding to the search parameters
+  nextPageToken: string // A token used to get the next page of a particular search query.
```
nickgros, line 54: nextPageToken should likely be optional

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,213 @@
...
+    } catch (err: any) {
+      displayToast(err.reason as string, 'danger')
+    } finally {
+      setIsLoading(false)
+    }
+  }
+
+  useEffect(() => {
+    getApprovals()
+  }, [accessToken])
```
nickgros, line 73: This will only run when the access token changes, so new data won't be fetched e.g. if accessorId changes.

ESLint should catch this if you define the function within the effect.

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,213 @@
...
+  accessorId: string
+  accessRequirementId?: string
+}
+export const AccessApprovalsTable: React.FunctionComponent<
+  AccessApprovalsTableProps
+> = ({ accessorId, accessRequirementId }) => {
+  const { accessToken } = useSynapseContext()
+  const [searchResult, setSearchResult] =
+    useState<AccessApprovalSearchResponse>()
+  const [isLoading, setIsLoading] = useState(false)
```
nickgros, line 28: Not necessary, but I think we could remove a lot of these state variables and simplify some logic if we delegate fetching this data to a custom `react-query` hook

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,213 @@
...
+  type SortedColumn = {
+    column: string
+    isDescending: boolean
+  }
+
+  const [sortedColumn, setSortedColumn] = useState<SortedColumn>({
+    column: '',
+    isDescending: false,
+  })
+  const sortList = (
```
nickgros, line 84: We should rarely sort client-side, which can cause issues at scale and with UX (for example, what happens to your sort when you fetch more data?)

See the sort property in the request object:
https://docs.synapse.org/rest/org/sagebionetworks/repo/model/dataaccess/AccessApprovalSearchRequest.html

If an aspect of the design indicates sorting functionality that the API doesn't support, then we will need to resolve that by either not implementing that sort, or adding that functionality to the backend service

src/lib/utils/SynapseClient.ts
```diff
@@ -1274,7 +1272,7 @@ export const getWikiPageKeyForAccessRequirement = (
   accessToken: string | undefined,
   ownerId: string | number,
 ): Promise<WikiPageKey> => {
-  const url = ACCESS_REQUIREMENT_WIKI_PAGE_KEY(ownerId)
+  const url = `/repo/v1/access_requirement/${ownerId}/wikikey`
```
nickgros, line 1275: Revert

src/lib/utils/SynapseClient.ts
```diff
@@ -3126,13 +3124,11 @@ export const updateSubmissionStatus = (
...
-      ENTITY_SCHEMA_BINDING(entityId),
-      accessToken,
-      undefined,
-      BackendDestinationEnum.REPO_ENDPOINT,
-    ),
+  return doGet<JsonSchemaObjectBinding>(
+    ENTITY_SCHEMA_BINDING(entityId),
+    accessToken,
+    undefined,
+    BackendDestinationEnum.REPO_ENDPOINT,
```
nickgros, line 3131: Revert

nickgros (changes requested):
See comments

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -101,7 +101,9 @@ export const AccessApprovalsTable: React.FunctionComponent<
         accessToken,
       )
       setSearchResult(sortRequest)
-      setArList(sortRequest.results)
+      if (arList) {
+        setArList([...arList, ...sortRequest.results])
+      }
```
yjyoo773, line 106: This is for when sorting after clicking "Load More". Without it would only show sorted ARs from the nextPageToken.

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+            </th>
+            <th className="thead-cell remove-border">Submitter</th>
+            <th className="thead-cell remove-border">Status</th>
+            <th className="thead-cell remove-border button-cell">
+              Modified Date
+              <button
+                style={{ float: 'right' }}
+                onClick={() =>
+                  sortColumn(
+                    'MODIFIED_ON' as AccessApprovalSortField,
```
nickgros, line 132: Use the value instead of casting
```suggestion
                    AccessApprovalSortField.MODIFIED_ON,
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+                />
+              </button>
+            </th>
+            <th className="thead-cell remove-border button-cell">
+              Expires
+              <button
+                style={{ float: 'right' }}
+                onClick={() =>
+                  sortColumn(
+                    'EXPIRED_ON' as AccessApprovalSortField,
```
nickgros, line 150:
```suggestion
                    AccessApprovalSortField.EXPIRED_ON,
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+}
+export const AccessApprovalsTable: React.FunctionComponent<
+  AccessApprovalsTableProps
+> = ({ accessorId, accessRequirementId }) => {
+  const { accessToken } = useSynapseContext()
+  const [searchResult, setSearchResult] =
+    useState<AccessApprovalSearchResponse>()
+  const [isLoading, setIsLoading] = useState(false)
+  const [isDescending, setIsDescending] = useState(false)
+  const [arList, setArList] = useState<AccessApprovalSearchResult[]>()
```
nickgros, line 34: Not clear to me why this is stored in state when we also have `searchResult`

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+          })}
+        </tbody>
+      </Table>
+      {searchResult?.nextPageToken &&
+        (isLoading ? (
+          <SynapseSpinner size={30} />
+        ) : (
+          <button className="load-more-button" onClick={onLoadMore}>
+            Show More
+          </button>
```
nickgros, line 203: Switch to react-bootstrap's `Button` and set the `variant` prop to `'outline'`. We've defined this button style in bootstrap already, so hopefully you can get rid of this `load-more-button` class

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+      ]
+      const searchRequest: AccessApprovalSearchRequest = {
+        accessorId,
+        accessRequirementId,
+        sort,
+      }
+      const sortRequest = await searchAccessApprovals(
+        searchRequest,
+        accessToken,
+      )
```
nickgros, line 102: No reason for sorted requests to be invoked differently than any other prop change.

nickgros (commented):
Some more comments, would also like to see at least a few basic tests

src/__tests__/lib/containers/AccessApprovalsTable.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  it('Shows the table of access approvals', async () => {
+    renderComponent()
+
+    await screen.findByText('AR ID')
+    await screen.findByText('Access Requirement Name')
+    await screen.findByText('Submitter')
+    await screen.findByText('Status')
+    await screen.findByText('Modified Date')
+    await screen.findByText('Expires')
+  })
```
nickgros, line 29: Additional tests: 
* Verify data is shown
* Verify additional data is shown after clicking 'Show More'

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  } remove-border`}
+                >
+                  {expiredOn}
+                </td>
+              </tr>
+            )
+          })}
+        </tbody>
+      </Table>
+    </div>
```
nickgros, line 125: Seems we lost the 'Show More' button, which would call `fetchNextPage` returned by the infinite query hook. The button should only be shown when `hasNextPage` (also returned by the hook) is true, and should be disabled or hidden while fetching the next page.

src/__tests__/lib/containers/AccessApprovalsTable.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  render(<AccessApprovalsTable accessorId="123" />, {
+    wrapper: createWrapper(wrapperProps),
+  })
+}
+
+describe('AccessApprovalsTable tests', () => {
+  beforeAll(() => server.listen())
+  afterEach(() => server.restoreHandlers())
+  afterAll(() => server.close())
+
```
nickgros, line 19: See SubmissionPage.test.tsx for examples on how to set up Mock Service Worker

nickgros (changes requested):
Looking very clean! See comments

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+                style={{ float: 'right' }}
+                onClick={() =>
+                  onSort(AccessApprovalSortField.MODIFIED_ON, isDescending)
+                }
+              >
+                <IconSvg
+                  options={{
+                    icon: 'sortTwoTone',
+                  }}
+                />
```
nickgros, line 71: Let's use SortIcon in Sort.tsx, which has props/styling that indicate the current sort

Something like this (I don't think this will work perfectly):
```suggestion
                <SortIcon
                  active={sort.field === 'CREATED_ON'}
                  direction={
                    sort.field === 'CREATED_ON' ? sort.direction : 'DESC'
                  }
                />
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+  AccessApprovalsTableProps
+> = ({ accessorId, accessRequirementId }) => {
+  const [isDescending, setIsDescending] = useState(false)
+  const [sort, setSort] = useState<AccessApprovalSearchSort[]>()
+
+  const searchRequest: AccessApprovalSearchRequest = {
+    accessorId,
+    accessRequirementId,
+    sort,
+  }
```
nickgros, line 34: Couple of (fairly picky) notes but hopefully you agree it helps clean it up:
* We're not using more than one sort, so we don't need to store an array
* isDescending can be determined from the sort object, so let's get rid of it
* No reason not to have a default sort state, which will allow us to not worry about an undefined value
* We should memoize the final request object, just to ensure we don't dispatch unnecessary requests

```suggestion
  const [sort, setSort] = useState<AccessApprovalSearchSort>({ field: AccessApprovalSortField.MODIFIED_ON, direction: SortDirection.DESC  })

  const searchRequest: AccessApprovalSearchRequest = useMemo(() => ({
    accessorId,
    accessRequirementId,
    sort: [sort],
  }), [accessorId, accessRequirementId, sort])
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const { data, hasNextPage, fetchNextPage, isFetching } =
+    useSearchAccessApprovalsInfinite(searchRequest)
+  const accessApprovals = data?.pages.flatMap(page => page.results) ?? []
+  const onSort = (field: AccessApprovalSortField, isDescending: boolean) => {
+    const direction: SortDirection = isDescending
+      ? ('DESC' as SortDirection)
+      : ('ASC' as SortDirection)
+    setSort([{ field, direction }])
+    setIsDescending(!isDescending)
+  }
```
nickgros, line 45: Update this to reflect that we got rid of `isDescending`.

The behavior here subtly changes too; if we switch the sorted field, it will always start 'DESC', rather than always switching directions even when you change fields. This is more predictable from a user's perspective

```suggestion
  const onSort = (field: AccessApprovalSortField) => {
    if (sort.field === field) {
      setSort({ field, direction: sort.direction === SortDirection.DESC ? SortDirection.ASC : SortDirection.DESC })
    } else {
      setSort({ field, direction: SortDirection.DESC })
    }
  }
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+            </th>
+            <th className="thead-cell remove-border">
+              Expires
+              <SortIcon
+                role="button"
+                style={{ float: 'right' }}
+                active={false}
+                direction={
+                  sort && sort[0].field === 'EXPIRED_ON'
+                    ? sort[0].direction === 'DESC'
```
nickgros, line 86: No longer undefined or an array
```suggestion
                  sort.field === 'EXPIRED_ON'
                    ? sort.direction === 'DESC'
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+            <th className="thead-cell remove-border">Status</th>
+            <th className="thead-cell remove-border">
+              Modified Date
+              <SortIcon
+                role="button"
+                style={{ float: 'right' }}
+                active={false}
+                direction={
+                  sort && sort[0].field === 'MODIFIED_ON'
+                    ? sort[0].direction === 'DESC'
```
nickgros, line 68: No longer undefined or an array
```suggestion
                  sort.field === 'MODIFIED_ON'
                    ? sort.direction === 'DESC'
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+              Access Requirement Name
+            </th>
+            <th className="thead-cell remove-border">Submitter</th>
+            <th className="thead-cell remove-border">Status</th>
+            <th className="thead-cell remove-border">
+              Modified Date
+              <SortIcon
+                role="button"
+                style={{ float: 'right' }}
+                active={false}
```
nickgros, line 65: Use active to indicate if the field is currently sorted

```suggestion
                active={sort.field === AccessApprovalSortField.MODIFIED_ON}
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  onSort(AccessApprovalSortField.MODIFIED_ON, isDescending)
+                }
+              />
+            </th>
+            <th className="thead-cell remove-border">
+              Expires
+              <SortIcon
+                role="button"
+                style={{ float: 'right' }}
+                active={false}
```
nickgros, line 83: Same as above
```suggestion
                active={sort.field === AccessApprovalSortField.EXPIRED_ON}
```

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  >
+                    {item.accessRequirementId}
+                  </a>
+                </td>
+                <td className="remove-border">{item.accessRequirementName}</td>
+                <td className="remove-border">
+                  <UserCard size={SMALL_USER_CARD} ownerId={item.submitterId} />
+                </td>
+                <td className="remove-border">
+                  {item.state[0] + item.state.slice(1).toLocaleLowerCase()}
```
nickgros, line 116: I think using lodash upperFirst is more readable, took me a second to realize why you were using .slice
```suggestion
                  {upperFirst(item.state.toLocaleLowerCase())}
```

src/lib/containers/IconSvg.tsx
```diff
@@ -249,6 +250,8 @@ const getIcon = (options: IconSvgOptions) => {
       return <ChatTwoTone style={muiSvgStyle}></ChatTwoTone>
     case 'accessManagement':
       return <AccessManagement style={muiSvgStyle}></AccessManagement>
+    case 'sortTwoTone':
+      return <SortTwoTone style={muiSvgStyle}></SortTwoTone>
```
nickgros, line 254: We can remove this if not using anymore

src/lib/style/components/_access-approvals.scss
```diff
@@ -0,0 +1,21 @@
+@use '../abstracts/variables' as SRC;
+@use 'sass:map';
+
+.AccessApprovalsTable {
+
+    .access-approval-header {
+        background-color: map.get(SRC.$colors, 'gray-200');
+        .thead-cell {
```
nickgros, line 8: Why not this and remove the class?
```suggestion
        th {
```

src/__tests__/lib/containers/AccessApprovalsTable.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+            nextPageToken: null,
+          },
+        ],
+        pageParams: [],
+      },
+      fetchNextPage: mockFetchNextPage,
+      hasNextPage: true,
+      isLoading: false,
+      isSuccess: true,
+    })
```
nickgros, line 62: Just a note for the future, I prefer if we mock the service with MSW rather than mocking react-query's behavior; the test becomes more brittle if we modify the component to use other react-query features later.

This is fine though, so no need to change this for this PR.

nickgros (changes requested):

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+                <td className="remove-border">
+                  <UserCard size={SMALL_USER_CARD} ownerId={item.submitterId} />
+                </td>
+                <td className="remove-border">
+                  {upperFirst(item.state.toLocaleLowerCase())}
+                </td>
+                <td className="remove-border">{modifiedOn}</td>
+                <td
+                  className={`${
+                    new Date() > new Date(expiredOn) ? 'expired' : ''
```
nickgros, line 127: Not sure if this check still works now that expiredOn is a string

yjyoo773: This does work. However, I've run into another edge case where if the expired Date does not exist it shows the current date and time. Should I set that case with the value null?

yjyoo773: When the value is null does it mean that the expiration date is `never` or it does not exist?

nickgros: I would think that it means it never expires

nickgros (approved):
Great work, @yjyoo773 !

PR #1541, opened 2022-05-26
PORTALS-2179: provide way to set the default visibility state of the …

jay-hodgson:
fill in gaps for EntityType icons, and fix svg size in card


yjyoo773 (approved):
LGTM!

PR #1546, opened 2022-06-01
SWC-6150

nickgros:
Fixes renderer for USERID facet filter when an ID belongs to a team

### DIFF
--- a/src/lib/containers/download_list/DownloadListTable.tsx
+++ b/src/lib/containers/download_list/DownloadListTable.tsx
@@ -26,7 +26,7 @@ import {
   FileHandleAssociation,
   PaginatedResults,
   Reference,
-  UserProfile,
+  UserGroupHeader,
 } from '../../utils/synapseTypes'
 import HasAccess, {
   getDownloadTypeForFileHandle,
@@ -118,7 +118,7 @@ export default function DownloadListTable(props: DownloadListTableProps) {
     ownerIds.push(...ownerIdsFromHeaders)
   }
   // use bang operator because filter function guarentee's that file handle will be defined
-  const userProfiles = useGetInfoFromIds<UserProfile>({
+  const userProfiles = useGetInfoFromIds<UserGroupHeader>({
     ids: ownerIds,
     type: 'USER_PROFILE',
   })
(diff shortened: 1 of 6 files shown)


src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -53,27 +57,27 @@ const getEntityHeaderItems = async (
...
 }
 
 // React hook to get user profiles or entities
 //******************************************************************************************* */
-export default function useGetInfoFromIds<T extends EntityHeader | UserProfile>(
-  props: UseGetInfoFromIdsProps,
-) {
+export default function useGetInfoFromIds<
+  T extends EntityHeader | UserGroupHeader,
+>(props: UseGetInfoFromIdsProps): T[] {
```
nickgros, line 80: Hook that stores/caches EntityHeaders or UserProfiles. Change to use UserGroupHeaders instead of UserProfiles, so it works for teams (and can be batched instead of making _n_ requests!)

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -26,7 +26,7 @@ import {
   FileHandleAssociation,
```
nickgros, line 26: I think this is the deprecated (unused today?) download list component

src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -8,7 +8,7 @@ import useDeepCompareEffect from 'use-deep-compare-effect'
 import { ElementWithTooltip } from '../../../containers/widgets/ElementWithTooltip'
```
nickgros, line 8: This is the component where the bug surfaced, but no actual changes at this level (just variable rename + type updates)

src/lib/utils/SynapseClient.ts
```diff
@@ -891,7 +891,7 @@ export const getUserGroupHeaders = (
  */
 export const getGroupHeadersBatch = (
   ids: string[],
-  accessToken: string | undefined,
+  accessToken?: string,
```
nickgros, line 894: Make accessToken optional

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -53,27 +57,27 @@ const getEntityHeaderItems = async (
   return [...newData.results, ...notFoundPlaceholders]
 }
 
-const getUserProfileItems = async (
+const getUserGroupHeaderItems = async (
   lookupList: string[],
-): Promise<UserProfile[]> => {
-  const newData = await getUserProfileWithProfilePicAttached(lookupList)
+): Promise<UserGroupHeader[]> => {
+  const newData = (await getGroupHeadersBatch(lookupList)).children
```
nickgros, line 63: Get headers instead of profiles

PR #1547, opened 2022-06-01
PORTALS-2176

jay-hodgson:
Screen Shot 2022-06-01 at 12 00 42 PM

Screen Shot 2022-06-01 at 12 00 50 PM

### DIFF
--- a/src/lib/style/components/_query-wrapper-text-input.scss
+++ b/src/lib/style/components/_query-wrapper-text-input.scss
@@ -39,6 +39,12 @@
       flex-grow: 0;
       flex-shrink: 0;
     }
+    &__helpbutton {
+      flex: 1;
+      flex-grow: 0;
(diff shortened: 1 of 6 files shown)


src/lib/style/components/_query-wrapper-text-input.scss
```diff
@@ -2,6 +2,13 @@
 
 .QueryWrapperSearchInput,
 .QueryWrapperSqlEditorInput {
+  &__helppopoverwrapper {
+    display: grid;
```
yjyoo773, line 6: Is it better to use grid in this type of case?

jay-hodgson: In this case I have one cell that I want to hardcode the width, align, and justify.  I actually initially used a flexbox, but it was too flexible ;)

PR #1550, opened 2022-06-02
SWC-6148

nickgros:
Table for Access Requirements

image


src/lib/style/components/_all.scss
```diff
@@ -19,5 +19,5 @@
   './user-profile-links', './blocking-loader', './transition', './popover',
   './featured-tool-card', './row_renderers/observation-card',
   './subsection-row-renderer', './dataset-editor', './fluid-modal',
-  './markdown-collapse', './discussion-search', './submission-page', './access-approvals';
-
+  './markdown-collapse', './discussion-search', './submission-page',
+  './governance-dashboard';
```
nickgros, line 23: ./governance-dashboard can hold the shared styles for all three tables

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -59,69 +59,65 @@ export const AccessApprovalsTable: React.FunctionComponent<
         Status in Access Requirements
       </Typography>
       <Table striped borderless bordered={false}>
-        <thead className="access-approval-header">
```
nickgros, line 62: Just dropped unnecessary class names and reorganized structure to match AccessRequirementTable (and match the now-shared CSS)

src/lib/testutils/TestingLibraryUtils.tsx
```diff
@@ -28,14 +28,25 @@ export const createWrapper = (props?: SynapseContextType) => {
...
   }
 }
+
+/**
+ * Creates a test wrapper for components being tested with @testing-library/react. This wrapper
+ * includes context and an isolated query cache.
+ */
+export const createWrapper = (props?: SynapseContextType) => {
+  return createWrapperAndQueryClient(props).wrapperFn
+}
```
nickgros, line 52: Behavior unchanged. The new method allows retrieving the QueryClient so it can be accessed and modified in a test

src/lib/containers/dataaccess/AccessRequirementTable.tsx
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      setSort({ field, direction: 'DESC' })
+    }
+  }
+
+  return (
+    <div className="bootstrap-4-backport">
+      <div className="SRC-split">
+        <Typography variant="headline3" style={{ marginBottom: 0 }}>
+          Access Requirements
```
yjyoo773, line 72: Should this have the number of ARs on table?

nickgros: This was missed earlier so we don't have a backend service to determine this number. I brought it up with Adam, we'll backfill this feature if we determine it's necessary.

yjyoo773 (approved):

PR #1551, opened 2022-06-02
SWX-6136: fix Powered By Synapse section breakpoint that causes a row with 4 items

jay-hodgson:
…  (instead of maintaining a 3x3 grid on large screens)

### DIFF
--- a/src/lib/style/components/_synapse-homepage.scss
+++ b/src/lib/style/components/_synapse-homepage.scss
@@ -306,6 +306,7 @@
     display: inline-flex;
     flex-wrap: wrap;
     align-content: center;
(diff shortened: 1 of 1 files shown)


jay-hodgson:
Without max-width in the style:
Screen Shot 2022-06-02 at 12 01 01 PM
With max-width in the style:
Screen Shot 2022-06-02 at 12 00 39 PM

nickgros (approved):
This could be brittle if the widths of the inner items change. I don't think grid or flex have any features that would let us get exactly what we want. Media queries would be more lines of code and likely just as brittle, so I think this is the best solution.

PR #1553, opened 2022-06-03
SWC-6147: change Dataset icon to match design

jay-hodgson:


jay-hodgson:
Quick test in the demo app (updating one of the components to use DATASET instead of EXPERIMENTAL_TOOLS)
Screen Shot 2022-06-03 at 11 18 35 AM

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -42,6 +42,7 @@ import {
 } from '../../../utils/SynapseConstants'
 
 import { ReactComponent as Data2Svg } from '../../../assets/icons/Data2.svg'
+import { ReactComponent as DatasetSvg } from '../../../assets/icons/Dataset.svg'
```
nickgros, line 45: I think this exists in MUI -- see '@material-ui/icons/GridOnTwoTone'. Can we use those in this Icon component?

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -138,26 +142,24 @@ const Icon: React.FunctionComponent<IconProps> = ({
...
 }) => {
+  const IconSvg = iconSvgMapping[value] || iconSvgMapping[type]
+  if (IconSvg) return IconSvg
+
   const iconSet = { ...defaultIcons, ...iconOptions }
   // see if the value has a corresponding icon, e.g. 'Active' in a studies table
   // or if the type of card has a corresponding icon, e.g. 'Publication'
-  const Icon = iconSet[value] || iconSet[type]
-  // TODO: get rid of dataset icon class, none of the icons should be special cased
-  const datasetCustomStyle =
```
jay-hodgson, line 146: Removed custom styling for Dataset

src/lib/containers/row_renderers/utils/Icon.tsx
```diff
@@ -138,26 +142,24 @@ const Icon: React.FunctionComponent<IconProps> = ({
   iconOptions,
   cssClass,
 }) => {
+  const IconSvg = iconSvgMapping[value] || iconSvgMapping[type]
```
jay-hodgson, line 145: If this type/value is supported by our IconSvg mapping above, then return it.

jay-hodgson:
Uses MUI icon instead.
Screen Shot 2022-06-03 at 3 44 03 PM

PR #1559, opened 2022-06-06
PORTALS-2204

jay-hodgson:


jay-hodgson:
Standard view:
Screen Shot 2022-06-06 at 4 22 12 PM

Loading:
Screen Shot 2022-06-06 at 4 22 45 PM

ACT also has access to the accessorChanges, which renders as it used to (small user card, and access type change)

src/lib/containers/IDUReport.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <>
+      {allRows.length > 0 && (
+        <div className="IDUReport">
+          {allRows.map((item: SubmissionInfo) => {
+            if (item) {
+              return <SubmissionInfoCard info={item} />
+            } else return false
+          })}
+          {/* To trigger loading the next page */}
+          <div ref={ref} />
```
nickgros, line 64: Is this desired/how the SWC version works vs a 'Show more' button?

src/lib/containers/IDUReportSubmissionInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const SubmissionInfoCard: React.FunctionComponent<
+  SubmissionInfoCardProps
+> = ({ info }: SubmissionInfoCardProps) => {
+  return (
+    <>
+      {info && (
+        <p className="SubmissionInfoCard">
+          <strong> Project Lead: </strong> <span>{info.projectLead}</span>{' '}
```
nickgros, line 21: Would it be a big change to use the Typography styled components?

src/lib/containers/IDUReportSubmissionInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const LoadingSubmissionInfoCard: React.FunctionComponent = () => {
+  return (
+    <p className="SubmissionInfoCard">
+      <SkeletonTable numCols={1} numRows={4} />
+      <Skeleton variant="rect" width={'100%'} height={80} />
+    </p>
+  )
+}
```
nickgros, line 56: These custom skeletons are really satisfying 🙂

src/lib/containers/IDUReport.tsx
```diff
@@ -0,0 +1,76 @@
...
+    isError,
+    error: newError,
+  } = useGetApprovedSubmissionInfoInfinite(accessRequirementId)
+
+  useEffect(() => {
+    if (isError && newError) {
+      handleError(newError)
+    }
+  }, [isError, newError, handleError])
+
```
nickgros, line 39: Can be simplified:
```suggestion
  } = useGetApprovedSubmissionInfoInfinite(accessRequirementId, {
      useErrorBoundary: true
  })
```

src/__tests__/lib/containers/IDUReport.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  })
+  it('loads more when inView', async () => {
+    mockAllIsIntersecting(true)
+
+    renderComponent()
+
+    const submissionInfo1 = await screen.findAllByText('Prof. Farnsworth')
+    expect(submissionInfo1).toHaveLength(1)
+    const submissionInfo2 = await screen.findAllByText('Wong')
+    expect(submissionInfo2).toHaveLength(1)
```
nickgros, line 81: First page loads before inView right?

Can this be changed to something like-
```suggestion
    renderComponent()

    const submissionInfo1 = await screen.findAllByText('Prof. Farnsworth')
    expect(submissionInfo1).toHaveLength(1)
    
    mockAllIsIntersecting(true)
    
    const submissionInfo2 = await screen.findAllByText('Wong')
    expect(submissionInfo2).toHaveLength(1)
```

src/lib/containers/IDUReport.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const allRows = data?.pages.flatMap(page => page.results) ?? []
+
+  return (
+    <>
+      {allRows.length > 0 && (
+        <div className="IDUReport">
+          {allRows.map((item: SubmissionInfo) => {
+            if (item) {
+              return <SubmissionInfoCard info={item} />
```
nickgros, line 60: Any lint error about missing `key` prop?

src/lib/containers/IDUReport.tsx
```diff
@@ -0,0 +1,76 @@
...
+            } else return false
```
nickgros, line 61: Include key if necessary (see above)
```suggestion
            } else return <React.Fragment key={'...'} />
```

src/lib/containers/IDUReportSubmissionInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+            info.accessorChanges.map(accessorChange => (
+              <div key={`${accessorChange.userId}-${accessorChange.type}`}>
+                <UserCard
+                  ownerId={accessorChange.userId}
+                  size={SMALL_USER_CARD}
+                />{' '}
+                {accessorChange.type}
+              </div>
+            ))}
+          <hr />
```
nickgros, line 42: Seems like this `<hr />` should be rendered between cards, rather than shown as a part of this component.

nickgros (changes requested):

src/lib/containers/IDUReport.tsx
jay-hodgson: Initially matching SWC implementation, as changes would need to be discuss with the governance team for this widget

src/lib/containers/IDUReportSubmissionInfo.tsx
jay-hodgson: Initially matching SWC implementation.  There are old IDUs that are hard-coded in the wiki, and this matches their styling.

src/lib/containers/IDUReportSubmissionInfo.tsx
jay-hodgson: Agreed

src/lib/containers/IDUReport.tsx
jay-hodgson: Oddly, no!  Why is that?  I'll set the key based on the values in item

src/lib/containers/IDUReportSubmissionInfo.tsx
jay-hodgson: Agreed, I'll fix

src/lib/containers/IDUReport.tsx
```diff
@@ -57,8 +48,13 @@ export const IDUReport: React.FunctionComponent<IDUReportProps> = (
         <div className="IDUReport">
           {allRows.map((item: SubmissionInfo) => {
             if (item) {
-              return <SubmissionInfoCard info={item} />
-            } else return false
```
nickgros: Can we remove the conditional? Not sure why item would be false

nickgros (approved):

nickgros (approved):

PR #1561, opened 2022-06-07
Small improvements to tables

nickgros:


src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -38,7 +38,7 @@ export const AccessApprovalsTable: React.FunctionComponent<
     [accessorId, accessRequirementId, sort],
   )
 
-  const { data, hasNextPage, fetchNextPage, isFetching } =
+  const { data, hasNextPage, fetchNextPage, isLoading } =
```
nickgros, line 41: `isLoading` is true whenever we are fetching data, and we don't have any data currently loaded
`isFetching` is true whenever data is being fetched (including when stale data is currently loaded)

`isLoading` is more appropriate for our spinners in these cases

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -136,10 +136,18 @@ export const AccessApprovalsTable: React.FunctionComponent<
...
+      {isLoading && (
+        <div className="SRC-center-text">
+          <SynapseSpinner size={40} />
+        </div>
+      )}
+      {!isLoading && accessApprovals.length === 0 && (
+        <Typography className="SRC-center-text" variant="body1">
+          No Results
+        </Typography>
+      )}
```
nickgros, line 148: Add no results state

src/lib/containers/AccessApprovalsTable.tsx
```diff
@@ -136,10 +136,18 @@ export const AccessApprovalsTable: React.FunctionComponent<
           })}
         </tbody>
       </Table>
+      {isLoading && (
+        <div className="SRC-center-text">
+          <SynapseSpinner size={40} />
+        </div>
+      )}
```
nickgros, line 143: Since `isLoading` is only true when we we have no data and are fetching more, it can be pulled out of a complex expression.

yjyoo773 (approved):

PR #1562, opened 2022-06-07
React Router v6 pre-migration

nickgros:
Followed the [React Router v6 migration guide](https://reactrouter.com/docs/en/v6/upgrading/v5) as far as I could

* Bumped to latest v5 version
* Made changes suggested by guide (mostly in deprecated Playground)
* Removed ShowDownload (unused, it used deprecated `withRouter`)
* Removed react-router (not needed since we have react-router-dom)

Only one place in SRC where v6 is incompatible: SynapseForm uses Prompt, which conditionally shows an alert before a user navigates away from a page (in this case, when the form contains unsaved data), which was not included in v6.

Corresponding change in portals here: https://github.com/Sage-Bionetworks/portals/pull/778
Ticket to finish the work: https://sagebionetworks.jira.com/browse/PORTALS-2208


esbuild.config.mjs
```diff
@@ -7,7 +7,7 @@ import GlobalsPlugin from 'esbuild-plugin-globals'
 const globals = {
   react: 'React',
   'react-dom': 'ReactDOM',
-  'react-router-dom': 'ReactRouterDom',
+  'react-router-dom': 'ReactRouterDOM',
```
nickgros, line 10: Properly match react-router-dom's named UMD export

src/demo/containers/playground/Playground.tsx
```diff
@@ -1,104 +1,81 @@
...
-  match,
-}: {
-  match?: RouteChildrenProps['match']
-  token: string
-}) => {
-  if (!match) {
-    return <div />
-  }
+const App = () => {
+  const { url } = useRouteMatch()
```
nickgros, line 21: Many of the React Router v5 -> v6 migration steps can be summarized as "use hooks instead of passing props, when possible".

src/demo/containers/App.tsx
```diff
@@ -170,13 +169,12 @@ export default class App extends React.Component<{}, AppState> {
...
-              path="/Playground"
-              component={(props: RouteChildrenProps) => (
-                <Playground {...props} token={token} />
-              )}
-            />
+            <Route exact={true} path="/">
+              <Demo />
+            </Route>
+            <Route path="/Playground">
+              <Playground />
```
jay-hodgson, line 176: If the Playground caused too much trouble, maybe we should delete it from the codebase?

jay-hodgson (approved):

PR #1569, opened 2022-06-08
SWC-6156, SWC-6157, SWC-6161, SWC-6158

jay-hodgson:


src/lib/containers/FullWidthAlert.tsx
```diff
@@ -1,10 +1,9 @@
 import React, { useEffect } from 'react'
 import { Button, Alert, AlertProps } from 'react-bootstrap'
 import {
-  InfoOutlined,
-  ErrorOutlined,
-  WarningOutlined,
-  CheckCircleOutlined,
+  InfoTwoTone,
```
jay-hodgson, line 4: Fix icons (use two tone versions, and use ErrorTwoTone for both warning and error)

src/lib/containers/FullWidthAlert.tsx
```diff
@@ -157,7 +156,7 @@ function FullWidthAlert(props: FullWidthAlertProps) {
         />
         <ButtonFromConfig
           config={primaryButtonConfig}
-          variant="secondary"
+          variant="primary"
```
jay-hodgson, line 159: Primary button should have primary style.
https://sagebionetworks.jira.com/browse/SWC-6158

src/lib/style/components/_global-alert.scss
```diff
@@ -38,6 +38,7 @@
       grid-row: 1 / span 1;
       justify-self: end;
       padding-bottom: 6px;
+      color: #878E95;
```
jay-hodgson, line 41: Close button should always be this grey.  https://sagebionetworks.jira.com/browse/SWC-6157

src/lib/containers/FullWidthAlert.tsx
nickgros: The style referred to in the design is 'sds-primary'

```suggestion
          variant="sds-primary"
```

Could be worth opening a ticket to replace all old button styles with these new ones

nickgros (changes requested):

src/lib/style/base/_core.scss
```diff
@@ -1286,6 +1286,55 @@ main,
...
+    -webkit-border-radius: $radius;
+    -moz-border-radius: $radius;
+    -ms-border-radius: $radius;
+    border-radius: $radius;
+    -o-border-radius: $radius;
+    -khtml-border-radius: $radius;
+  }
+  
+  // Set up default bootstrap alert styles across the portals
+  .bootstrap-4-backport.alert {
```
jay-hodgson, line 1299: I think design would like all portals to have matching alert styles, so let's remove css from portal side and rely on this shared definition.

jay-hodgson:
After reviewing https://www.figma.com/file/0oPm5lLSUva8kyfVNMS6FA/Sage-Style-%26-Component-Library
Screen Shot 2022-06-08 at 11 59 09 AM
Screen Shot 2022-06-08 at 11 59 28 AM
Screen Shot 2022-06-08 at 11 59 46 AM
Screen Shot 2022-06-08 at 12 00 28 PM
Screen Shot 2022-06-08 at 12 10 03 PM

src/lib/style/base/_core.scss
nickgros: Is this correct, or should it be this?

```suggestion
  .bootstrap-4-backport .alert {
```

If the original is correct, then this will only impact FullWidthAlert, so I think this style should be moved there (and should probably also include the .FullWidthAlert class)

If the suggestion is correct, then this will impact many components in SRC, which I would want to spot check

nickgros (approved):

PR #1570, opened 2022-06-08
Show ID in AR table

nickgros:
image

### DIFF
--- a/src/lib/containers/EntityIdList.tsx
+++ b/src/lib/containers/EntityIdList.tsx
@@ -16,7 +16,7 @@ const EntityIdList: React.FC<EntityIdListProps> = props => {
   let mounted: boolean = true
 
   useEffect(() => {
-    if (mounted && inView) {
+    if (inView) {
       getEntityTypes()
     }
     return () => {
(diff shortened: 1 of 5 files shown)


src/lib/containers/EntityIdList.tsx
```diff
@@ -16,7 +16,7 @@ const EntityIdList: React.FC<EntityIdListProps> = props => {
   let mounted: boolean = true
 
   useEffect(() => {
-    if (mounted && inView) {
+    if (inView) {
```
nickgros, line 19: mounted will never be false because the effect won't be invoked after mounted is set to false in the cleanup.

src/lib/containers/EntityIdList.tsx
```diff
@@ -29,12 +29,16 @@ const EntityIdList: React.FC<EntityIdListProps> = props => {
 
     getEntityHeadersByIds(entityIdList, accessToken)
       .then(entity => {
-        const newEntityList = entity.results.map(el => {
-          return (
-            <EntityLink entity={el} key={el.id} displayTextField={'name'} />
-          )
-        })
-        setEntityLinkArray(newEntityList)
+        if (mounted) {
```
nickgros, line 32: mounted should be checked in the asynchronous callback, where it is possible for mounted to be false when this is invoked

src/lib/containers/EntityLink.tsx
```diff
@@ -52,21 +52,19 @@ export const EntityLink = (props: EntityLinkProps) => {
...
+          rel="noopener noreferrer"
+          href={`${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!Synapse:${entity.id!}${
+            versionNumber ? `.${versionNumber}` : ''
+          }`}
+        >
+          {showIcon && (
+            <EntityTypeIcon type={type} style={{ marginRight: '6px' }} />
+          )}
+          {entity[displayTextField]}
+        </a>
```
nickgros, line 67: Remove the <p> rather than adding custom styling to make it inline. Wrapped in <p> where it was important for styling (for example, in SynapseTableCell renderers)

PR #1576, opened 2022-06-09
SWC-6162 - Optimize QueryWrapper, include version in bundle

nickgros:
Mostly reducing the frequency of table query requests. This doesn't fix the main bug described in SWC-6162. The real bug is in SWC, so these changes are fine to go in develop.

### DIFF
--- a/src/lib/containers/QueryWrapper.tsx
+++ b/src/lib/containers/QueryWrapper.tsx
@@ -135,7 +135,6 @@ export function QueryWrapper(props: QueryWrapperProps) {
     fetchNextPage,
     isFetchingNextPage,
     isLoading: queryIsLoading,
-    refetch,
     error,
     isPreviousData: newQueryIsFetching,
     remove,
@@ -310,8 +309,6 @@ export function QueryWrapper(props: QueryWrapperProps) {
      */
     remove()
     // end TODO
-
-    refetch()
   }
 
   /**
@@ -319,7 +316,7 @@ export function QueryWrapper(props: QueryWrapperProps) {
    * this is to remove the facet from the charts, search and filter.
    * @return data: QueryResultBundle
    */
-  function removeLockedFacetData() {
+  const dataWithLockedFacetRemoved = useMemo(() => {
     const lockedFacet = props.lockedFacet?.facet
     if (lockedFacet && data) {
       // for details page, return data without the "locked" facet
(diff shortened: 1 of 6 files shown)


src/lib/containers/QueryWrapper.tsx
```diff
@@ -310,16 +309,14 @@ export function QueryWrapper(props: QueryWrapperProps) {
      */
     remove()
     // end TODO
-
-    refetch()
```
nickgros, line 314: Seemed to be unnecessary in my manual testing

src/lib/containers/QueryWrapper.tsx
```diff
@@ -310,16 +309,14 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
   }
 
   /**
    * remove a particular facet name (e.g. study) and its all possible values based on the parameter specified in the url
    * this is to remove the facet from the charts, search and filter.
    * @return data: QueryResultBundle
    */
-  function removeLockedFacetData() {
+  const dataWithLockedFacetRemoved = useMemo(() => {
```
nickgros, line 319: Memoize rather than invoking this function every render

src/lib/containers/table/SynapseTable.tsx
```diff
@@ -940,7 +940,7 @@ export default class SynapseTable extends React.Component<
     if (isShowingAddToV2DownloadListColumn) {
       tableColumnHeaderElements.unshift(
         <th
-          data-testId="AddToDownloadListV2ColumnHeader"
+          data-testid="AddToDownloadListV2ColumnHeader"
```
nickgros, line 943: Resulted in console warning

src/lib/umd.index.ts
```diff
@@ -99,4 +99,14 @@ const SynapseComponents = {
   ForumSearch,
 }
 
-export { SynapseComponents, SynapseConstants, SynapseClient, SynapseContext }
+// Include the version in the build
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-member-access
+const SynapseReactClientVersion = require('../../package.json').version
```
nickgros, line 104: I want to do this:
image

src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
```diff
@@ -20,6 +20,10 @@ import {
 } from '../../../synapseTypes'
 import { entityQueryKeys } from './queryKeys'
 
+const sharedQueryDefaults = {
+  refetchOnWindowFocus: false,
```
nickgros, line 24: Refetch on window focus is handy for when you come back to a window after a while, but that might be a little too often for table queries.

jay-hodgson (approved):

PR #1583, opened 2022-06-11
PORTALS-2216: Remove font awesome dependency

jay-hodgson:
and yarn upgrade


src/lib/containers/EntityBadgeIcons.tsx
```diff
@@ -1,15 +1,14 @@
...
-  faAlignLeft,
-  faCheck,
-  faComment,
-  faGlobeAmericas,
-  faLock,
-  faTag,
-  faUnlink,
-} from '@fortawesome/free-solid-svg-icons'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+  DescriptionTwoTone,
```
jay-hodgson, line 4: Initially I was pulling material icons directly, but realized that the IconSvg wrapper was useful in a number of contexts (to add styling to a wrapper).  So there's still a mix

src/lib/containers/EntityIcon.tsx
```diff
@@ -1,49 +1,9 @@
...
-  faFolder,
-  faFile,
-  faListAlt,
-  faServer,
-  faTable,
-  faThList,
-  faDocker,
-)
-
-const _ENTITY_TYPE_ICON_MAP = new Map<EntityType, IconDefinition>([
```
jay-hodgson, line 33: Mapping already exists in IconSvg, use it instead

src/lib/containers/EntityIcon.tsx
```diff
@@ -55,20 +15,22 @@ const getIconTypeForEntity = (type: EntityType): IconDefinition | '' => {
...
+      return type2SvgIconName[type]!
     default:
       return ''
   }
 }
 
 type EntityTypeIconProps = {
   type: EntityType
+  style?: React.CSSProperties
+  className?: string
```
jay-hodgson, line 27: Some callers were setting explicit style and class names

src/lib/containers/ForumSearch.tsx
```diff
@@ -104,7 +104,7 @@ const ForumSearch = (props: ForumSearchProps) => {
     <div className="bootstrap-4-backport ForumSearch">
       <div>
         <span className="SearchIcon">
-          <IconSvg options={{ icon: 'searchTwoTone' }} />
+          <IconSvg options={{ icon: 'search' }} />
```
jay-hodgson, line 107: Renamed searchTwoTone to search

src/lib/containers/IconSvg.tsx
```diff
@@ -151,15 +188,17 @@ export type SVGStyleProps = {
   color?: string
   verticalAlign?: string
   fill?: string
+  width?: string
 }
 
 const getIcon = (options: IconSvgOptions) => {
-  const { icon, color } = options
+  const { icon, color, size } = options
```
jay-hodgson, line 195: Setting size didn't seem to do anything to the material icons.  Now it sets the width of the svg (so it has a direct impact on the size since the aspect is maintained)

src/lib/containers/download_list/DownloadListTable.tsx
```diff
@@ -379,15 +371,16 @@ export default function DownloadListTable(props: DownloadListTableProps) {
...
-                  <FontAwesomeIcon
-                    icon={
-                      sortedColumn.column === 'file'
-                        ? sortedColumn.isDescending === false
-                          ? faSortAmountDown
-                          : faSortAmountUp
-                        : faSortAmountDown
-                    }
-                    color={sortedColumn.column === 'file' ? 'white' : ''}
+                  <IconSvg
```
jay-hodgson, line 374: Ok, these don't look great, but this component uses the old Download List services and is deprecated.  It should just be deleted at this point instead of patching!

jay-hodgson:
Updates eventsource from 1.0.7 to 2.0.2, so fixes https://github.com/advisories/GHSA-6h5x-7c5m-7cr7 (we are tracking in https://sagebionetworks.jira.com/browse/PORTALS-2202)

jay-hodgson:
No longer see json-schema@0.2.3 in the yarn.lock file, looks like it left only the json-schema@0.4.0, json-schema@^0.4.0 entry.
I think this also means that https://github.com/advisories/GHSA-896r-f27r-55mw should also be fixed, right?  (we're tracking here: https://sagebionetworks.jira.com/browse/PORTALS-2201)

jay-hodgson:
We will certainly run into additional issues, but I think this needs to be done.

Before
Screen Shot 2022-06-10 at 11 34 58 PM

After
Screen Shot 2022-06-10 at 11 56 17 PM

Before
Screen Shot 2022-06-10 at 11 41 57 PM

After
Screen Shot 2022-06-10 at 11 42 09 PM

src/lib/containers/personal_access_token/AccessTokenCard.tsx
```diff
@@ -130,7 +125,7 @@ export const AccessTokenCard: React.FunctionComponent<AccessTokenCardProps> = ({
             }
           }}
         >
-          <FontAwesomeIcon icon={faTrash} aria-hidden="true" />
+          <IconSvg options={{ icon: 'delete' }} aria-hidden="true" />
```
nickgros, line 128: I don't think the aria-hidden gets passed along. IconSvg hould pass most props to its `span` wrapper.

nickgros (approved):
Looks fantastic! Tweaks can come over time, good to just rip off the [faBandage](https://fontawesome.com/icons/bandage?s=solid) for now

PR #1584, opened 2022-06-13
Replace Enzyme tests, remove deprecated download list components

nickgros:
Depends on https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4896

* Replace some of the existing tests using Enzyme with tests that use React Testing Library
* Minor logic changes to improve testability
* Remove deprecated download list components (so we don't have to update their tests)


src/__tests__/lib/containers/AccessRequirements.test.tsx
```diff
@@ -104,126 +156,50 @@ describe('Access Requirement List works as expect', () => {
...
-    expect(wrapper.find(ACTAccessRequirementComponent)).toHaveLength(1)
+  it('Renders a Access Requirements List with valid props', async () => {
+    init(props)
+    // The ManagedACTAccessRequirementComponent renders as "RequestDataAccess"
+    await screen.findByTestId('RequestDataAccess')
+    // The other three access requirements render as "AcceptedRequirements"
+    expect(await screen.findAllByTestId('AcceptedRequirements')).toHaveLength(3)
   })
 
-  it.only('Renders a Access Requirements List with completion order', async () => {
```
nickgros, line 116: Before, only this test was running (it.only causes the others to be skipped!)

src/lib/containers/CardContainer.tsx
```diff
@@ -59,7 +59,7 @@ export const CardContainer = (props: CardContainerProps) => {
       case OBSERVATION_CARD:
         return <ObservationCard {...props} />
       default:
-        return <div /> // this should never happen
+        return <div key={props.key} /> // this should never happen
```
nickgros, line 62: I think it does happen in the test. For now, give it a key to suppress a warning.

src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -150,17 +155,21 @@ export default function AccessRequirementList({
...
+              accessToken,
+              accessRequirementFromProps,
+            )
+          if (!isCancelled) {
+            setAccessRequirements(sortedAccessRequirements)
+          }
         }
 
-        const userProfile = await SynapseClient.getUserProfile(accessToken)
-        setUser(userProfile)
```
nickgros, line 162: Leave fetching the userProfile to react-query (should at least be removed from this hook, because fetching the profile does not need to be blocked by fetching Access Requirements)

src/lib/containers/CardContainer.tsx
jay-hodgson: Please remove or update the comment if it's incorrect

jay-hodgson (approved):

PR #1593, opened 2022-06-14
SWC-6127: Add View Sharing Settings button (callback)

jay-hodgson:
And new Download Permission Required image support, and other minor changes from the updated design


jay-hodgson:
Screen Shot 2022-06-14 at 3 52 49 PM

jay-hodgson:
This is not going to work properly in the context of the portals.  Even though this is an edge case (AD Knowledge portal files should be downloadable if you're logged in and you have been approved for the associated Access Restrictions), it should not be broken in this case.  Recommend in this case we just send the user to the associated Synapse.org entity page (as we did previously) when they click on View Sharing Settings button.

nickgros (approved):

PR #1597, opened 2022-06-16
SWC-6169

nickgros:


nickgros:
Wrong Jira ticket, see #1598 (same changes)

PR #1598, opened 2022-06-16
SWC-6170

nickgros:


src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -100,21 +117,9 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
   const { data: path } = useGetEntityPath(entityId)
   const projectId = path?.path[1]?.id
 
-  const { refetch } = useGetEntity<RequiredProperties<Dataset, 'items'>>(
-    entityId,
-    undefined,
-    {
-      enabled: !datasetToUpdate,
-      onSuccess: dataset => {
```
jay-hodgson, line 108: Ah, so this onSuccess is not run in the case of a cache hit.  Nice sleuthing

PR #1605, opened 2022-06-20
PORTALS-2168 - bump markdown-it-synapse

nickgros:


jay-hodgson (approved):
Need to fix merge conflicts, but approving update to markdown-it-synapse@1.1.10

PR #1606, opened 2022-06-21
SWC-5974

nickgros:
- Fix tooltip persisting on nav drawer

https://user-images.githubusercontent.com/17580037/175111670-8103c6b7-dbf1-4b48-9c7e-e3c524a7ac6f.mov

- Fix avatar not appearing in nav drawer
image

- Rewrite UserCard tests to not use Enzyme
- Use react-query to fetch UserBundle
- Various changes to reduce test errors and warnings

### DIFF
--- a/src/lib/utils/tooltip/Tooltip.tsx
+++ b/src/lib/utils/tooltip/Tooltip.tsx
@@ -10,12 +10,10 @@ export type TooltipProps = MuiTooltipProps
 
 const useTooltipStyle = makeStyles(theme => ({
   arrow: {
-    // gray-700
-    color: '#878E95',
+    color: theme.palette.common.black,
   },
   tooltip: {
-    // gray-700
-    backgroundColor: '#878E95',
+    backgroundColor: theme.palette.common.black,
     fontSize: '14px',
     borderRadius: '2px',
   },


src/__tests__/lib/containers/UserCardTests.test.tsx
```diff
@@ -1,226 +1,241 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import * as React from 'react'
-import { mount, shallow } from 'enzyme'
```
nickgros, line 2: Removed Enzyme

src/lib/containers/Avatar.tsx
```diff
@@ -51,44 +54,45 @@ export const Avatar: React.FunctionComponent<AvatarProps> = ({
...
+
+  const conditionalStyles: React.CSSProperties = hasProfileImage
+    ? {
+        backgroundImage: `url(${imageURL})`,
+      }
+    : { background: getColor(userProfile.userName) }
+
+  if (isLoadingAvatar) {
+    return <Skeleton className={sizeClass} variant="circle" />
+  }
```
nickgros, line 67: Skeleton for when Avatar is loading

src/lib/containers/Avatar.tsx
```diff
@@ -51,44 +54,45 @@ export const Avatar: React.FunctionComponent<AvatarProps> = ({
...
-            cursor: cursorStyle,
-          }}
-          className={`${sizeClass} SRC-centerContentInline`}
-        >
-          {userProfile.firstName &&
-            (userProfile.firstName[0] || userProfile.userName[0])}
-        </div>
-      )}
+      {showCardOnHover && <OverlayComponent />}
+      <div
```
nickgros, line 72: Just render one div and conditionally change props/content

src/lib/containers/UserCard.tsx
```diff
@@ -18,14 +19,14 @@ export type UserCardSize =
...
-  /** If set, the corresponding image will be shown for the user. */
-  preSignedURL?: string
   /** An alias that resolves the ownerId for the UserProfile. You must supply one of `userProfile`, `alias`, `ownerId` */
   alias?: string
   /** The unique ownerId of the UserProfile. You must supply one of `userProfile`, `alias`, `ownerId` */
   ownerId?: string
+  /** Whether or not to hide the user's Synapse email address */
+  hideEmail?: boolean
+  /** If set, the corresponding image will be shown for the user. */
+  preSignedURL?: string
```
nickgros, line 29: Just reordered so userProfile, alias, and ownerId are all next to each other, because one of those is needed.

src/lib/containers/UserCard.tsx
```diff
@@ -62,48 +63,67 @@ export const UserCard: React.FunctionComponent<UserCardProps> = (
...
-    principalId!,
-    {
-      enabled: !!principalId,
-    },
-  )
+  const { data: fetchedProfile } = useGetUserProfile(principalId!, {
+    enabled: !!principalId && !initialProfile,
+  })
+
+  const userProfile = initialProfile ?? fetchedProfile
```
nickgros, line 76: Only fetch a user profile if we weren't provided an initialProfile object (and have a principal ID)

Then, define `userProfile` as the initial profile, falling back to the fetched profile, if it exists.

src/lib/containers/UserCard.tsx
```diff
@@ -62,48 +63,67 @@ export const UserCard: React.FunctionComponent<UserCardProps> = (
...
-      }
+    let isCanceled = false
+    if (!initialPreSignedURL && (ownerId || userProfile?.ownerId)) {
+      SynapseClient.getProfilePicPreviewPresignedUrl(
+        (ownerId ?? userProfile?.ownerId)!,
+      ).then(result => {
+        if (!isCanceled) {
+          setFetchedPresignedUrl(result)
+        }
+      })
```
nickgros, line 87: If we weren't provided `initialPreSignedURL` and we have an ID, fetch the presigned URL for the preview from Synapse.

src/lib/utils/hooks/useOverlay.tsx
```diff
@@ -1,63 +1,104 @@
...
+function resetTimer(timer: NodeJS.Timeout | null) {
+  if (timer) {
+    clearTimeout(timer)
   }
 }
 
+const DEFAULT_DELAY_SHOW_MS = 250
+const DEFAULT_DELAY_HIDE_MS = 500
+
 export function useOverlay(
```
nickgros, line 20: For some reason, this re-renders like crazy for no reason. Doesn't seem to cause performance issues, but makes it hard to test. I added a bunch of `useCallback`s and now keep track of the mounted state to ensure we don't update state after unmounting. No major logic changes.

src/lib/utils/hooks/usePreFetchResource.ts
```diff
@@ -9,14 +19,16 @@ import { useEffect, useState } from 'react'
 export default function usePreFetchResource(
   preSignedURL?: string,
 ): string | undefined {
-  const [resourceURL, setResourceURL] = useState<string>()
+  const [resourceURL, setResourceURL] = useState<string | undefined>(undefined)
 
   useEffect(() => {
+    let isMounted = true
```
nickgros, line 25: Track mounted state

src/lib/containers/SynapseNavDrawer.tsx
```diff
@@ -158,73 +219,10 @@ export const SynapseNavDrawer: React.FunctionComponent<
     setSelectedItem(navItem)
   }
 
-  const handleDrawerClose = () => {
+  const handleDrawerClose = useCallback(() => {
```
nickgros, line 222:
```suggestion
  const handleDrawerClose = () => {
```

src/lib/containers/SynapseNavDrawer.tsx
```diff
@@ -328,6 +355,14 @@ export const SynapseNavDrawer: React.FunctionComponent<
...
         className={`SynapseNavContentDrawer`}
         onClose={handleDrawerClose}
+        ModalProps={{
+          BackdropProps: {
+            onClick: e => {
+              e.stopPropagation()
+              handleDrawerClose()
+            },
+          },
+        }}
```
nickgros, line 365:
```suggestion
```

src/lib/utils/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,31 @@
...
+    // gray-700
+    color: '#878E95',
+  },
+  tooltip: {
+    // gray-700
+    backgroundColor: '#878E95',
+    fontSize: '14px',
+    borderRadius: '2px',
+  },
+}))
```
nickgros, line 22: Styles pulled from Figma. Think we should wait to fix the references to use variables rather than hard-coded colors when we go to MUI v5

src/lib/containers/SynapseNavDrawer.tsx
```diff
@@ -305,21 +330,33 @@ export const SynapseNavDrawer: React.FunctionComponent<
               tooltip="Your Account"
               onClickOpenNavMenu={NavItem.PROFILE}
               additionalChildren={
-                <Avatar userProfile={currentUserProfile} avatarSize="SMALL" />
+                <UserCard
+                  userProfile={currentUserProfile}
+                  size="AVATAR"
+                  avatarSize="SMALL"
+                />
```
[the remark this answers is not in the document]
jay-hodgson: Agree, and I think the second option is sufficient

src/lib/containers/UserCard.tsx
jay-hodgson: Nice use of the cleanup function

src/lib/utils/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function Tooltip(props: TooltipProps) {
+  const classes = useTooltipStyle()
+  return (
+    <MuiTooltip classes={classes} arrow TransitionComponent={Fade} {...props}>
```
jay-hodgson, line 27: Why did you need to wrap the MuiTooltip?  Any idea why the react-tooltip was not behaving properly?
Will this tooltip look different than other tooltips in the application?

nickgros: I wrapped the MUI tooltip to ensure the provided style will always be included, and to give us the flexibility to change the implementation later while still using this component.

The tooltip will look different. I can change it to match the React Tooltip until we can change everything at once.

PR #1607, opened 2022-06-22
PORTALS-2222: refetch download statistics if files were added to the cart

jay-hodgson:


src/lib/containers/AddToDownloadListV2.tsx
```diff
@@ -15,12 +16,24 @@ const AddToDownloadListV2: React.FunctionComponent<
 > = props => {
   const { entityId, entityVersionNumber } = props
   const { accessToken } = useSynapseContext()
+  const { refetch } = useGetDownloadListStatistics({
```
jay-hodgson, line 19: Initially tried using the queryClient.invalidateQueries('downloadliststatsv2') (with various options), but there was still a significant lag time (about 10 seconds).
Even with the queryClient.setQueryData() (copying the old response data and manually incrementing the total file count) did not do the trick.  This seems to work, however.

nickgros: Did you do `queryClient.invalidateQueries('downloadliststatsv2')` or `queryClient.invalidateQueries(['downloadliststatsv2'])`? 

The query key must be an array, so only the latter would actually work

nickgros (approved):
See the comment but if you already tried that and it doesn't work, this is fine

src/lib/containers/AddToDownloadListV2.tsx
jay-hodgson: I was not properly setting as an array, and then I tricked myself in local testing (I made the change to use an array, but the browser was running the old code because I had not restarted the dev server!).  
Anyway, with the array of query keys (with single relevant key), the download count was still not updating in a timely manner in the SynapseNavDrawer.  I didn't test ShowDownloadV2 (which is what is used in the portals), but since they use the same strategy I would not expect thse to behave differently.

jay-hodgson: Nick noticed that I was setting exact to true during the invalidation, which caused it to be a no-op.  After fixing, the item is updated somewhere between 1-3 seconds.  I think this is acceptable performance.

https://user-images.githubusercontent.com/1864447/175127285-5815708e-5ccf-4858-a135-6d4a5ff26dd1.mov

PR #1608, opened 2022-06-22
PORTALS-2206

yjyoo773:


src/__tests__/lib/containers/QueryWrapper.test.tsx
```diff
@@ -74,7 +76,7 @@ describe('basic functionality', () => {
     expect(providedContext).toEqual(
       expect.objectContaining({
         data: syn16787123Json,
-        hasNextPage: true,
+        // hasNextPage: true,
```
nickgros, line 79:
```suggestion
```

src/lib/containers/QueryContext.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isFacetsAvailable: boolean
+  /**
+   * A facet may be "locked" so that it is not modifiable by the user, for example when showing only data relevant to a particular facet value on a Details Page.
+   * The value of a locked facet will result in a client-side modification of the result bundle data.
+   */
+  lockedFacet?: LockedFacet
+}
+
+export type PaginatedQueryContextType = QueryContextType & {
+  /** Navigates to a particular page */
```
nickgros, line 54: No preference for if you make this 0-indexed or 1-indexed, but either way it should be explicitly documented. Currently it's 1-indexed.
```suggestion
  /** Navigates to a particular page, where the first page has value 1 */
```

src/lib/containers/QueryContext.tsx
```diff
@@ -0,0 +1,125 @@
...
+  goToPage: (pageNum: number) => Promise<void>
+}
```
nickgros, line 56: Also, can we store the page size and page number in the context object instead of in TablePagination? That way 
* The values can be based on the current value of the query rather than UI state
* If some other component needs to access either value, it can
```suggestion
  /** The current page number, where page 1 is the first page. */
  currentPage: number
  /** Updates the page size */
  setPageSize: (pageSize: number) => void
  /** The current page size. */
  pageSize: number
}
```

You might be able to put pageSize in QueryContextType instead of PaginatedQueryContextType (infinite queries will have a page size too). But I'm not sure if that's useful

src/lib/containers/QueryWrapper.tsx
```diff
@@ -156,72 +77,13 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
-    }
-
-    await fetchPreviousPage()
-    setCurrentPage(currentPage - 1)
+  const goToPage = async (pageNum: number) => {
+    const lastQueryRequestDeepClone = getLastQueryRequest()
+    lastQueryRequestDeepClone.query.offset =
+      (pageNum - 1) * (lastQueryRequest.query.limit ?? DEFAULT_PAGE_SIZE)
+    executeQueryRequest(lastQueryRequestDeepClone)
   }
```
nickgros, line 85: As I mentioned in QueryContext, let's determine the page number here, so it's tied to query state rather than UI state.

Not sure if this math is correct, hopefully you get the idea.

```suggestion
  const pageSize = lastQueryRequest.query.limit ?? DEFAULT_PAGE_SIZE
  
  const currentPageNumber = Math.ceil((lastQueryRequestDeepClone.query.offset + pageSize) / pageSize)
  
  const goToPage = async (pageNum: number) => {
    const lastQueryRequestDeepClone = getLastQueryRequest()
    lastQueryRequestDeepClone.query.offset =
      (pageNum - 1) * pageSize
    executeQueryRequest(lastQueryRequestDeepClone)
  }
```

src/lib/containers/table/TablePagination.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useEffect, useState } from 'react'
+import { Pagination } from '@material-ui/lab'
+import { usePaginatedQueryContext } from '../QueryContext'
+import Typography from '../../utils/typography/Typography'
+
+export const TablePagination = () => {
+  const [pageSize, setPageSize] = useState(25)
+  const [page, setPage] = useState(1)
```
nickgros, line 8: After suggested changes, these would be pulled from context

src/lib/containers/table/TablePagination.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const handlePageSize = (event: React.ChangeEvent<{ value: unknown }>) => {
+    const value = event.target.value as number
+    setPageSize(value)
+  }
+
+  useEffect(() => {
+    const lastQueryRequestDeepClone = getLastQueryRequest()
+    lastQueryRequestDeepClone.query.offset = (page - 1) * pageSize
+    lastQueryRequestDeepClone.query.limit = pageSize
+    executeQueryRequest(lastQueryRequestDeepClone)
```
nickgros, line 28: This could be moved up to QueryWrapper, and then `setPageSize` could be passed in the PaginatedQueryContext

nickgros (changes requested):

src/lib/containers/QueryContext.tsx
```diff
@@ -33,15 +33,13 @@ export type QueryContextType = {
...
-  /** Returns true when loading a new page of query results */
-  isLoadingNewPage: boolean
   /** Returns true when loading a brand new query result bundle. Will not be true when just loading the next page of query results. */
   isLoadingNewBundle: boolean
   /** The error returned by the query request, if one is encountered */
   error: SynapseClientError | null
   /** The status of the asynchronous job. */
   asyncJobStatus?: AsynchronousJobStatus<QueryBundleRequest, QueryResultBundle>
-  /** Whether or not facets are available to be filtered upon based on the current data */
+  /** Whether or not facgit ets are available to be filtered upon based on the current data */
```
nickgros, line 42:
```suggestion
  /** Whether or not facets are available to be filtered upon based on the current data */
```

nickgros (approved):

PR #1609, opened 2022-06-22
PORTALS-2223: apply same fix from synapse.org scss to template styles in SRC

jay-hodgson:


nickgros (approved):

jay-hodgson:
Test failure in SubmissionPage, ignoring for this PR.

PR #1611, opened 2022-06-23
HasAccessV2

nickgros:
I noticed HasAccess has a lot of functionality we no longer use (most of it is now handled by DirectDownload) and decided this was easier than trying to rewrite comprehensive tests for HasAccess in React Testing Library.

Compared to HasAccess:
- Remove unused props and unused functionality
- Functional, uses hooks, react-query cache
- Tests written in React Testing Library


src/__tests__/lib/containers/HasAccessV2.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+ *  - Entity is not a FileEntity
+ *  - Entity does not have a dataFileHandleId
+ *
+ * Access Requirement states:
+ *  - No access requirements
+ *  - Meet all access requirements
+ *  - Do not meet all access requirements
+ *  - Do not meet all access requirements, and AR type is not supported by this component
+ */
+
```
nickgros, line 65: Cases are now tests, so this is redundant
```suggestion
```

src/lib/containers/HasAccessV2.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+  const { accessToken } = useSynapseContext()
+  const { data: entityBundle, error: entityFetchError } = useGetEntityBundle(
+    entityId,
+    {
+      includeEntity: true,
+      includePermissions: true,
+    },
+    parseInt(entityVersionNumber ?? ''),
+  )
```
nickgros, line 119: HasAccess v1 attempted to fetch the file handle to determine permission (and I think, for the original Download List, to determine if the file handle was external). Instead, just fetch the permission object from the entity bundle.

src/__tests__/lib/containers/HasAccessV2.test.tsx
```diff
@@ -0,0 +1,319 @@
...
+          objectId: entityId,
+        }
+
+      expect(onGetRestrictionInformation).toHaveBeenLastCalledWith(
+        expectedRestrictionInformationRequest,
+      )
+    })
+  })
+
+  describe('Opens the Access Requirements panel', () => {
```
jay-hodgson, line 314: Nice set of tests, with placeholders for other scenarios

src/lib/containers/HasAccessV2.tsx
```diff
@@ -0,0 +1,321 @@
...
+                This button has an svg inside of it which is problematic because more often than not clicking this button will 
+                instead click that svg. The event listener in the portals will break as a result.
+
+                Though the svg may get the actual click event, because of event bubbling this button will get its onClick called.
+                Once onClick is called we can manually dispatch an event off of this button. This does pose a problem, we end up in a 
+                infinite loop because this button keeps disptaching click events, so we can use the isTrusted to recognize if onClick was
+                triggered programmatically or by user click. Lastly, using { bubbles: true } ensures the event bubbles up to the document level.
+
+              */
+            const clickEvent = new MouseEvent('click', { bubbles: true })
```
jay-hodgson, line 304: Interesting, nice workaround here.  Is there a better way to trigger the sign-in modal on the portals side (instead of using SRC_SIGN_IN_CLASS)?

jay-hodgson (approved):

src/lib/containers/HasAccessV2.tsx
nickgros: Probably--this code and comment exists in HasAccess v1, so I didn't touch it.

I think it would make more sense to pass a callback called `onShowLogin` via `SynapseContext`. Each app using SRC could trigger showing the login screen however is most appropriate

PR #1619, opened 2022-06-28
Replace React-Styleguidist with Storybook, built with Vite

nickgros:
[Live Demo](https://nickgros.com/Synapse-React-Client/) - deployed via GHAction

* Replaces react-styleguidist with Storybook, and replaces each styleguidist markdown page with Storybook stories
* Drops react-scripts
* Adds Vite for bundling Storybook

With CRA/craco, getting Storybook to work with our library was going to require lots of custom webpack configuration. Additionally, 
* craco is no longer actively maintained
* our test configuration wasn't naturally compatible with CRA or craco
* we don't actually deploy a web application in this repository
So I decided that the cost of trying to keep using CRA was starting to outweigh the benefits of dropping it for our own configuration.

Rather than just ejecting from CRA and further configuring webpack, I decided to use [Vite](https://vitejs.dev/) because it seems like a promising technology, with large performance improvements over webpack (see [Why Vite](https://vitejs.dev/guide/why.html)).


.github/workflows/npmpublish.yml
```diff
@@ -32,10 +32,3 @@ jobs:
               ref: "refs/tags/v${{ env.PACKAGE_VERSION }}",
               sha: context.sha
             })
-      - name: Build docs
-        run: yarn build:docs
-      - name: Deploy docs to GitHub Pages
-        uses: JamesIves/github-pages-deploy-action@4.1.5
-        with:
-          branch: gh-pages
-          folder: styleguide
```
nickgros, line 41: Moved to new action

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,23 @@
+name: 'Publish Storybook'
+on:
+  push:
+    branches:
+      - develop
```
nickgros, line 5: Build and publish docs on pushes to `develop`. No reason to tie Storybook updates with tagged releases.

.storybook/main.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    const customStorybookConfig = defineConfig({
+      base,
+      plugins: [
+        svgr(),
+        viteExternalsPlugin({
+          react: 'React',
+          'react-dom': 'ReactDOM',
+          'sql-parser/browser/sql-parser': 'SQLParser',
```
nickgros, line 33:
```suggestion
```

jest.config.js
```diff
@@ -0,0 +1,26 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
```
nickgros, line 2: Since we dropped react-scripts, we now have to maintain our own config.

This is greatly improved over our previous setup, where we were passing some configuration options via the package.json config, and others via the command line, because react-scripts doesn't allow overriding all of the options we need to override.

src/__tests__/lib/containers/ErrorPage.test.tsx
```diff
@@ -12,7 +12,7 @@ describe('DirectDownload: basic functionality', () => {
   it('should render the correct content', () => {
     render(<ErrorPage {...propsNoAccess} />)
     // Should actually be an image, but our test platform doesn't currently load SVGs imported as React Components
-    screen.getByText('no-access.svg')
+    expect(screen.getByRole('img').getAttribute('title')).toBe('noAccess')
```
nickgros, line 15: Updated a handful of tests now that SVG importing is subtly different under this Vite configuration.

src/lib/containers/IconSvg.tsx
```diff
@@ -85,90 +85,93 @@ import Rat from '../assets/mui_components/Rat'
...
-  | 'sync'
-  | 'clipboardCheck'
-  | 'add'
-  | 'warning'
-  | 'circle'
-  | 'block'
-  | 'checkCircle'
-  | 'errorOutlined'
-  | 'phone'
+export const IconStrings = [
```
nickgros, line 88: Creating this type with an array, so we can import the array in a Story to show all icons.

src/lib/style/components/_all.scss
```diff
@@ -5,19 +5,19 @@
...
-  './entity_finder/all', './entity-badge', './download-list-v2',
-  './experimental-mode', './user-search-box', './entity-metadata',
-  './download-cart-page', './global-alert', './annotations', 'plotlywrapper',
-  './synapse-nav-drawer', './favorites-page', './skeletons', './toasts',
-  './user-profile-links', './blocking-loader', './transition', './popover',
-  './featured-tool-card', './row_renderers/observation-card',
-  './subsection-row-renderer', './dataset-editor', './fluid-modal',
-  './markdown-collapse', './discussion-search', './submission-page',
-  './governance-dashboard';
+  './query_filter/enum-facet-filter', './query_filter/facet-filter-header',
```
nickgros, line 8: `./query_filter/facet-filter-header` was imported inline, just making sure to import it via the full CSS build

src/lib/index.ts
```diff
@@ -1,67 +1,67 @@
-import 'regenerator-runtime/runtime'
```
nickgros, line 1: Just reorganized. Still using these polyfills

vite.config.ts
```diff
@@ -0,0 +1,52 @@
+import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfill'
+import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfill'
+import react from '@vitejs/plugin-react'
+import path from 'path'
+import { defineConfig } from 'vite'
+import svgr from 'vite-plugin-svgr'
+
+export default defineConfig({
```
nickgros, line 8: This config lets us run the deprecated demo app. Was useful to get Vite working with a basic demo of our library.

Likely will be worth keeping around. Eventually, Vite will likely use esbuild to create a bundle, at which point we can probably combine this config with the esbuild config.

package.json
```diff
@@ -88,27 +88,36 @@
     "tslib": "^2.3.1",
     "ua-parser-js": "^1.0.2",
     "universal-cookie": "^4.0.3",
-    "use-deep-compare-effect": "^1.3.1"
+    "use-deep-compare-effect": "^1.3.1",
+    "vite-plugin-svgr": "^2.2.0"
```
nickgros, line 92: Move to devDependencies

jay-hodgson (approved):
Assuming that the UMD works in SWC, approved!

PR #1620, opened 2022-06-29
QueryWrapper: Split fetching stats and rows

nickgros:
When we split QueryWrapper and InfiniteQueryWrapper, one optimization that we forgot to include in QueryWrapper is re-using the non-row data for a query (such as total count of results) when navigating between pages. Fetching data other than rows can be very expensive on the backend, so we should only request this data once. This should greatly improve performance when navigating between pages.

### DIFF
--- a/src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
+++ b/src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
@@ -5,6 +5,7 @@ import {
   UseInfiniteQueryOptions,
   useQuery,
   UseQueryOptions,
+  UseQueryResult,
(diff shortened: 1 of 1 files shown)


src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
```diff
@@ -49,7 +49,7 @@ export default function useGetQueryResultBundle(
   )
 }
 
-export function useGetQueryResultBundleWithAsyncStatus(
+function _useGetQueryResultBundleWithAsyncStatus(
```
nickgros, line 52: This function now just handles creating the query and making the request.

yjyoo773 (approved):

PR #1626, opened 2022-06-30
Attempt to fix AccessHistoryDashboard.test

nickgros:
Test has been sporadically failing across multiple builds


src/lib/containers/UserSearchBoxV2.tsx
```diff
@@ -147,7 +147,7 @@ const UserSearchBoxV2: React.FC<UserSearchBoxProps> = props => {
             ]
           : true
       }
-      id={htmlId}
+      inputId={inputId}
```
nickgros, line 150: We were using the wrong prop, so `getByLabel` didn't work before fixing this

yjyoo773 (approved):

PR #1627, opened 2022-06-30
Test suite improvements

nickgros:
SWC-6182

* Rewrite many tests using Enzyme
* Identify and fix tests with memory leaks using jest's `--detectLeaks` flag.
  * After replacing "leaking" polyfills in the setup files, 18 test suites were leaking
  * After fixing imports, test styles, etc, almost all memory leak warnings are gone. Only a couple of files fail intermittently when running --detectLeaks.
* Reorganize mocks to simplify MSW handlers


src/__tests__/lib/containers/CardContainer.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import React from 'react'
-import { SynapseConstants } from '../../../lib'
+import { SynapseConstants } from '../../../lib/utils'
```
nickgros, line 4: Changes like this usually resolved the memory leak warning on a test suite...not sure why

src/__tests__/lib/containers/HasAccess.test.tsx
```diff
@@ -1,318 +0,0 @@
-import { mount } from 'enzyme'
```
nickgros, line 1: Removed, since we have HasAccessV2

src/__tests__/lib/containers/GenericCard.test.tsx
```diff
@@ -26,6 +26,7 @@ import {
   ShowMore,
 } from '../../../lib/containers/row_renderers/utils'
 import IconSvg from '../../../lib/containers/IconSvg'
+import { server } from '../../../mocks/msw/server'
```
nickgros, line 29: Adding MSW also fixed a lot of "memory leaks". It might be that some components are rendering that hit the real Synapse backend, and the latency/type of failure causes an asynchronous issue.

src/lib/containers/HasAccess.tsx
```diff
@@ -1,517 +0,0 @@
-/**
```
nickgros, line 1: Removed because we have HasAccessV2

src/lib/containers/MarkdownSynapse.tsx
```diff
@@ -21,23 +23,39 @@ const TOC_CLASS = {
...
+declare const markdownitContainer: any
+declare const markdownitEmphasisAlt: any
+declare const markdownitInlineComments: any
+declare const markdownitBr: any
+declare const markdownitMath: any
+
+declare const markdownit: typeof MarkdownIt
+declare const sanitizeHtml: typeof SanitizeHtml
+
+type WidgetParams = {
```
nickgros, line 43: Tried to add some types, clean up some lint warnings.

src/lib/containers/evaluation_queues/CalendarWithIconFormGroup.tsx
```diff
@@ -18,10 +19,11 @@ export type CalendarWithIconFormGroupProps = {
 export const CalendarWithIconFormGroup: React.FunctionComponent<
   CalendarWithIconFormGroupProps
 > = ({ value, setterCallback, label, disabled = false, isValidDate }) => {
+  const [id] = useState(uniqueId('calendar-with-icon-form-group-'))
```
nickgros, line 22: Generate an ID to map label with input to improve accessibility.

src/lib/umd.index.ts
```diff
@@ -62,7 +61,6 @@ const SynapseComponents = {
   UserCard,
   StatisticsPlot,
   HasAccess,
-  HasAccessV2,
```
nickgros, line 65: Replace HasAccess with v2. Use the same exported name, so no need to update SWC.

src/lib/testutils/EnzymeHelpers.ts
```diff
@@ -10,7 +10,6 @@ export async function resolveAllPending(
 ) {
   await act(async (): Promise<any> => {
     await Promise.resolve(wrapper)
-    await new Promise(resolve => setImmediate(resolve))
```
nickgros, line 13: Lost `setImmediate` (which is nonstandard) when switching polyfills. Didn't seem to affect anything, and we're disinvesting from Enzyme, so I don't feel we need to look into this.

src/mocks/entity/index.ts
```diff
@@ -0,0 +1,12 @@
+import mockDatasetData from './mockDataset'
+import { MockEntityData } from './MockEntityData'
+import mockFileEntityData from './mockFileEntity'
+import mockProjectEntityData from './mockProject'
+
+const mockEntities: MockEntityData[] = [
+  mockFileEntityData,
+  mockProjectEntityData,
+  mockDatasetData,
+]
```
nickgros, line 10: Similar to what we already do for mock user profiles, construct a set of mock entities that MSW can pluck from by ID

src/mocks/entity/mockEntity.ts
```diff
@@ -1,218 +1,28 @@
...
+import { MOCK_USER_ID } from '../user/mock_user_profile'
+import mockFile from './mockFileEntity'
+import mockProject from './mockProject'
+
+const MOCK_PROJECT_ID = mockProject.id
+const mockProjectEntity = mockProject.entity
+const MOCK_FILE_NAME = mockFile.name
+const mockFileEntityHeader = mockFile.entityHeader
 
-export const MOCK_FILE_ENTITY_ID = 'syn123'
```
nickgros, line 28: Moved some of the mock data to individual files, where 1 file = 1 entity

src/mocks/msw/handlers/entityHandlers.ts
```diff
@@ -51,12 +45,13 @@ export const entityHandlers = [
...
-        if (requestBody.name === MOCK_FILE_NAME) {
-          response = mockFileEntity
-          status = 200
-        } else if (requestBody.name === MOCK_PROJECT_NAME) {
-          response = mockProjectEntity
+        const requestBody = req.body as Entity
+
+        const entityData = mockEntities.find(
+          entity => entity.name === requestBody.name,
+        )
```
nickgros, line 52: In a given handler, find the entity that matches by ID, and return the requested data.

src/setupTests.js
```diff
@@ -1,5 +1,4 @@
-import 'react-app-polyfill/ie11'
-import 'react-app-polyfill/stable'
+import 'whatwg-fetch'
```
nickgros, line 1: Both of the react-app-polyfills were detected as leaking by jest. We only needed a `fetch` polyfill.

package.json
```diff
@@ -190,6 +190,7 @@
     "prettier": "^2.6.2",
     "process": "^0.11.10",
     "react-select-event": "^5.3.0",
+    "react-test-renderer": "16.14.0",
```
nickgros, line 193: Used to make snapshots in MarkdownSynapse instead of comparing HTML strings

jay-hodgson (approved):

PR #1630, opened 2022-07-01
SWC-6177

yjyoo773:


yjyoo773:
Screen Shot 2022-07-01 at 10 02 48 AM

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -216,10 +212,20 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
           ...datasetToUpdate,
           items: items,
         }
+        displayToast(SAVE_THE_DATASET_TO_CONTINUE, 'info', {
+          title: toastMessageTitle,
+          primaryButtonConfig: {
+            text: 'Save changes to Draft',
+            onClick: () => mutation.mutate(updatedDataSet),
+          },
+        })
```
nickgros, line 226: I think `displayToast` should be called outside of the `setDatasetToUpdate` function. This creates the potential for the following:
* User adds item `a` to Dataset, toast is shown where `onClick` saves Dataset has item `a`
* User removes item `a` and adds item `b` to Dataset, toast is shown where `onClick` saves Dataset with  `b`
* User clicks Save button on first toast, dataset is saved with just `a`

Instead, let's move this out of the setter and have the onClick call `mutation.mutate(datasetToUpdate)`

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -231,19 +237,26 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
-      } removed from the Dataset`,
+      displayToast(SAVE_THE_DATASET_TO_CONTINUE, 'info', {
+        title: `${selectedIds.size} Item${
+          selectedIds.size === 1 ? '' : 's'
+        } removed from the Dataset`,
+        primaryButtonConfig: {
+          text: 'Save changes to Draft',
+          onClick: () => mutation.mutate(updatedDataSet),
+        },
+      })
```
nickgros, line 256: Same scenario as above, please move out of the setter function.

nickgros (changes requested):

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
yjyoo773: The reason I added `displayToast` inside `setDatasetToUpdate` was that when setting as `mutation.mutate(datasetToUpdate)` the state isn't updated when the button is clicked. Also, the toast title depends on the data within `setDatasetToUpdate`.

PR #1632, opened 2022-07-01
SWC-6177b

yjyoo773:


src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -216,10 +212,20 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
         }
+        displayToast(SAVE_THE_DATASET_TO_CONTINUE, 'info', {
+          title: toastMessageTitle,
+          primaryButtonConfig: {
+            text: 'Save changes to Draft',
+            onClick: () => mutation.mutate(updatedDataSet),
+          },
+        })
+
+        return updatedDataSet
```
yjyoo773, line 228: Added displayToast inside setDatasetToUpdate was that when setting as mutation.mutate(datasetToUpdate) the state isn't updated when the button is clicked. Also, the toast title depends on the data within setDatasetToUpdate.

nickgros: Could you do something like
```tsx
function fireUpdate() {
	mutation.mutate(datasetToUpdate)
}

function addItemsToDataset(itemsToAdd: Reference[]) {
	if (datasetToUpdate) {
		let toastMessageTitle = ''
		setDatasetToUpdate(datasetToUpdate => {
			// existing logic to update state
			// update toastMessageTitle as needed
		})
		displayToast(SAVE_THE_DATASET_TO_CONTINUE, 'info', {
			title: toastMessageTitle,
			primaryButtonConfig: {
				text: 'Save changes to Draft',
				onClick: fireUpdate,
			},
		})
	} // unchanged from here
}


```

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -187,24 +207,20 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
         if (updatedItems.length > 0) {
-          toastMessageTitle += ` and ${updatedItems.length} Item${
+          toastTitle += ` and ${updatedItems.length} Item${
             updatedItems.length === 1 ? '' : 's'
           } updated`
         } else {
           // if no items were updated, title = "X items(s) added" + " to Dataset"
-          toastMessageTitle += ` to Dataset`
+          toastTitle += ` to Dataset`
         }
```
nickgros, line 222: Just to clean up the code a bit, can this be pulled out into more straightforward functions, each with a single responsibility (and could also be tested individually)?

In my mind you could have 
* A function that takes two lists of dataset items and returns the counts of added, changed, and removed items
* A function that takes the counts of added, changed, and removed items, and returns the appropriate toast message

nickgros: With this change, the toast message shouldn't have to be stored in state.

nickgros (changes requested):
Actual behavior looks good, see comments on how this could be simplified

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -112,6 +115,23 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
+      displayToast(SAVE_THE_DATASET_TO_CONTINUE, 'info', {
+        title: toastMessageTitle,
+        primaryButtonConfig: {
+          text: 'Save changes to Draft',
+          onClick: () => mutation.mutate(datasetToUpdate),
+        },
+      })
+    }
+    setPreviousDatasetToUpdate(datasetToUpdate)
+  }, [setDatasetToUpdate])
```
nickgros, line 133: Looks like a typo-this should be the variable itself, not the setter
```suggestion
  }, [datasetToUpdate])
```

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -112,6 +115,23 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
     datasetToUpdate && datasetToUpdate.items.length === selectedIds.size
   )
 
+  useEffect(() => {
```
nickgros, line 118: I wonder if it would make sense to have a custom hook that encapsulates this behavior, looking something like

```tsx
function useOnChange<T>(
    obj: T,
    onChangeFn: (obj: T, prevObj: T) => void,
    comparator: (obj: T, prevObj: T) => boolean
  ) {
  const [previousObj, setPreviousObj] = useState(obj)
  useEffect(() => {
    if (!comparator(previousObj, obj)) {
      onChangeFn(obj, previousObj)
    }
    setPreviousObj(obj)
  }, [obj])
}
```

I don't think we need it now (I am not sure if we have this sort of behavior anywhere else right now).

What are your thoughts?

yjyoo773: I don't think I've seen this behavior else where either. Do you think this would be used in future cases?

nickgros: Not sure, let's just leave it for now

nickgros (changes requested):

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -162,60 +182,74 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
   })
 
+  function getDataSetDifference(
+    oldDataSet: DatasetItem[],
+    newDataSet: DatasetItem[],
+  ) {
+    const unchangedItems = oldDataSet.filter(
+      item => !newDataSet.find(newItem => newItem.entityId === item.entityId),
+    )
+    const updatedItems = newDataSet.filter(newItem =>
```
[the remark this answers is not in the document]
yjyoo773: The `updatedItems` was based on what was previously written which the toast message would count files as updated even if the version number hadn't change. However, I believe it does make more sense to include where the version number does change.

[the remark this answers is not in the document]
yjyoo773: Adding the condition to set updatedItems where the entityId matches but version number differs tends to delete items where the entityId and version number are identical.

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -162,60 +182,91 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
+        }
+        return results
+      },
+      { updatedItems: [], newItems: [] } as {
+        updatedItems: DatasetItem[]
+        newItems: DatasetItem[]
+      },
+    )
+
+    return { unchangedItems, updatedItems, newItems, deletedItems }
```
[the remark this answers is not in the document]
nickgros: Oh I see, maybe we do need the items per below

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -162,60 +182,91 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
+        ) {
+          unchangedItems.push(result)
+        }
+        return results
+      },
+      { updatedItems: [], newItems: [] } as {
+        updatedItems: DatasetItem[]
+        newItems: DatasetItem[]
+      },
+    )
```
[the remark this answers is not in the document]
nickgros: Can be done while preserving items too:
```js
const newItems, updatedItems, unchangedItems = []
for (const item of newDataSet) {
  const oldItem = oldDataset.find(old => old.entityId === item.entityId)
  if (oldItem) {
    if (oldItem.versionNumber === item.versionNumber) {
      unchangedItems.push(item)
    } else {
      changedItems.push(item)
    }
  } else {
    newItems.push(item)
  }
}
// If an old item was not unchanged or updated, then it was definitely removed.
const removedItems = oldDataSet.filter(old => !unchangedItems.find(item => item.entityId != old.entityId) && !updatedItems.find(item => item.entityId != old.entityId)
```

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -162,60 +182,91 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
-            ),
+        const refToDatasetItem = itemsToAdd.map(item => ({
+          entityId: item.targetId,
+          versionNumber: item.targetVersionNumber!,
+        }))
+        const { unchangedItems, updatedItems, newItems } = getDataSetDifference(
+          datasetToUpdate.items,
+          refToDatasetItem,
         )
+        const items = [...unchangedItems, ...updatedItems, ...newItems]
```
nickgros, line 268: I see, we do need items, rather than just counts, because this was refactored to use the same logic 👍

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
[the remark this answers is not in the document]
yjyoo773: This deletes existing items that weren't selected from the EntityFinderModal. It doesn't include all of unchangedItems.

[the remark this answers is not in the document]
yjyoo773: Yes, that would be a more appropriate naming

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -160,60 +180,91 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
+        }
+        if (
+          oldDataSet.find(
+            old =>
+              old.entityId === result.entityId &&
+              old.versionNumber === result.versionNumber,
+          )
+        ) {
+          unchangedItems.push(result)
+        }
```
nickgros, line 214: These are all mutually exclusive cases so instead of going through oldDataSet 3 times, can you just do this?

```suggestion
        const oldItem = oldDataSet.find(old => old.entityId === result.entityId)
        if (oldItem) {
          if (result.versionNumber !== old.versionNumber) {
            results['updatedItems'].push(result)
          } else {
            unchangedItems.push(result)
          }
        } else {
          results['newItems'].push(result)
	}
```

src/lib/containers/table/datasets/DatasetItemsEditor.tsx
```diff
@@ -160,60 +180,91 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
     }
   })
 
+  function getDataSetDifference(
+    oldDataSet: EntityRef[],
+    newDataSet: EntityRef[],
```
nickgros, line 185: Please rename newDataSet to changedItems or something similar to indicate what's happening

src/__tests__/lib/containers/table/datasets/DatasetItemsEditor.test.tsx
```diff
@@ -585,4 +590,77 @@ describe('Dataset Items Editor tests', () => {
...
+      // Add identical item to existing dataset with different version
+      mockEntityFinderToAddItems([mockFileReference])
+      addItemsViaEntityFinder()
+
+      // Verify toast shows no item has been added and 1 has updated
+      expect(mockToastFn).toBeCalledWith(
+        expect.anything(),
+        'info',
+        expect.objectContaining({
+          title: '0 Items added and 1 Item updated to Dataset',
```
nickgros, line 624: "updated to Dataset" is a little odd, I think we currently drop it in that case. Recommend one of these two:
```suggestion
          title: '0 Items added and 1 Item updated',
```

```suggestion
          title: '0 Items added to and 1 Item updated in the Dataset',
```

nickgros (approved):

PR #1636, opened 2022-07-05
PORTALS-2004

jay-hodgson:


src/lib/containers/GenericCard.tsx
```diff
@@ -101,6 +103,32 @@ export const getCutoff = (summary: string) => {
   return { previewText }
 }
 
+export const getFileHandleAssociation = (
```
jay-hodgson, line 105: Refactored for reuse in Programs component

jay-hodgson:
Screen Shot 2022-07-05 at 12 14 10 PM

jay-hodgson:
Screen Shot 2022-07-05 at 12 23 50 PM

src/lib/containers/home_page/programs/Programs.Desktop.tsx
```diff
@@ -19,7 +18,7 @@ export default function ProgramsDesktop({
         style={{ backgroundColor: color.replace(')', ',.05)') }}
       >
         <div className="Programs__Card__header__icon">
-          <Icon type={iconValue} iconOptions={iconOptions} />
+          <ImageFileHandle fileHandleAssociation={fileHandleAssociation!} />
```
nickgros, line 21: Can we handle the undefined case instead of asserting that it's defined? Not sure what happens if undefined gets passed into ImageFileHandle but it seems like it may be possible if there's a config error

nickgros: Also might indicate a loading state? Not sure

src/lib/containers/home_page/programs/Programs.tsx
```diff
@@ -78,17 +84,21 @@ export const Programs: React.FC<ProgramsProps> = (props: ProgramsProps) => {
...
         const link = values[linkColumnIndex]
-        const iconValue = values[iconColumnIndex]
+        const iconFileHandleIdValue = values[iconColumnIndex]
         const matchValue = values[matchColIndex]
 
+        const fileHandleAssociation = getFileHandleAssociation(
+          entity.data,
+          iconFileHandleIdValue,
+          `syn${el.rowId}`,
+        )
```
nickgros, line 94: Not sure if inline calls to getFileHandleAssociation need to be memoized. My concern is that each render will create an new fileHandleAssociation object, causing ImageFileHandle to re-render and re-fetch the image.

If you didn't see this happen locally, then we probably don't need to worry about it.

nickgros (approved):

PR #1639, opened 2022-07-05
SWC-6131

yjyoo773:


jay-hodgson:
This looks great.  Could you expand the EntityFinder.test.tsx to include a new test?
"Updates the search button text when only table types are selectable", with selectableTypes set to an array with many table types?

jay-hodgson (approved):

PR #1640, opened 2022-07-06
PORTALS-2229: pass along table configuration to SynapseTable in Stand…

jay-hodgson:
…aloneQueryWrapper, and add a cards icon


src/lib/containers/table/StandaloneQueryWrapper.stories.tsx
```diff
@@ -15,6 +15,12 @@ export const WithoutTopLevelControls = Template.bind({})
 WithoutTopLevelControls.args = {
   title: 'Data Files',
   sql: `SELECT * FROM syn21994970.5 where dhPortalIndex = 'TRUE' LIMIT 10`,
+  columnLinks: [
```
jay-hodgson, line 18: This is to exercise the change to StandaloneQueryWrapper.  If columnLinks is passed to SynapseTable, then the studyOrProject column is processed as markdown (it contains a link, in markdown formatting).  If columnLinks is not passed to SynapseTable, then the column shows the raw value.

yjyoo773 (approved):

PR #1653, opened 2022-07-12
SWC-6207: Update Dataset icon, and add DatasetCollection

jay-hodgson:


src/lib/utils/SynapseClient.ts
```diff
@@ -1374,10 +1374,10 @@ export const setAccessTokenCookie = async (
       // See - https://github.com/reactivestack/cookies/issues/189
       await delay(100)
     } else {
-      // set's cookie in session storage
+      // sets cookie
```
jay-hodgson, line 1377: For local testing only (not accessed in a production environment).

jay-hodgson:
New Dataset and Dataset Collection icons (and add support for DS collection in EntityIcon):
Screen Shot 2022-07-12 at 11 03 45 AM

src/lib/containers/IconSvg.tsx
```diff
@@ -374,7 +377,9 @@ const getIcon = (options: IconSvgOptions) => {
     case 'discussion':
       return <QuestionAnswerTwoTone style={muiSvgStyle} />
     case 'dataset':
-      return <GridOnTwoTone style={muiSvgStyle} />
```
yjyoo773, line 377: Is the GridOnTwoTone being used elsewhere or can we remove it from import?

jay-hodgson: Good call, probably not used.  I'll check

yjyoo773 (approved):

PR #1656, opened 2022-07-13
SWC-6205

nickgros:
https://user-images.githubusercontent.com/17580037/178755079-e7736b85-a018-4b87-b6bf-fa7a55860fa5.mov

* When in Entity Finder dual-pane view, toggle the selected tree to be open when navigating from the right pane.
* Add an error boundary to EntityDetailsList, which contains all of the API calls that result in populating in the table portion of the dual-pane view.

### DIFF
--- a/src/lib/containers/entity_finder/details/EntityDetailsList.tsx
+++ b/src/lib/containers/entity_finder/details/EntityDetailsList.tsx
@@ -16,6 +16,7 @@ import { SearchDetails } from './configurations/SearchDetails'
 import { getIsAllSelectedFromInfiniteList } from '../../../utils/hooks/useGetIsAllSelectedInfiniteList'
 import { DetailsView } from './view/DetailsView'
 import { EntityTreeContainer } from '../tree/EntityTree'
+import { SynapseErrorBoundary } from '../../ErrorBanner'
 
 export enum EntityDetailsListDataConfigurationType {
   HEADER_LIST, // simply displays one or more entity headers. incompatible with pagination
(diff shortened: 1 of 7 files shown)


src/lib/containers/entity_finder/details/EntityDetailsList.tsx
```diff
@@ -141,5 +142,5 @@ export const EntityDetailsList: React.FunctionComponent<
     setComponent(getComponentFromConfiguration(configuration))
   }, [configuration, sharedProps])
 
-  return component
+  return <SynapseErrorBoundary>{component}</SynapseErrorBoundary>
```
nickgros, line 145: There are edge cases that just break the right pane of the entity finder. Don't need to fix them today (they are hard states to get into, e.g. navigating to a folder where you don't have VIEW access). But this will make them easier to deal with in development.

src/lib/containers/entity_finder/tree/VirtualizedTree.tsx
```diff
@@ -255,6 +249,28 @@ export function Node(
...
+    if (
+      treeNodeType === EntityTreeNodeType.DUAL_PANE &&
+      isSelected &&
+      !isOpen
+    ) {
+      toggleExpand()
+    }
+    // Intentionally only toggle the expanded state when isSelected changes, otherwise the node cannot be un-expanded
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isSelected])
```
nickgros, line 267: This is new. If `isSelected` changes, then expand the node.

jay-hodgson (approved):

PR #1661, opened 2022-07-13
SWC-6196: absolute positioning to place on the right, and move up a c…

jay-hodgson:
…ouple of pixels


jay-hodgson:
Screen Shot 2022-07-13 at 2 37 19 PM

Screen Shot 2022-07-13 at 2 34 34 PM

PR #1663, opened 2022-07-14
SWC-6194

nickgros:


src/lib/style/components/_governance-dashboard.scss
```diff
@@ -14,14 +14,14 @@
     justify-content: flex-start;
     box-shadow: 0px 3px 10px rgba(0, 0, 0, 0.05);
     .Tab {
-      @include SrcMixins.ignoreLink();
```
nickgros, line 17: This mixin gets overridden by more specific styles in SWC (like `#rootPanel a`).

Applying the class `ignoreLink` ends up setting color to `unset`, which is also not desired. Unfortunately the easiest thing to do is to use !important to remove the underline.

PR #1672, opened 2022-07-18
SWC-6240: fix for queries that use a column alias

jay-hodgson:


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -846,6 +846,7 @@ export default class SynapseTable extends React.Component<
           const isLockedFacetColumn =
             column.name.toLowerCase() === lockedFacet?.facet?.toLowerCase() // used in details page to disable filter the column
           const isEntityIDColumn =
+            columnModel &&
```
jay-hodgson, line 849: Oops

jay-hodgson, line 849: This is the bug fix.  Verify that we found a columnModel before looking at the CM name and type.

PR #1679, opened 2022-07-21
Fix flaky tests, AccessTokenPage styling, Storybook auth issues

nickgros:
Failure in AccessHistoryDashboard: https://app.travis-ci.com/github/Sage-Bionetworks/Synapse-React-Client/builds/253426839

Failure in AccessTokenPage: https://app.travis-ci.com/github/Sage-Bionetworks/Synapse-React-Client/builds/253472852

Also made minor updates to styling in AccessTokenPage

Made authentication token acquisition more robust in storybook

### DIFF
--- a/src/__tests__/lib/containers/dataaccess/AccessHistoryDashboard.test.tsx
+++ b/src/__tests__/lib/containers/dataaccess/AccessHistoryDashboard.test.tsx
@@ -179,7 +179,7 @@ describe('AccessHistoryDashboard tests', () => {
     )
 
     userEvent.type(arNameInput, mockAccessRequirement.name)
-    selectEvent.select(
+    await selectEvent.select(
       arNameInput,
       getOptionLabel(mockAccessRequirement.id, mockAccessRequirement.name),
     )
--- a/src/lib/containers/CopyToClipboardInput.tsx
+++ b/src/lib/containers/CopyToClipboardInput.tsx
@@ -45,9 +45,12 @@ export const CopyToClipboardInput: React.FunctionComponent<
         show={showModal}
         autohide={true}
       ></ToastMessage>
-      <div className="SRC-copyToClipboardInputContainer" ref={ref}>
+      <div
+        className="SRC-copyToClipboardInputContainer SRC-marginBottomTop"
+        ref={ref}
+      >
         <FormControl
-          className="SRC-marginBottomTop SRC-copyToClipboardInput"
+          className="SRC-copyToClipboardInput"
           type="text"
           style={{
             width: inputWidth,
--- a/src/lib/containers/StyleGuidistComponentWrapper.tsx
+++ b/src/lib/containers/StyleGuidistComponentWrapper.tsx
@@ -15,25 +15,27 @@ import {
 export async function sessionChangeHandler() {
   detectSSOCode()
   const accessToken = await getAccessTokenFromCookie()
-  global.accessToken = accessToken
   const profile = await getUserProfile(accessToken)
-  global.currentUserProfile = profile
+  let date
   if (accessToken) {
     getAuthenticatedOn(accessToken).then(authenticatedOn => {
-      const date = moment(authenticatedOn.authenticatedOn).format('L LT')
+      date = moment(authenticatedOn.authenticatedOn).format('L LT')
     })
   }
-  return profile
+  return { accessToken, profile, authenticatedOn: date }
 }
 
 export const StyleGuidistComponentWrapper: React.FC = props => {
+  const [accessToken, setAccessToken] = React.useState(undefined)
   useEffect(() => {
-    sessionChangeHandler()
-  }, [])
+    sessionChangeHandler().then(data => {
+      setAccessToken(data.accessToken)
+    })
+  })
   return (
     <SynapseContextProvider
       synapseContext={{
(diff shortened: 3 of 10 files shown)


src/__tests__/lib/containers/dataaccess/AccessHistoryDashboard.test.tsx
```diff
@@ -179,7 +179,7 @@ describe('AccessHistoryDashboard tests', () => {
     )
 
     userEvent.type(arNameInput, mockAccessRequirement.name)
-    selectEvent.select(
+    await selectEvent.select(
       arNameInput,
       getOptionLabel(mockAccessRequirement.id, mockAccessRequirement.name),
     )
```
nickgros, line 185: The `react-select` element asynchronously loads the options from the API, so we should `await` the select event.

src/lib/containers/CopyToClipboardInput.tsx
```diff
@@ -45,9 +45,12 @@ export const CopyToClipboardInput: React.FunctionComponent<
...
         autohide={true}
       ></ToastMessage>
-      <div className="SRC-copyToClipboardInputContainer" ref={ref}>
+      <div
+        className="SRC-copyToClipboardInputContainer SRC-marginBottomTop"
+        ref={ref}
+      >
         <FormControl
-          className="SRC-marginBottomTop SRC-copyToClipboardInput"
+          className="SRC-copyToClipboardInput"
```
nickgros, line 53: Fix appearance since this is now wrapped in bootstrap-4-backport

src/lib/containers/StyleGuidistComponentWrapper.tsx
```diff
@@ -15,25 +15,27 @@ import {
...
 
 export const StyleGuidistComponentWrapper: React.FC = props => {
+  const [accessToken, setAccessToken] = React.useState(undefined)
   useEffect(() => {
-    sessionChangeHandler()
-  }, [])
+    sessionChangeHandler().then(data => {
+      setAccessToken(data.accessToken)
+    })
+  })
```
nickgros, line 34: Since we're pulling a cookie instead of inheriting a prop, we can't keep track of when the access token changes. So, let's run this check after every render.

Additionally, store it in state instead of in a global variable, so React can re-render context if it changes.

src/lib/containers/personal_access_token/AccessTokenPage.tsx
```diff
@@ -18,53 +15,23 @@ export const AccessTokenPage: React.FunctionComponent<AccessTokenPageProps> = ({
   title,
   body,
 }: AccessTokenPageProps) => {
-  const { accessToken } = useSynapseContext()
-  const [isLoading, setIsLoading] = useState(false)
-
   const [showCreateTokenModal, setShowCreateTokenModal] = useState(false)
 
   const {
```
nickgros, line 20: This test suite was failing with logs after tests were done. Since those are often caused by state updates triggered after fetch, let's just replace this with a react-query hook, which handles that scenario internally. Also has the benefit of cleaning up a lot of this code. Test suite passed without any changes.

src/lib/containers/personal_access_token/AccessTokenPage.tsx
```diff
@@ -75,7 +42,7 @@ export const AccessTokenPage: React.FunctionComponent<AccessTokenPageProps> = ({
         </div>
         <div className="PersonalAccessTokenPage__Header__CreateButton">
           <Button
-            variant="primary"
+            variant="sds-primary"
```
nickgros, line 45: Worked in some style updates as well while I was here

src/lib/containers/synapse_form_wrapper/SynapseFormWrapper.stories.tsx
```diff
@@ -4,7 +4,7 @@ import SynapseFormWrapper from './SynapseFormWrapper'
 import { SynapseContextConsumer } from '../../utils/SynapseContext'
 
 export default {
-  title: 'Synapse/AccessTokenPage',
+  title: 'Synapse/SynapseFormWrapper',
```
nickgros, line 7: Wrong title caused these to collide. Discovered that SynapseFormWrapper isn't working in Storybook since we're missing some polyfills. Will create follow up ticket.

src/lib/utils/SynapseClient.ts
```diff
@@ -1403,7 +1403,7 @@ export const setAccessTokenCookie = async (
  */
 export const getAccessTokenFromCookie = async () => {
   if (IS_OUTSIDE_SYNAPSE_ORG) {
-    return cookies.get(ACCESS_TOKEN_COOKIE_KEY)
+    return Promise.resolve(cookies.get(ACCESS_TOKEN_COOKIE_KEY) as string)
```
nickgros, line 1406: This function should always return a Promise.

src/lib/containers/synapse_form_wrapper/SynapseFormWrapper.stories.tsx
nickgros: https://sagebionetworks.jira.com/browse/SWC-6245

jay-hodgson (approved):

PR #1687, opened 2022-07-26
SWC-6217

yjyoo773:


yjyoo773:
Screen Shot 2022-07-26 at 10 20 37 AM
Creating a new client
Screen Shot 2022-07-26 at 10 20 52 AM
Editing/Deleting a client
Screen Shot 2022-07-26 at 10 21 04 AM
Confirmation when deleting / editing redirect uri

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -0,0 +1,280 @@
...
+                value={clientUri}
+              />
+            </Col>
+          </Row>
+          <Row>
+            {!isEdit && (
+              <>
+                <Col lg={6} md={6} sm={12} xs={12}>
+                  <Form.Group className="required">
+                    <Form.Label>Redirect URI(s)</Form.Label>
```
nickgros, line 169: May be worth adding a tooltip or HelpPopover to indicate that the URIs should be comma separated.

I'm not 100% sure comma-separated is the right control--Is it possible that a URI contains a comma?

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -1,16 +1,27 @@
...
 import { SynapseClient } from '../../..'
+import { displayToast } from '../../../../containers/ToastMessage'
 import { SynapseClientError } from '../../../SynapseClientError'
 import { useSynapseContext } from '../../../SynapseContext'
-import { OAuthClientList } from '../../../synapseTypes/OAuthClient'
+import { OAuthClient, OAuthClientList } from '../../../synapseTypes/OAuthClient'
+
+const oAuthQueryKeys = {
+  all: (accessToken: string) => ['oAuthClient', accessToken],
+  edit: (accessToken: string, id: string) => ['oAuthClient', accessToken, id],
```
nickgros, line 16: Do we ever fetch an individual OAuth Client? Seems like this query key is unnecessary

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -19,3 +30,67 @@ export function useGetOAuthClientInfinite(
...
+    (clientId: string) =>
+      SynapseClient.deleteOAuthClient(clientId, accessToken!),
+    {
+      ...options,
+      onSuccess: async (updatedClient, clientId, ctx) => {
+        await queryClient.invalidateQueries(oAuthQueryKeys.all(accessToken!))
+        queryClient.setQueryData(
+          oAuthQueryKeys.edit(accessToken!, clientId),
+          updatedClient,
+        )
```
nickgros, line 50: If above comment is true, this shouldn't be necessary

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -19,3 +30,67 @@ export function useGetOAuthClientInfinite(
...
+        if (options?.onSuccess) {
+          await options.onSuccess(updatedClient, clientId, ctx)
+        }
+      },
+    },
+  )
+}
+
+export function useMutateOAuthClient(
```
nickgros, line 59: I think it might be cleaner to split this into two hooks, `useCreateOAuthClient` and `useEditOAuthClient`. I think this is the first place where we're using mutations for a "create" action, so I'm definitely open to hearing why this might be better, if you think so 🙂

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -19,3 +30,67 @@ export function useGetOAuthClientInfinite(
...
+  }
+
+  return useMutation(update, {
+    onSuccess: async (updatedClient, variables, ctx) => {
+      const clientId = updatedClient.client_id
+      queryClient.invalidateQueries(oAuthQueryKeys.all(accessToken!))
+      queryClient.setQueryData(
+        oAuthQueryKeys.edit(accessToken!, clientId!),
+        updatedClient,
+      )
```
nickgros, line 87: Ditto above

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -19,3 +30,67 @@ export function useGetOAuthClientInfinite(
...
+      if (options?.onSuccess) {
+        await options.onSuccess(updatedClient, variables.client, ctx)
+      }
+    },
+    onError: (err: any) => {
+      displayToast(err.reason as string, 'danger')
+    },
```
nickgros, line 94: This level should just manage data. Let the caller decide how to present errors.
```suggestion
```

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -0,0 +1,280 @@
...
+      <ConfirmModal
+        onClose={() => {
+          setIsShowingConfirmModal(false)
+          setIsDelete(false)
+          setSelectedClient(undefined)
+        }}
+        isDelete={isDelete}
+        showModal={isShowingConfirmModal}
+        client={isDelete ? client! : updatedClient!}
+      />
```
nickgros, line 277: IMO the design for this is unlike anything else we have in the product, so we should disregard it. Recommend just using the `WarningModal` component that we already have.

src/lib/containers/oauth/OAuthConfirmationModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+import {
+  useDeleteOAuthClient,
+  useMutateOAuthClient,
+} from '../../utils/hooks/SynapseAPI'
+import { OAuthClient } from '../../utils/synapseTypes/OAuthClient'
+import Typography from '../../utils/typography/Typography'
+import { displayToast } from '../ToastMessage'
+import { AlertIcon } from '../../assets/icons/Alert'
+
+export type ConfirmModalProps = {
```
nickgros, line 12: Prefer to use WarningModal instead of this custom one-off component, see comment on CreateOAuthClient.tsx

src/lib/style/components/_oauth-editor.scss
```diff
@@ -0,0 +1,34 @@
+@use '../abstracts/variables' as SRC;
+@use 'sass:map';
+
+.OAuthEditor {
+  td {
+    button {
+      color: #2b688f;
+    }
+  }
+}
```
nickgros, line 10: Suggest deleting in favor of using Bootstrap button, something like
```tsx
<Button variant="outline" size="sm" >
	EDIT
</Button>
```

```suggestion
```

src/lib/containers/oauth/OAuthManagement.tsx
```diff
@@ -33,7 +49,17 @@ export const OAuthManagement: React.FunctionComponent = () => {
...
+                <td>
+                  <button
+                    onClick={() => {
+                      setIsShowingCreateClientModal(true)
+                      setSelectedClient(item)
+                      setIsEdit(true)
+                    }}
+                  >
+                    EDIT
+                  </button>
```
nickgros, line 61: Just for style consistency with other table components:

```tsx
```
```suggestion
                  <Button
                  variant="outline"
                  size="sm"
                    onClick={() => {
                      setIsShowingCreateClientModal(true)
                      setSelectedClient(item)
                      setIsEdit(true)
                    }}
                  >
                    EDIT
                  </Button>
```

src/lib/containers/oauth/OAuthManagement.tsx
```diff
@@ -33,7 +49,17 @@ export const OAuthManagement: React.FunctionComponent = () => {
                   {item.verified ? 'Yes' : 'SUBMIT_VERIFICATION_PLACE_HOLDER'}
                 </td>
                 <td>GENERATE_PLACEHOLDER</td>
-                <td>ACTIONS_PLACEHOLDER</td>
+                <td>
+                  <button
+                    onClick={() => {
+                      setIsShowingCreateClientModal(true)
+                      setSelectedClient(item)
+                      setIsEdit(true)
```
nickgros, line 57: I think these will all get batched into one update, but updating the data before showing the modal is a little safer 
```suggestion
                      setSelectedClient(item)
                      setIsEdit(true)
                      setIsShowingCreateClientModal(true)
```

src/lib/containers/oauth/OAuthManagement.tsx
```diff
@@ -1,16 +1,32 @@
...
 import { useGetOAuthClientInfinite } from '../../utils/hooks/SynapseAPI'
+import { CreateOAuthModal } from './CreateOAuthClient'
+import { OAuthClient } from '../../utils/synapseTypes/OAuthClient'
 
 export const OAuthManagement: React.FunctionComponent = () => {
+  const [isShowingCreateClientModal, setIsShowingCreateClientModal] =
+    useState(false)
+  const [isEdit, setIsEdit] = useState<boolean>(false)
+  const [selectedClient, setSelectedClient] = useState<OAuthClient>()
+  const [isShowingConfirmModal, setIsShowingConfirmModal] = useState(false)
```
nickgros, line 15: Why control the confirm modal's state here instead of in the CreateOAuthClientModal?

src/__tests__/lib/containers/oauth/CreateOAuthClient.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+        )}/auth/v1/oauth2/client`,
+        async (req, res, ctx) => {
+          return res(ctx.status(200), ctx.json(req.body))
+        },
+      ),
+    )
+    renderComponent()
+    const inputName = (await screen.findAllByRole('textbox'))[0]
+    const inputHomePage = (await screen.findAllByRole('textbox'))[1]
+    const inputRedirectURI = (await screen.findAllByRole('textbox'))[2]
```
nickgros, line 76: Include the label in the query, either with `findByRole('textbox', {name: 'Client Name'})` or `findByLabelText("Client Name")`, etc., for readability/accessibility.

src/__tests__/lib/containers/oauth/CreateOAuthClient.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+    const saveButton = screen.queryByRole('button', { name: 'Save' })
```
nickgros, line 78: getByRole will throw an error if the button doesn't exist, which seems preferable here.
```suggestion
    const saveButton = screen.getByRole('button', { name: 'Save' })
```

src/__tests__/lib/containers/oauth/CreateOAuthClient.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+    renderComponent({
+      ...defaultProps,
+      setIsShowingConfirmModal: showModal,
+      isEdit: true,
+      client: mockClient,
+    })
+    const inputName = (await screen.findAllByRole('textbox'))[0]
+    const saveButton = screen.queryByRole('button', { name: 'Save' })
+    fireEvent.change(inputName, { target: { value: 'rename' } })
```
nickgros, line 133: Prefer userEvent.type

src/lib/containers/oauth/OAuthConfirmationModal.tsx
nickgros: Kind of unusual to have all this much logic in the modal when it could be captured in handlers passed to the modal

src/__tests__/lib/containers/oauth/CreateOAuthClient.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  setIsShowingConfirmModal: jest.fn(),
+}
+
+function renderComponent(props: CreateOAuthModalProps = defaultProps) {
+  render(<CreateOAuthModal {...props} />, {
+    wrapper: createWrapper(),
+  })
+}
+
+describe('Create 0Auth Client', () => {
```
nickgros, line 47:
```suggestion
describe('Create OAuth Client', () => {
```

src/__tests__/lib/containers/oauth/CreateOAuthClient.test.tsx
```diff
@@ -0,0 +1,171 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
```
nickgros, line 1: Test logic looks great, left some notes on how to improve style/queries

nickgros (changes requested):

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
yjyoo773: This was something I saw when I was working on `mtb` which used a hook for creating and updating cases along with others. I thought this would look "cleaner" since it would be less redundant when importing and calling it for `CreateOAuthClient.tsx`.

src/lib/containers/oauth/CreateOAuthClient.tsx
yjyoo773: It is technically possible for a URI to contain a comma but is considered bad practice. Do you think we should still go with comma separation?

src/lib/containers/oauth/OAuthManagement.tsx
yjyoo773: This was from an error I got while running tests. When testing for OAuthManagement I constantly got a type error of `useMutateOAuthClient` from `CreateOAuthClient.tsx` not being a function. I ended up having the component not render unless either one of the modal's render which made me put the confirm modal's state in OAuthManagement. It was a hacky way of solving the problem and am open to different solutions

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -0,0 +1,280 @@
...
+            </Typography>
+          </Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <Typography variant="body1">
+            To protect you and your users, your consent screen and application
+            will need to be verified by Sage Bionetworks. Before your consent
+            screen and application are verified by Sage Bionetworks, you can
+            still test your application with limitations. Learn more about how
+            your app will behave before it's verified.
```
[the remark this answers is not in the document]
yjyoo773: There should be a link to it. At least in the design but also the link is not provided in the design.

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -6,22 +6,16 @@ import {
...
 import { SynapseClient } from '../../..'
-import { displayToast } from '../../../../containers/ToastMessage'
 import { SynapseClientError } from '../../../SynapseClientError'
 import { useSynapseContext } from '../../../SynapseContext'
 import { OAuthClient, OAuthClientList } from '../../../synapseTypes/OAuthClient'
 
-const oAuthQueryKeys = {
-  all: (accessToken: string) => ['oAuthClient', accessToken],
-  edit: (accessToken: string, id: string) => ['oAuthClient', accessToken, id],
-}
```
nickgros: Sorry if I was unclear, IMO it's ok to keep the `all` key (consider renaming it to `list`, since it's only used for listing OAuth Clients for now) and then invalidate that key. Just don't need the `edit` key, since it's not used.

src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -43,11 +37,7 @@ export function useDeleteOAuthClient(
     {
       ...options,
       onSuccess: async (updatedClient, clientId, ctx) => {
-        await queryClient.invalidateQueries(oAuthQueryKeys.all(accessToken!))
-        queryClient.setQueryData(
-          oAuthQueryKeys.edit(accessToken!, clientId),
-          updatedClient,
-        )
+        await queryClient.invalidateQueries()
```
nickgros, line 40: See above comment. Just invalidate the `all`/`list` key. I think invalidating everything is more heavy handed than you need to be.

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -67,11 +76,27 @@ export const CreateOAuthModal: React.FunctionComponent<
...
     onSuccess: () => {
       displayToast(`Successfully saved`, 'success')
       hide()
     },
+    onError: (err: any) => {
+      displayToast(err.reason as string, 'danger')
+    },
+  })
+
+  const deleteClient = useDeleteOAuthClient({
```
nickgros, line 95: When using multiple mutations, I've started doing something like
```suggestion
  const { mutate: delete } = useDeleteOAuthClient({
```

Then instead of `deleteClient.mutate(...)` you could do `delete(...)`. I think it's more readable but I'll leave it up to you.

nickgros (approved):

PR #1689, opened 2022-07-27
Generalize useGetIsAllSelectedInfiniteList

nickgros:
I was considering re-using `useGetIsAllSelectedFromInfiniteList` for the Trash Can page. I ultimately decided against it, but first rewrote the hook as it only worked with the Entity Finder.

In the hook and underlying function, replace Entity Finder-specific data structures with predicate arguments, so that the utility can be used with items of any type.

### DIFF
--- a/src/lib/containers/entity_finder/EntityFinder.tsx
+++ b/src/lib/containers/entity_finder/EntityFinder.tsx
@@ -15,12 +15,19 @@ import 'react-reflex/styles.css'
 import { SizeMe } from 'react-sizeme'
 import Arrow from '../../assets/icons/Arrow'
 import { SynapseClient } from '../../utils'
-import { entityTypeToFriendlyName } from '../../utils/functions/EntityTypeUtils'
+import {
+  entityTypeToFriendlyName,
+  getEntityTypeFromHeader,
+} from '../../utils/functions/EntityTypeUtils'
 import { SYNAPSE_ENTITY_ID_REGEX } from '../../utils/functions/RegularExpressions'
 import { useSynapseContext } from '../../utils/SynapseContext'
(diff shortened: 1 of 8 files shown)


src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -142,64 +150,21 @@ export const EntityFinder: React.FunctionComponent<EntityFinderProps> = ({
...
-    const match = selected.get(entity.targetId)
-    if (match == null) {
-      return false
-    }
-    if (match === NO_VERSION_NUMBER) {
-      return entity.targetVersionNumber === undefined
-    }
-    return match === entity.targetVersionNumber
-  }
+  const [selectedEntities, toggleSelection] = useEntitySelection(selectMultiple)
```
nickgros, line 153: I moved the reducer to its own file, captured in one hook `useEntitySelection`.

src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -142,64 +150,21 @@ export const EntityFinder: React.FunctionComponent<EntityFinderProps> = ({
...
-  const [selectedEntities, toggleSelection] = useReducer(
-    entitySelectionReducer,
-    Map<string, number>(),
+  const isSelectable = useCallback(
+    (entity: EntityHeader | ProjectHeader | Hit) => {
+      const type = getEntityTypeFromHeader(entity)
+      return selectableTypes.includes(type)
+    },
+    [selectableTypes],
   )
```
nickgros, line 168: Create new functions to determine if a row is selected for select-all purposes. Using `useCallback` to ensure that this function will update when the underlying selected data updates.

jay-hodgson (approved):

PR #1690, opened 2022-07-27
Add aria-label for checkboxes without a visible label

nickgros:
Instead of providing an empty string as for a label, which is inaccessible, apply the label to the checkbox using `aria-label` and add the ability to hide the label with a new `hideLabel` prop.

### DIFF
--- a/src/lib/containers/widgets/Checkbox.tsx
+++ b/src/lib/containers/widgets/Checkbox.tsx
@@ -3,6 +3,7 @@ import React, { useEffect, useState } from 'react'
 
 export type CheckboxProps = {
   label: string
+  hideLabel?: boolean
   checked?: boolean
   className?: string
   onChange: (newValue: boolean) => void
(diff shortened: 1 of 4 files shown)


jay-hodgson (approved):

src/lib/containers/widgets/Checkbox.tsx
```diff
@@ -49,14 +51,15 @@ export const Checkbox: React.FunctionComponent<CheckboxProps> = (
...
+        aria-label={props.label}
         type="checkbox"
         checked={checked}
         id={uniqueId}
         onChange={handleCheckboxChange}
         disabled={disabled}
         data-testid={props['data-testid']}
       />
-      <label htmlFor={uniqueId}>{props.label}</label>
+      {<label htmlFor={uniqueId}>{hideLabel ? <></> : props.label}</label>}
```
jay-hodgson, line 62: Is a react fragment preferable to an empty string here?

nickgros: Not sure if it matters, I'm trying not to imply that the label is the empty string. Unfortunately, removing the label breaks our checkbox styling.

jay-hodgson: Not suggesting to remove the label, just asking about the label text ('' vs <>/<>)

PR #1696, opened 2022-07-29
SWC-5637: Set modal-footer-margin-between to 0 because it breaks alignment

nickgros:
image

Spacing is still OK when there are multiple buttons:
image

### DIFF
--- a/src/lib/style/bootstrap4_backports/_base-import.scss
+++ b/src/lib/style/bootstrap4_backports/_base-import.scss
@@ -15,6 +15,7 @@
   $input-border-color: SRC.$border-color-gray;
   $input-border-width: 1px;
   $input-border-radius: 2px;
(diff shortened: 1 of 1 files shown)


src/lib/style/bootstrap4_backports/_base-import.scss
```diff
@@ -15,6 +15,7 @@
   $input-border-color: SRC.$border-color-gray;
   $input-border-width: 1px;
   $input-border-radius: 2px;
+  $modal-footer-margin-between: 0rem;
```
nickgros, line 18: By default, this bootstrap 4 variable is nonzero, and adds margin around each item in the modal footer. The margin causes buttons to be improperly offset from the right edge of the footer content

jay-hodgson (approved):

PR #1699, opened 2022-07-29
SWC-6219

yjyoo773:


yjyoo773:
Screen Shot 2022-07-29 at 2 50 31 PM
Screen Shot 2022-07-29 at 2 50 41 PM

src/lib/containers/oauth/OAuthManagement.tsx
```diff
@@ -5,19 +5,43 @@ import moment from 'moment'
...
+  const warningHeader = 'Are you absolutely sure?'
+  const warningBody =
+    'If you have an existing secret, generating a new secret will make your application invalid after generation. This action cannot be undone.'
+
+  const onShowSecret = async () => {
+    setIsShowingSecretWarning(false)
+    setSelectedClient(undefined)
+    setIsShowingSecret(true)
+
+    const secret = await SynapseClient.createOAuthClientSecret(
```
jay-hodgson, line 38: What happens if this call fails?  Since this is not in a try/catch, I'm concerned that the error is lost (does this propagate to an error boundary?)

jay-hodgson: At the very least, I'd like to see a displayToast in a catch block so an error is not lost

jay-hodgson (approved):

PR #1708, opened 2022-08-10
SWC-5961: UserCardSmall improvements

nickgros:
- Refactor UserCardSmall to always render the same block of code
- When showFullName, render as Full Name (@username) instead of @username Full Name
- Remove color styling. It will have the correct color if it's a link.

SWC-5961 will utilize the new name ordering and should not have a color applied, as it will not be a link.

image

### DIFF
--- a/src/lib/containers/UserCardSmall.tsx
+++ b/src/lib/containers/UserCardSmall.tsx
@@ -99,7 +99,8 @@ export const UserCardSmall = (props: UserCardSmallProps) => {
   const fullName =
     showFullName && (userProfile.firstName || userProfile.lastName) ? (
       <span className={'user-fullname'}>
-        {`${userProfile.firstName ?? ''}`}&nbsp;
+        {`${userProfile.firstName ?? ''}`}
+        {userProfile.firstName && userProfile.lastName ? '\u00A0' : ''}
         {`${userProfile.lastName ?? ''}`}
       </span>
     ) : null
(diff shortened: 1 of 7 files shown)


src/lib/containers/UserCardSmall.tsx
```diff
@@ -96,64 +96,45 @@ export const UserCardSmall = (props: UserCardSmallProps) => {
...
+        {`${userProfile.firstName ?? ''}`}
+        {userProfile.firstName && userProfile.lastName ? '\u00A0' : ''}
+        {`${userProfile.lastName ?? ''}`}
+      </span>
+    ) : null
+
+  const Tag = showCardOnHover || !disableLink ? 'a' : 'span'
 
-  return showCardOnHover ? (
+  const style: React.CSSProperties = showCardOnHover
```
jay-hodgson, line 110: Should probably avoid nested ternary operators for readability.  This case is pretty straightforward, but could be expanded into an if/then conditional instead.

src/lib/containers/UserCardSmall.tsx
```diff
@@ -96,64 +96,45 @@ export const UserCardSmall = (props: UserCardSmallProps) => {
...
-      {`${userProfile.firstName ?? ''}`}&nbsp;{`${userProfile.lastName ?? ''}`}
-    </span>
-  ) : (
-    <></>
-  )
+  const fullName =
+    showFullName && (userProfile.firstName || userProfile.lastName) ? (
+      <span className={'user-fullname'}>
+        {`${userProfile.firstName ?? ''}`}
+        {userProfile.firstName && userProfile.lastName ? '\u00A0' : ''}
```
jay-hodgson, line 103: Constant for the no-break space?

jay-hodgson (approved):

PR #1714, opened 2022-08-11
SWC-5961: Show tooltip in Entity Tree

nickgros:
Dual-pane:

https://user-images.githubusercontent.com/17580037/184215040-9f916aa9-3652-48ea-8b4a-4251c1e70f8a.mov

Single-pane:

https://user-images.githubusercontent.com/17580037/184215144-a23bd93a-585f-4d27-9c97-ed8607e6a2d5.mov

### DIFF
--- a/src/lib/containers/entity_finder/tree/VirtualizedTree.tsx
+++ b/src/lib/containers/entity_finder/tree/VirtualizedTree.tsx
@@ -275,7 +275,7 @@ export function Node(
     return <></>
   }
 
-  let tooltipContent = ''
+  let tooltipContent: React.ReactNode = ''
   if (isEntityHeaderNode(node)) {
     tooltipContent = (
       <div style={{ textAlign: 'center' }}>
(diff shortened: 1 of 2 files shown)


src/lib/containers/entity_finder/tree/VirtualizedTree.tsx
```diff
@@ -267,6 +276,40 @@ export function Node(
...
+  let tooltipContent: React.ReactNode = ''
+  if (isEntityHeaderNode(node)) {
+    tooltipContent = (
+      <div style={{ textAlign: 'center' }}>
+        <Typography display="inline" component="span" variant="smallText1">
+          {node.name}
+          <br />
+          <b>ID: </b>
+          {node.id}
+          {'modifiedBy' in node && (
```
nickgros, line 288: Note that we may not have `modifiedBy` or `modifiedOn`. That will happen if the node was retrieved using the entity path, which does not return those fields. Currently, the node that appears this way is the project in the "Current Project" configuration. The tooltip will just be missing the `modified[On|By]` fields.

jay-hodgson (approved):

PR #1719, opened 2022-08-16
Update type for QueryResultBundle: Row.values may include nulls

nickgros:
- The change causes typing issues in many portal components where we make the assumption that the rows will not contain a null value. Since these are typically controlled, curated tables, this is a safe assumption. In those cases, emit a warning to the console. We can refine how we handle those cases as needed.
- In other cases, use filtering or nullish coalescing to resolve type errors

### DIFF
--- a/src/lib/containers/UpsetPlot.tsx
+++ b/src/lib/containers/UpsetPlot.tsx
@@ -93,7 +93,11 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
 
         for (const row of queryResult?.queryResults.rows ?? []) {
           for (let j = 1; j < row.values.length; j += 1) {
-            const rowValues: string[] = row.values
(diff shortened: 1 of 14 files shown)


src/lib/containers/UpsetPlot.tsx
```diff
@@ -93,7 +93,11 @@ const UpsetPlot: React.FunctionComponent<UpsetPlotProps> = ({
 
         for (const row of queryResult?.queryResults.rows ?? []) {
           for (let j = 1; j < row.values.length; j += 1) {
-            const rowValues: string[] = row.values
+            const rowValues: string[] = row.values as string[]
+            if (rowValues.some(value => value === null)) {
+              console.warn('Row has null value(s) when no nulls expected')
```
jay-hodgson, line 98: Message could be a reusable constant

jay-hodgson (approved):

PR #1721, opened 2022-08-16
PORTALS-2272: Check selectColumn for deep equality before resetting visible columns

nickgros:
- Remove unused toggleColumnSelection from SynapseTable
- Remove async from goToPage.
- Apply key to React.Fragment
- PORTALS-2272: Check selectColumn for deep equality before resetting visible columns

### DIFF
--- a/src/__tests__/lib/containers/widgets/ElementWithTooltip.test.tsx
+++ b/src/__tests__/lib/containers/widgets/ElementWithTooltip.test.tsx
@@ -1,9 +1,10 @@
-import * as React from 'react'
+import { act, cleanup, render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import _ from 'lodash-es'
-import { render, fireEvent, cleanup, screen } from '@testing-library/react'
-import { ElementWithTooltip } from '../../../../lib/containers/widgets/ElementWithTooltip'
+import React from 'react'
 import Columns from '../../../../lib/assets/icons/columns'
 import { IconSvgOptions } from '../../../../lib/containers/IconSvg'
+import { ElementWithTooltip } from '../../../../lib/containers/widgets/ElementWithTooltip'
 
 const mockCallback = jest.fn()
 type ElementWithTooltipProps = React.ComponentProps<typeof ElementWithTooltip>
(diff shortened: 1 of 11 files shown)


src/lib/containers/table/SynapseTable.tsx
```diff
@@ -980,29 +979,6 @@ export default class SynapseTable extends React.Component<
     const { data } = this.props.queryContext
     return data?.queryResult?.queryResults.headers.length ?? 0
   }
-  /**
-   * Handles the toggle of a column select, this will cause the table to
-   * either show the column or hide depending on the prior state of the column
-   *
-   * @memberof SynapseTable
-   */
-  public toggleColumnSelection = (columnName: string) => {
```
nickgros, line 989: Unused.

src/lib/style/base/_core.scss
```diff
@@ -1205,7 +1205,7 @@ hr ~ .SRC-wrapper {
   padding: 0px;
 }
 
-#addAndRemoveColumns ~ div.dropdown-menu {
+#addAndRemoveColumns-dropdown.dropdown-menu {
```
nickgros, line 1208: The style was previously applied relative to the tooltip element. When the Tooltip implementation changed, this style broke.

src/lib/containers/widgets/ElementWithTooltip.tsx
```diff
@@ -102,13 +100,7 @@ export const ElementWithTooltip: FunctionComponent<ElementWithTooltipProps> = ({
...
     )
   } else {
-    const outerChild = children as JSX.Element
-    tooltipTrigger = React.cloneElement(outerChild, {
-      className: `${outerChild.props.className} SRC-hand-cursor`,
-      id: idForToolTip,
-      'data-for': idForToolTip,
-      'data-tip': tooltipText,
-    })
+    tooltipTrigger = <div className="SRC-hand-cursor">{children}</div>
```
nickgros, line 103: Wrapping in a div seems to work ok (and much more simply)

src/mocks/msw/handlers/tableQueryHandlers.ts
```diff
@@ -9,16 +9,51 @@ import {
...
+}
+
+function removeBundleFieldsUsingMask(
+  queryResultBundle: QueryResultBundle,
+  partMask: number,
+): QueryResultBundle {
+  const result = cloneDeep(queryResultBundle)
+  Object.entries(BIT_TO_FIELD_MAP).forEach(([bit, field]) => {
+    if ((partMask & parseInt(bit)) === 0) {
+      delete result[field]
```
jay-hodgson, line 47: Are you cleaning up the response because the backend returns an object that contains fields that were not requested?

src/__tests__/lib/containers/widgets/ElementWithTooltip.test.tsx
```diff
@@ -83,6 +87,12 @@ describe('basic function', () => {
...
     expect(tooltipTrigger!.classList.contains('my_class')).toBe(true)
     expect(tooltipTrigger!.innerHTML).toBe('hello world')
-    expect(tooltipTrigger!.attributes['data-tip'].value).toBe(props.tooltipText)
+
+    userEvent.hover(tooltipTrigger!)
+
+    act(() => {
+      jest.runAllTimers()
+    })
+    await screen.findByText(props.tooltipText)
```
jay-hodgson, line 96: These utility functions to jest.useFakeTimers() and jest.runAllTimers()  are useful!

jay-hodgson (approved):

src/mocks/msw/handlers/tableQueryHandlers.ts
nickgros: Correct. It's easier to build the entire response and have MSW strip it out based on request, than to set it up for each test.

One place it's important for the response data to match this behavior is when testing pagination in QueryWrapper. We request the rows separately from the statistics, so when a user navigates to a different page, only new rows are fetched. If the response with new row data also contains non-row data, the naive merge that we do to combine the responses will overwrite the old statistics data, which wouldn't happen if that data was never returned.

PR #1723, opened 2022-08-17
SWC-5846

nickgros:
Update EntityFinder to request favorites sorted by name, ascending.

### DIFF
--- a/src/__tests__/lib/utils/hooks/SynapseAPI/useFavorites.test.tsx
+++ b/src/__tests__/lib/utils/hooks/SynapseAPI/useFavorites.test.tsx
@@ -8,7 +8,7 @@ import {
 import { MOCK_CONTEXT_VALUE } from '../../../../../mocks/MockSynapseContext'
 import { QueryClient } from 'react-query'
 import { SynapseContextProvider } from '../../../../../lib/utils/SynapseContext'
-
+import { SynapseClient } from '../../../../../lib/utils'
 const queryClient = new QueryClient()
 
 const wrapper = (props: { children: React.ReactChildren }) => (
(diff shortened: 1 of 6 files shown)


src/lib/utils/hooks/SynapseAPI/user/useFavorites.ts
```diff
@@ -8,8 +8,14 @@ import { SynapseClient } from '../../..'
...
 import { useSynapseContext } from '../../../SynapseContext'
 import { EntityHeader, PaginatedResults } from '../../../synapseTypes'
+import {
+  FavoriteSortBy,
+  FavoriteSortDirection,
+} from '../../../synapseTypes/FavoriteSortBy'
 
 export function useGetFavorites(
+  sort: FavoriteSortBy = 'FAVORITED_ON',
+  sortDirection: FavoriteSortDirection = 'DESC',
```
nickgros, line 18: This is the sort that's currently used on the Favorites page. It 
 * has its own client-side sorting and
 * has no column for the "favorited on" date, (not returned by the backend)

Thus a user cannot revert the sort to be the "favorited on" datetime, so I didn't want to change it.

src/lib/containers/entity_finder/details/configurations/FavoritesDetails.tsx
```diff
@@ -10,7 +10,7 @@ export const FavoritesDetails: React.FunctionComponent<
   FavoritesDetailsProps
 > = ({ ...sharedProps }) => {
   const { data, isLoading, hasNextPage, fetchNextPage, isFetchingNextPage } =
-    useGetFavoritesInfinite({ useErrorBoundary: true })
+    useGetFavoritesInfinite('NAME', 'ASC', { useErrorBoundary: true })
 
   const entities = data?.pages.flatMap(page => page.results) ?? []
```
nickgros, line 16: Below, we could add functionality to allow the user to change the direction of the sort as in other Entity Finder configurations, but would require some refactoring since it currently expects all of `name`, `createdOn`, `modifiedOn` to be sortable when in this case only `name` is.

src/__tests__/lib/utils/hooks/SynapseAPI/useFavorites.test.tsx
```diff
@@ -33,12 +33,14 @@ const expected: PaginatedResults<EntityHeader> = {
...
+      isLatestVersion: true,
     },
   ],
 }
 
-const SynapseClient = require('../../../../../lib/utils/SynapseClient')
-SynapseClient.getUserFavorites = jest.fn().mockResolvedValue(expected)
+const mockGetUserFavorites = jest
+  .spyOn(SynapseClient, 'getUserFavorites')
+  .mockResolvedValue(expected)
```
nickgros, line 43: Prefer `spyOn(...).mock` over assignment since it preserves types, and we don't need an ugly inline import.

jay-hodgson (approved):

PR #1728, opened 2022-08-19
SWC-6182: Remove Enzyme

nickgros:
* Rewrite some tests for RangeSlider (hard to complete because interacting with the slider is not possible in jsdom)
* Rewrite some tests for SynapseForm (hard to complete because existing tests rely heavily on class method testing, component should never be used outside of SynapseFormWrapper).
* Remove Enzyme


src/__tests__/lib/containers/widgets/RangeSlider.test.tsx
```diff
@@ -17,46 +20,55 @@ function createTestProps(overrides?: RangeSliderProps): RangeSliderProps {
...
 
-  it('should call the callbackFn on change when doUpdateOnApply is false', () => {
-    init({ ...props, doUpdateOnApply: false })
-    mockCallback.mockClear()
-    wrapper.find('Slider').simulate('change', [2, 10])
-    expect(mockCallback).toHaveBeenCalledWith({ min: 2, max: 10 })
-  })
+    /**
+     * Hard to test updating values in JSDOM, so skipping this test.
+     * It would be worth coming back to this if we use an interaction testing library,
```
nickgros, line 60:
```suggestion
     * It would be worth coming back to this if we use an interaction testing library that runs in a browser,
```

PR #1736, opened 2022-08-23
Fill in fields for mockDataset

nickgros:


src/mocks/entity/mockDataset.ts
```diff
@@ -85,13 +106,15 @@ const versionInfo: VersionInfo[] = [
 
 const bundle: EntityBundle = {
   entity: mockDatasetEntity,
+  entityType: EntityType.DATASET,
```
nickgros, line 109: Bundle would be successfully fetched in EntityModal but entityType would be undefined

jay-hodgson (approved):

PR #1737, opened 2022-08-23
PORTALS-2271: svg should be white if the background has the primary c…

jay-hodgson:
…olor

Before: 
Screen Shot 2022-08-23 at 2 25 22 PM

After: 
Screen Shot 2022-08-23 at 2 28 33 PM

### DIFF
--- a/src/lib/style/base/_core.scss
+++ b/src/lib/style/base/_core.scss
@@ -143,6 +143,12 @@ svg.SRC-hoverBox {
       }
     }
   }
+  .SRC-primary-background-color {
+    svg,
+    path {
+      fill: #fff;
+    }
+  }
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/style/base/_core.scss
```diff
@@ -143,6 +143,12 @@ svg.SRC-hoverBox {
       }
     }
   }
+  .SRC-primary-background-color {
+    svg,
+    path {
+      fill: #fff;
+    }
+  }
```
nickgros, line 151: I don't think this is used in very many places, so it should be fine. It might be worth explicitly handling this behavior in IconSvg so we can properly style fill/color depending on what's appropriate.

nickgros (approved):

PR #1741, opened 2022-08-25
PORTALS-2283

yjyoo773:


package.json
```diff
@@ -33,6 +33,7 @@
     "dagre": "^0.8.5",
     "downshift": "^6.1.2",
     "gh-pages": "^3.2.3",
+    "html-react-parser": "^3.0.4",
```
nickgros, line 36: Is this necessary instead of using `dangerouslySetInnerHTML`?

src/lib/containers/CertificationQuiz.tsx
```diff
@@ -0,0 +1,210 @@
...
+const CertificationQuiz: React.FunctionComponent = () => {
+  const { accessToken } = useSynapseContext()
+  const [quiz, setQuiz] = useState<Quiz | undefined>()
+  const [questionResponse, setQuestionResponse] = useState<QuestionResponse[]>(
+    [],
+  )
+  const [passingRecord, setPassingRecord] = useState<PassingRecord>()
+
+  const gettingStartedUrl =
+    'https://help.synapse.org/docs/Getting-Started.2055471150.html'
```
nickgros, line 29:
```suggestion
  const GETTING_STARTED_URL =
    'https://help.synapse.org/docs/Getting-Started.2055471150.html'
```

src/lib/containers/CertificationQuiz.tsx
```diff
@@ -0,0 +1,210 @@
...
+    } catch (err: any) {
+      console.log(err)
+    }
+  }
+
+  useEffect(() => {
+    if (accessToken) {
+      getQuiz()
+    }
+  }, [accessToken])
```
nickgros, line 43: IMO all of the state/effects are cleaner if we used `react-query`, but I recognize that you would have to manually configure the options and manipulate the cache afterwards so the quiz doesn't get overwritten. So I'll leave it up to you whether to keep what you have or make the change.

```tsx
const { data: quiz } = useGetCertifyQuiz({
  enabled: !!accessToken,
  staleTime: Infinity,
  cacheTime: Infinity
})
```

package.json
yjyoo773: No it is not. Would you recommend using `dangerouslySetInnerHTML` over the parser?

src/lib/containers/CertificationQuiz.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  useEffect(() => {
+    if (!accessToken) {
+      new Error('Please sign in to take certification quiz')
+    }
+  }, [accessToken])
```
nickgros, line 49: The error isn't thrown. Also, the call to the certifiedUserTest endpoint throws a 403 if the user isn't signed in. Could we just do this? -

1. make the API call regardless of whether or not the user has a token
2. Then pass the error to the error boundary either with ReactQuery's `useErrorBoundary: true` or `react-error-boundary`'s useErrorHandler().handleError.
3. The error boundary will render the appropriate alert

src/lib/containers/CertificationQuiz.tsx
```diff
@@ -0,0 +1,210 @@
...
+        answerIndex: [answer],
+        concreteType: MULTICHOICE_RESPONSE_CONCRETE_TYPE_VALUE,
+      },
+    ]
+    setQuestionResponse(newState)
+  }
+
+  const handleRetakeQuiz = () => {
+    window.location.reload()
+  }
```
nickgros, line 65: Not great UX, could we just clear state (or if using react-query, clear cache) and refetch a new quiz?

src/__tests__/lib/containers/CertificationQuiz.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const getQuizHandler = rest.get(
+  `${getEndpoint(
+    BackendDestinationEnum.REPO_ENDPOINT,
+  )}/repo/v1/certifiedUserTest`,
+  async (req, res, ctx) => {
+    const response: Quiz = {
+      id: 123,
+      header: 'Mock Certification Quiz',
+      questions: [],
```
nickgros, line 31: Would like to see 1-2 mock questions in here. Would want to see these additional scenarios: 
- Choose an answer, verify that the selected answer was submitted
- Simulate case where submit clicked and not all questions have been answered
- Simulate case where all answers are correct
- Simulate case where an answer was wrong

nickgros (changes requested):
See comments

package.json
nickgros: If it works without issues, yes. I see no reason to 
- include a new dependency
- parse the HTML that we already expect to be well-formed

src/lib/containers/CertificationQuiz.tsx
```diff
@@ -0,0 +1,213 @@
...
+          </Typography>
+          {passingRecord.passed ? (
+            displayToast(
+              `You passed the Synapse Certification Quiz on ${passingRecord.passedOn}`,
+              'success',
+            )
+          ) : (
+            <Typography variant="body1">
+              Please review the items shown in red below, and{' '}
+              <a style={{ color: '#407ba0' }} onClick={handleRetakeQuiz}>
```
nickgros, line 116: You can get the standardized link styling if you have an href. Then preventDefault so it doesn't do anything
```suggestion
              <a href="#" onClick={e => { 
                e.preventDefault()
                handleRetakeQuiz()
              }}>
```

src/__tests__/lib/containers/CertificationQuiz.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    await waitFor(() =>
+      expect(mockToastFn).toBeCalledWith(
+        'Please answer all of the questions and try again.',
+        'warning',
+      ),
+    )
+  })
+
+  it('Submit quiz that did not pass', async () => {
```
nickgros, line 93: Can you add onto this test or write a new test to verify the `handleRetakeQuiz` works?

nickgros (changes requested):
Looking good! Just a few more changes

nickgros (approved):
Great work @yjyoo773 !

PR #1743, opened 2022-08-26
SWC-6077: Contextually update latest version text

nickgros:
Replace `showVersionSelection`, `mustSelectVersionNumber`, and `latestVersionText` with an enum `versionSelection` that captures each possible case where we would want to change how a user can select versions.

The possible values for `versionSelection`, with examples, are
|Value|Description|Example Scenario|
|---|---|---|
|DISALLOWED|the user cannot select a version|Moving an entity, adding projects to a project view|
|REQUIRED|the user must select a version|Adding files to a dataset|
|TRACKED|the user may select a version, and contextually, changes to the latest version will be tracked if selected|Adding items to an entity list widget in a wiki
|UNTRACKED|the user may select a version, and contextually, changes to the latest version will NOT be tracked if selected. | Importing table columns|

Additionally, add a help tooltip when selecting a table that has no snapshots when a version number is required (e.g. creating a dataset collection).

image

This is a breaking change, so SWC must be updated accordingly.

### DIFF
--- a/src/lib/containers/entity_finder/details/view/DetailsViewTableRenderers.tsx
+++ b/src/lib/containers/entity_finder/details/view/DetailsViewTableRenderers.tsx
@@ -387,9 +387,13 @@ export const DetailsViewVersionRenderer = ({
     versions,
   ])
 
+  if (!isSelected || !isVersionableEntity) {
+    return <></>
+  }
+
   return (
     <div>
-      {isSelected && versions && versions.length > 0 && (
+      {versions && versions.length > 0 ? (
         <Form.Control
           role="listbox"
           size="sm"
@@ -419,6 +423,8 @@ export const DetailsViewVersionRenderer = ({
             )
           })}
         </Form.Control>
+      ) : (
+        latestVersionText
       )}
     </div>
   )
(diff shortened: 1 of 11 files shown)


src/lib/utils/functions/EntityTypeUtils.ts
```diff
@@ -66,6 +66,26 @@ export function isContainerType(type: EntityType): boolean {
   }
 }
 
+export function isTableType(type: EntityType): boolean {
```
nickgros, line 69: Every time we add another function like this, I get a stronger urge to replace our data types with classes

src/lib/containers/entity_finder/details/view/DetailsViewTableRenderers.tsx
```diff
@@ -347,29 +349,70 @@ export const DatasetEditorVersionRenderer = ({
...
+  } = rowData
   const { data: versionData } = useGetVersionsInfinite(id, {
     enabled: isVersionableEntity,
     staleTime: 60 * 1000, // 60 seconds
   })
   const versions = versionData?.pages.flatMap(page => page.results) ?? []
 
   useEffect(() => {
-    // If 'mustSelectVersionNumber' is true and the user has selected this entity and hasn't selected a version, then we force a version selection
+    // If VersionSelectionType.REQUIRED, and the user has selected this entity and hasn't selected a version, then we force a version selection
```
nickgros, line 412:
```suggestion
    // If versionSelectionType is REQUIRED, and the user has selected this entity and hasn't selected a version, then we force a version selection
```

jay-hodgson (approved):

PR #1749, opened 2022-08-30
SWC-6265

yjyoo773:


src/lib/utils/hooks/SynapseAPI/oauth/useOAuthClient.ts
```diff
@@ -78,5 +79,8 @@ export function useMutateOAuthClient(
         await options.onSuccess(updatedClient, variables.client, ctx)
       }
     },
+    onError: (err: any) => {
```
jay-hodgson, line 82: This is a good pattern to follow.  Also add the same onError to useDeleteOAuthClient()?

jay-hodgson (approved):

PR #1751, opened 2022-08-30
PORTALS-2296

jay-hodgson:
Abort starting new queries if component is unmounted


src/lib/containers/table/EntityIDColumnCopyIcon.tsx
```diff
@@ -32,13 +45,17 @@ const EntityIDColumnCopyIcon = () => {
...
+      abortController?.signal,
+    )
       .then((data: QueryResultBundle) => {
         setIdData(data)
       })
       .catch((err: any) => {
         console.error(err)
-        displayToast('Unable to query for all IDs', 'danger')
+        displayToast('Unable to query for all Synapse IDs to copy', 'danger')
       })
```
nickgros, line 59: If the query is cancelled, the `AbortError` would be caught here, and the toast would be shown. Seems fine to me.

nickgros (approved):

PR #1760, opened 2022-09-06
PORTALS-2157

nickgros:
- Replace search param-based SQL manipulation with QueryFilter addition
- Remove `linkCountToDisaggregatedQuery` option from SynapseTable
- Remove `sql-parser` from package.json


nickgros:
We're also using `additionalFilters` to determine what sort of content to show when there are no results, with the assumption being that present `additionalFilters` implies an interactive result (the user needs to fix the query) while no `additionalFilters` implies a non-interactive result ("There is no content, check back later"). This will be incorrect with this change, so we have to change how we decide which empty content placeholder to show

nickgros:
Closed in favor of #1868 and others

PR #1761, opened 2022-09-06
Update polyfills to fix broken storybook

nickgros:
Upgrading and bundling sanitize-html caused some packages relying on missing polyfills to be included.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -28,6 +28,7 @@ module.exports = {
       base,
       define: {
         'process.browser': 'true',
+        'process.env.NODE_DEBUG': undefined,
       },
       plugins: [
         // TODO: The following will fix SynapseFormWrapper in development mode but will break the storybook static build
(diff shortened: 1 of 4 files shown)


.storybook/main.ts
```diff
@@ -28,6 +28,7 @@ module.exports = {
       base,
       define: {
         'process.browser': 'true',
+        'process.env.NODE_DEBUG': undefined,
```
nickgros, line 31: Running in a browser, so disable NODE_DEBUG for those scripts that check this flag

package.json
```diff
@@ -199,6 +199,7 @@
     "timers-browserify": "^2.0.12",
     "ts-jest": "^28.0.5",
     "typescript": "4.7.4",
+    "util": "^0.12.4",
```
nickgros, line 202: Polyfill for node's `util`

jay-hodgson (approved):

PR #1764, opened 2022-09-08
PORTALS-2157: Rename LockedFacet to LockedColumn

nickgros:
Pre-work for PORTALS-2157

The "lock" we apply to hide charts/facets/active filters can be applied to any column; the locked column does not necessarily need to be faceted.

This is a breaking change, but it is easily caught by TypeScript.

### DIFF
--- a/src/lib/utils/functions/queryUtils.ts
+++ b/src/lib/utils/functions/queryUtils.ts
@@ -1,11 +1,13 @@
+import { cloneDeep } from 'lodash-es'
+import { SynapseClient, SynapseConstants } from '..'
+import { LockedColumn } from '../../containers/QueryContext'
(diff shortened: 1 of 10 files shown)


src/lib/utils/functions/queryUtils.ts
```diff
@@ -98,3 +100,22 @@ export const isSingleNotSetValue = (facet: FacetColumnResult): boolean => {
       SynapseConstants.VALUE_NOT_SET
   )
 }
+
+export function removeLockedColumnFromFacetData(
```
jay-hodgson, line 104: Appreciate this refactor

jay-hodgson (approved):

PR #1766, opened 2022-09-09
Dependency upgrades, add Storybook ESLint plugin

nickgros:


package.json
```diff
@@ -182,7 +183,6 @@
     "jsdom": "^20.0.0",
     "lint-staged": ">=10",
     "msw": "^0.38.2",
-    "npm": "^6.9.0",
```
nickgros, line 185: Not sure why we had this (old) version of npm as a dependency.

jay-hodgson (approved):

PR #1771, opened 2022-09-12
PORTALS-1824

jay-hodgson:


jay-hodgson:
Demo
https://user-images.githubusercontent.com/1864447/189722264-638ea975-7753-4c17-8e8e-d066e50fab30.mov

src/lib/containers/QuerySortSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+  sortConfig: SortConfiguration
+  facetAliases?: Record<string, string>
+}
+
+const QuerySortSelector: React.FunctionComponent<QuerySortSelectorProps> = ({
+  sortConfig,
+  facetAliases,
+}) => {
+  const { defaultColumn, defaultDirection, sortableColumns } = sortConfig
+  const infiniteQueryContext = useInfiniteQueryContext()
```
nickgros, line 24: Can just use `useQueryContext` here, because you aren't using any fields exclusive to the InfiniteQueryContext.

src/lib/containers/QuerySortSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+    setSortColumn(value)
+    setSortDirection(newSortDirection)
+  }
+
+  return (
+    <div className="QuerySortSelector bootstrap-4-backport">
+      <Typography variant="label" className="sort-by-label SRC-inlineBlock">
+        Sort by
+      </Typography>
+      <Creatable
```
nickgros, line 65: Don't use `Creatable`; we use it in the annotations editor because in that scenario the user can set a value outside of the enum. Not true here (right?), so instead use `Select`

src/lib/containers/QuerySortSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+  })
+
+  const onChange = (value?: string) => {
+    const lastQueryRequestDeepClone = getLastQueryRequest()
+    let newSortDirection: SortDirection = 'ASC'
+    if (value === sortColumn) {
+      // flip sort direction
+      newSortDirection = sortDirection === 'ASC' ? 'DESC' : 'ASC'
+    }
```
nickgros, line 44: More of a design issue--this functionality isn't clear from the UI, in my opinion.

nickgros (changes requested):
Looks slick. Is `Creatable` necessary or can we change it to `Select` (or other appropriate component)?

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <QuerySortSelector {...props} />
+      </QueryVisualizationContextProvider>
+    </QueryContextProvider>,
+    {
+      wrapper: createWrapper(),
+    },
+  )
+}
+
+describe('QuerySortSelector tests', () => {
```
nickgros, line 44: Can we also test flipping the sort direction when re-selecting it (if you decide to keep that functionality?)

jay-hodgson: Is there a way for me to interrogate the query request object (to verify the SortItems)?

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+  it('Executes query request on sort', async () => {
+    renderComponent(props, queryContext)
+    const input = screen.getByRole('combobox')
+    // User typically enters the beginning of a column name to populate the selections
+    await userEvent.type(input, 'journal')
+    await screen.findAllByText(new RegExp('Open Access Journals'))
+    // Make a selection
+    act(() => {
+      fireEvent.keyDown(input, { key: 'Enter' })
+    })
```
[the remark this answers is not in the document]
jay-hodgson: Copy/paste from another test.  I'll try that!

src/lib/containers/QuerySortSelector.tsx
jay-hodgson: Yes, I totally agree.  This is not obvious and my implementation deviates from the design.  The design had the selected item indicating the current sort direction, and another item in the combobox (for the same column) that indicates the opposite sort.  This is not how select boxes work though (the selected item needs to be one of the options in the select combobox).

src/lib/containers/QuerySortSelector.tsx
jay-hodgson: You're right.  I started with Select, ran into all kinds of trouble with the types, so I changed it to utilize the same solutions from the annotations editor.  But I forgot to change it back to a Select.

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+    data,
+    hasNextPage: false,
+    getLastQueryRequest: getLastQueryRequest,
+    executeQueryRequest: executeQueryRequest,
+  }
+
+  it('renders without crashing', () => {
+    const tree = renderComponent(props, queryContext)
+    expect(tree).toBeDefined()
+  })
```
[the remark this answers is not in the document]
jay-hodgson: Removed

src/__tests__/lib/containers/QuerySortSelector.test.tsx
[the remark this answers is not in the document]
jay-hodgson: Updated

src/lib/containers/QuerySortSelector.tsx
jay-hodgson: Done

src/lib/containers/QuerySortSelector.tsx
jay-hodgson: Changed to be a `Select`

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  it('Executes query request on sort', async () => {
+    renderComponent(props, queryContext)
+    const input = screen.getByRole('combobox')
+    // User typically enters the beginning of a column name to populate the selections
+    await userEvent.type(input, 'journal')
+    await screen.findAllByText(new RegExp('Open Access Journals'))
+    // Make a selection
+    act(() => {
+      userEvent.keyboard('{Enter}')
+    })
```
nickgros, line 90: `userEvent` wraps calls in `act` internally (and I think you need `await` here, but not 100% sure)
```suggestion
    await userEvent.keyboard('{Enter}')
```

src/__tests__/lib/containers/QuerySortSelector.test.tsx
nickgros: Like this?
```
expect(executeQueryRequest).toHaveBeenCalledWith(expect.objectContaining({
	query: expect.objectContaining({
		sort: {
			// expected sort column and direction here
		}
	})
}))
```

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -78,18 +79,42 @@ describe('QuerySortSelector tests', () => {
...
+      column: 'journal',
+      direction: 'ASC',
+    }
+    verifyExpectedSortItem(expectedSortItem)
+
+    // Selecting the item again should flip the sort direction
+    sortByOpenAccessJournals()
+    expectedSortItem.direction = 'DESC'
+    verifyExpectedSortItem(expectedSortItem)
   })
```
nickgros, line 119: These are async, so the test is probably passing before actually checking anything

```suggestion
  it('Executes query request on sort', async () => {
    renderComponent(props, queryContext)
    await sortByOpenAccessJournals()

    const expectedSortItem: SortItem = {
      column: 'journal',
      direction: 'ASC',
    }
    await verifyExpectedSortItem(expectedSortItem)

    // Selecting the item again should flip the sort direction
    await sortByOpenAccessJournals()
    expectedSortItem.direction = 'DESC'
    await verifyExpectedSortItem(expectedSortItem)
  })
```

src/__tests__/lib/containers/QuerySortSelector.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const expectedSortItem: SortItem = {
+      column: 'journal',
+      direction: 'ASC',
+    }
+    verifyExpectedSortItem(expectedSortItem)
+
+    // Selecting the item again should flip the sort direction
+    sortByOpenAccessJournals()
+    expectedSortItem.direction = 'DESC'
+    verifyExpectedSortItem(expectedSortItem)
```
nickgros, line 118: These are async, so the test is probably passing before actually checking anything

```suggestion
    renderComponent(props, queryContext)
    await sortByOpenAccessJournals()

    const expectedSortItem: SortItem = {
      column: 'journal',
      direction: 'ASC',
    }
    await verifyExpectedSortItem(expectedSortItem)

    // Selecting the item again should flip the sort direction
    await sortByOpenAccessJournals()
    expectedSortItem.direction = 'DESC'
    await verifyExpectedSortItem(expectedSortItem)
```

nickgros (approved):

PR #1773, opened 2022-09-13
Refactor: replace duplicate logic in Infinite/QueryWrapper with useImmutableTableQuery

nickgros:
Most of the changes are refactoring I did to simplify PORTALS-2157 -- attempting to reduce duplicate code and improve test-ability. There is also one non-refactor change where we will now use the React Query cache in QueryWrapper and InfiniteQueryWrapper.

- Create custom hook `useImmutableTableQuery`, which has the following responsibilities
	- Stores the initial and current table queries in state
	- Provides clones of the table queries for consumers to read
	- Provides functions for consumers to perform actions on the query
		-	e.g. update the entire query object, change the page size or page number. We can provide more specific controls as needed, like adding/removing a facet filter, for example
	- Handles side effects related to query updates (calling callbacks, updating URL)
- QueryWrapper and InfiniteQueryWrapper were refactored to use `useImmutableTableQuery`
- Changed cache behavior in both QueryWrapper and InfiniteQueryWrapper
	- We were clearing the cache on every query change because of weird UX -- a delay in the controls without a loading state made it unclear that the query updated.
	- Since we made this decision I think the UI has changed in such a way that it is more clear when an updated query goes through, so let's take advantage of the cache!

### DIFF
--- a/src/__tests__/lib/utils/hooks/useImmutableTableQuery.test.tsx
+++ b/src/__tests__/lib/utils/hooks/useImmutableTableQuery.test.tsx
@@ -1,0 +1,186 @@
+import { renderHook, act } from '@testing-library/react-hooks'
+import { cloneDeep } from 'lodash-es'
+import useImmutableTableQuery, {
+  UseImmutableTableQueryOptions,
+} from '../../../../lib/containers/useImmutableTableQuery'
+import * as DeepLinkingUtils from '../../../../lib/utils/functions/deepLinkingUtils'
+
+const options: UseImmutableTableQueryOptions = {
+  initQueryRequest: {
+    concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
+    entityId: 'syn123',
+    partMask: 1,
+    query: {
+      sql: 'SELECT * FROM syn123.3 WHERE "foo"=\'bar\'',
+    },
+  },
+}
+
+describe('useImmutableTableQuery tests', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('Returns the entity ID and version in the passed query', () => {
+    const { result } = renderHook(() => useImmutableTableQuery(options))
+    expect(result.current.entityId).toEqual('syn123')
+    expect(result.current.versionNumber).toEqual(3)
+  })
+
+  it('Updates the query using a new query object', () => {
+    const { result } = renderHook(() => useImmutableTableQuery(options))
+    expect(result.current.getLastQueryRequest()).toEqual(
+      options.initQueryRequest,
+    )
+
+    const newQuery = cloneDeep(options.initQueryRequest)
+    newQuery.query.sql = 'SELECT * FROM syn123.3 WHERE "foo"=\'baz\''
+
+    // Call under test
+    act(() => {
+      result.current.setQuery(newQuery)
+    })
+    expect(result.current.getLastQueryRequest()).toEqual(newQuery)
+    expect(result.current.getLastQueryRequest()).not.toBe(newQuery)
+  })
+
+  it('Updates the query using a transformer function', () => {
+    const { result } = renderHook(() => useImmutableTableQuery(options))
+    expect(result.current.getLastQueryRequest()).toEqual(
+      options.initQueryRequest,
+    )
+
+    const newQuery = cloneDeep(options.initQueryRequest)
+    newQuery.query.sql = 'SELECT * FROM syn123.3 WHERE "foo"=\'baz\''
+
+    // Call under test
+    act(() => {
+      result.current.setQuery(currentQuery => {
+        currentQuery.query.sql = newQuery.query.sql
+        return currentQuery
+      })
+    })
+    expect(result.current.getLastQueryRequest()).toEqual(newQuery)
+  })
+
+  it('Returns a deep clone of the initial request', () => {
+    const { result } = renderHook(() => useImmutableTableQuery(options))
+    expect(result.current.getInitQueryRequest()).toEqual(
+      options.initQueryRequest,
+    )
+    expect(result.current.getInitQueryRequest()).not.toBe(
+      options.initQueryRequest,
+    )
+  })
+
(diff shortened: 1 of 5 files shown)


src/lib/containers/InfiniteQueryWrapper.tsx
```diff
@@ -150,95 +164,16 @@ export function InfiniteQueryWrapper(props: InfiniteQueryWrapperProps) {
     return infiniteData?.pages[currentPage].responseBody
   }, [currentPage, infiniteData])
 
-  useDeepCompareEffect(() => {
-    if (onQueryChange) {
-      onQueryChange(JSON.stringify(lastQueryRequest.query))
-    }
-  }, [onQueryChange, lastQueryRequest.query])
```
nickgros, line 157: This effect was totally removed from both `QueryWrapper`s -- we already call `onQueryChange` every time `setQuery`/`executeQueryRequest` is called. Not sure why we would need this effect as well.

src/lib/containers/InfiniteQueryWrapper.tsx
```diff
@@ -150,95 +164,16 @@ export function InfiniteQueryWrapper(props: InfiniteQueryWrapperProps) {
...
-      }
-    }
-    /**
-     * TODO: We remove the cached data because it can interfere with user controls, such as the QueryFilter.
-     * For example, if you filter on a facet with value ["a"], then value ["a", "b"], then value ["a"] again,
-     * we'll have a cache hit on the last query, so we won't get a loading state, but the controls haven't been updated
-     * to handle cache hits. Forcing a cache miss here fixes this, but ideally the controls should handle this case.
-     */
-    remove()
-    // end TODO
```
nickgros, line 239: This was entirely removed. See PR description for more info.

src/lib/containers/useImmutableTableQuery.ts
```diff
@@ -0,0 +1,171 @@
...
+import * as DeepLinkingUtils from '../utils/functions/deepLinkingUtils'
+import { DEFAULT_PAGE_SIZE } from '../utils/SynapseConstants'
+import { parseEntityIdAndVersionFromSqlStatement } from '../utils/functions/sqlFunctions'
+
+export type ImmutableTableQueryResult = {
+  /** The ID of the table parsed from the SQL query */
+  entityId: string
+  /** The version number of the table parsed from the SQL query */
+  versionNumber?: number
+  getInitQueryRequest: () => QueryBundleRequest
```
nickgros, line 13: Ideally, this would be an immutable object, rather than a function that returns a clone. But that's a larger change where we may have to commit to a new dependency (until we get [Records and Tuples](https://github.com/tc39/proposal-record-tuple) 🙂)

src/lib/containers/useImmutableTableQuery.ts
```diff
@@ -0,0 +1,171 @@
...
+  getLastQueryRequest: () => QueryBundleRequest
+  setQuery: (
+    queryRequest:
+      | QueryBundleRequest
+      | ((lastQueryRequest: QueryBundleRequest) => QueryBundleRequest),
+  ) => void
```
nickgros, line 19: You can now update the query with a function that transforms a mutable clone of the current request. This would be more useful if we returned an immutable version of the query request as above.

```suggestion
  setQuery: (
    queryRequest:
      | QueryBundleRequest
      | ((lastQueryRequest: QueryBundleRequest) => QueryBundleRequest),
  ) => void
```

src/lib/containers/useImmutableTableQuery.ts
```diff
@@ -0,0 +1,171 @@
...
+  pageSize: number
+  /** The current page of results. The first page is `1` */
+  currentPage: number
+  setPageSize: (pageSize: number) => void
+  /** pageNumber is 1-indexed */
+  goToPage: (pageNumber: number) => void
+
+  /**
+   * TODO: This hook could handle all potential query transformations, such as
+   *    - addFacetFilter, removeFacetFilter, clearFacetFilters
```
jay-hodgson, line 29: Like this idea instead of directly changing the Query

src/lib/containers/useImmutableTableQuery.ts
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      newQueryRequest = queryRequest
+    }
+
+    // Clone the query request before storing it in state, in case the original is mutated
+    const clonedQueryRequest = cloneDeep(newQueryRequest)
+    setLastQueryRequest(clonedQueryRequest)
+
+    if (clonedQueryRequest.query) {
+      const clonedQueryRequestJson = JSON.stringify(clonedQueryRequest.query)
```
jay-hodgson, line 123: I know it was probably doing this before, but it looks like we only need `clonedQueryRequestJson` and `stringifiedQuery` if `shouldDeepLink` is true (so these two lines could be moved into the `if (shouldDeepLink)` block).

src/__tests__/lib/utils/hooks/useImmutableTableQuery.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const newQuery = cloneDeep(options.initQueryRequest)
+    newQuery.query.sql = 'SELECT * FROM syn123.3 WHERE "foo"=\'baz\''
+
+    // Call under test
+    act(() => {
+      result.current.setQuery(newQuery)
+    })
+    expect(result.current.getLastQueryRequest()).toEqual(newQuery)
+    expect(result.current.getLastQueryRequest()).not.toBe(newQuery)
```
jay-hodgson, line 44: Ohhh, this is testing that it is making a clone, right?  Could you add a comment?  It took me a minute of staring at lines 43 and 44 before coming to this conclusion :)

jay-hodgson: Or remove since it's explicitly tested by the "Returns a deep clone of the initial request" below

jay-hodgson (approved):
Great set of tests too

PR #1775, opened 2022-09-13
PORTALS-2307

jay-hodgson:


src/lib/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -2,9 +2,11 @@
 @use 'sass:map';
 
 .QueryWrapperPlotNav {
+  .SRC-wrapper {
+    display: grid;
```
jay-hodgson, line 6: Used to only have a grid layout when in desktop mode (so in mobile the layout was based on the html output, which was not quite right).

jay-hodgson: Namely, I saw the search bar above the top level controls.  There may have been other issues

src/lib/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -19,45 +21,11 @@
...
-      }
-      .FilterAndView {
-        grid-row: 8 / span 1;
-      }
-
       // column (depends on if we are showing facet filters)
       .isShowingFacetFilters {
         &.QueryFilter {
           grid-column: 1 / span 1;
+          grid-row: 3 / span 100;
```
jay-hodgson, line 28: Shares grid-row 3 on desktop (since it is 2 column)

src/lib/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -73,18 +41,58 @@
...
+    }
+    .FacetNav {
+      grid-row: 8 / span 1;
+    }
+    .FilterAndView {
+      grid-row: 9 / span 1;
+    }
+  }
+
+  .SRC-wrapper .isHidingFacetFilters {
```
jay-hodgson, line 86: Moving this out is the main fix for PORTALS-2307 (apply display:none to QueryFilter children, when isHidingFacetFilters)

nickgros (approved):

PR #1780, opened 2022-09-15
SWC-6265: Prevent modal to close when failed to edit

yjyoo773:


yjyoo773:
The original way of creating/updating OAuthClient where I had them in one function was not compatible with `SynapseClientError`. This led to not being able to get the `onError` when using the hooks.
I ended up splitting the functionalities into separate functions to be able to use `SynapseClientError`.

yjyoo773:
https://user-images.githubusercontent.com/60489495/190466642-81349eef-3a7a-4396-a1f9-806d9697c2e5.mov

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -85,16 +86,24 @@ export const CreateOAuthModal: React.FunctionComponent<
...
     onSuccess: () => {
-      displayToast(`Successfully saved`, 'success')
-      hide()
+      displayToast('Successfully created', 'success')
     },
-    onError: (err: any) => {
-      displayToast(err.reason as string, 'danger')
+    onError: error => {
+      displayToast(error.reason as string, 'danger')
+    },
```
nickgros, line 97: For create/edit, instead of pushing a toast, can you store the error in state and show it in the modal in a bootstrap `Alert`? The response in these cases is usually a form error, and we should keep that on the screen with the form instead of showing it in an ephemeral toast.

nickgros (changes requested):
See comment--let's display form errors in a more user-friendly way

src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -121,15 +133,15 @@ export const CreateOAuthModal: React.FunctionComponent<
...
         if (warnTrigger === true) {
           setIsShowingConfirmModal(true)
-          hide()
         } else {
           if (isEdit) {
-            mutate({ action: 'UPDATE', client: oAuthClient })
+            updateClient(oAuthClient)
           } else {
-            mutate({ action: 'CREATE', client: oAuthClient })
+            createClient(oAuthClient)
```
nickgros, line 140: Should probably await these mutation calls so the modal doesn't hide right away.

nickgros (approved):

PR #1781, opened 2022-09-15
SWC-6277: AccessRequirements jsx element is being recreated on re-render

jay-hodgson:


src/lib/containers/HasAccessV2.tsx
```diff
@@ -227,8 +226,7 @@ export function HasAccessV2(props: HasAccessProps) {
   }
 
   // Show Access Requirements
-  const AccessRequirements = () => {
-    const { entityId } = props
+  const accessRequirementsJsx = useMemo(() => {
```
nickgros, line 229: This works, but I'd rather do one of these to avoid having to reason with memoization:
* Pull `AccessRequirements` out of `HasAccess` and pass along all of the data as props
* Remove this separate function component and Inline all of the logic within the HasAccess component

I think I'm ok with this for now, because this is a really bad bug!

nickgros (approved):

src/lib/containers/HasAccessV2.tsx
nickgros: Lets enable this lint rule!! https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

nickgros: https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1782

PR #1787, opened 2022-09-16
PORTALS-2322: Simplify SCSS build

nickgros:
There is no need to compile our SCSS separately. Let esbuild/vite load and compile it.

This supports migration to Yarn berry because we don't have to configure a standalone sass compiler to be PnP-aware when our bundler(s) will already be able to handle it.


src/lib/umd.index.ts
```diff
@@ -44,14 +44,16 @@ import UserCard from './containers/UserCard'
...
 import { SynapseClient, SynapseConstants } from './utils'
 import {
   SynapseContextConsumer,
   SynapseContextProvider,
   useSynapseContext,
 } from './utils/SynapseContext'
 
+// Also include scss in the bundle
+import './style/main.scss'
+
```
nickgros, line 56: It was kinda hiding--let's make it a bit more obvious that this is also the entrypoint for our scss

jay-hodgson (approved):

PR #1792, opened 2022-09-20
SWC-6281

nickgros:
https://sagebionetworks.jira.com/browse/SWC-6281


src/lib/containers/FluidModal.tsx
```diff
@@ -60,7 +60,7 @@ export const FluidModal = (props: FluidModalProps) => {
     >
       <Modal.Header closeButton>
         <Modal.Title>
-          <Typography variant={'headline1'}>
+          <Typography variant={'headline1'} component="span">
```
nickgros, line 63: This should be a `span`, not a `p`, because it can contain other elements.

jay-hodgson (approved):

PR #1794, opened 2022-09-21
BREAKING CHANGE: Specify downloadCartPageUrl in context

nickgros:
When we add the Synapse Table renderer to the markdown component (PORTALS-2314), we need to specify the download cart page URL to ensure it links to the correct place. Instead of passing this information through the markdown renderer, it should be in global context.

### DIFF
--- a/src/__tests__/lib/containers/DownloadConfirmation.test.tsx
+++ b/src/__tests__/lib/containers/DownloadConfirmation.test.tsx
@@ -109,12 +109,6 @@ describe('DownloadConfirmation', () => {
     jest.clearAllMocks()
   })
 
-  it('should render without crashing with just a cancel button', () => {
-    renderComponent(props)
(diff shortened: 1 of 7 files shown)


src/__tests__/lib/containers/DownloadConfirmation.test.tsx
```diff
@@ -109,12 +109,6 @@ describe('DownloadConfirmation', () => {
     jest.clearAllMocks()
   })
 
-  it('should render without crashing with just a cancel button', () => {
-    renderComponent(props)
-    screen.getByRole('alert')
-    screen.getByRole('button', { name: 'Cancel' })
-  })
-
```
nickgros, line 117: Superfluous test that yielded a `not wrapped in act` warning

jay-hodgson (approved):

PR #1797, opened 2022-09-22
PORTALS-2302: add background shade in facets area

yjyoo773:
Screen Shot 2022-09-22 at 9 41 38 AM


nickgros:
@yjyoo773 could you also check these two things that seem to be present in the design? Note--not sure if the components I'm referring to are the ones where the style needs to change
* A bottom border for the top level controls (the box containing TOOLS (175)...)
* No margin between the top level controls and the facets area

jay-hodgson (approved):

PR #1809, opened 2022-09-29
PORTALS-2316: Replace sanitize-html with xss

nickgros:
We're using xss to clean HTML in SWC, which is more configurable. Let's simplify our configs by using the same library. I also have more confidence in the SWC configuration, which we have been been using to clean arbitrary HTML for far longer.

This also brings the bundle size down from 3.06 MB to 2.91 MB

### DIFF
--- a/src/lib/containers/markdown/MarkdownSynapse.tsx
+++ b/src/lib/containers/markdown/MarkdownSynapse.tsx
@@ -1,7 +1,8 @@
+import React from 'react'
 import MarkdownIt from 'markdown-it'
-import * as React from 'react'
-import sanitizeHtml from 'sanitize-html'
+import xss from 'xss'
 import { SynapseClient } from '../../utils'
+import { xssOptions } from '../../utils/functions/SanitizeHtmlUtils'
 import { SynapseClientError } from '../../utils/SynapseClientError'
 import { SynapseContext } from '../../utils/SynapseContext'
 import {
@@ -179,63 +180,7 @@ export default class MarkdownSynapse extends React.Component<
     const initText = this.props.renderInline
       ? this.state.md.renderInline(markdown)
       : this.state.md.render(markdown)
-    const cleanText = sanitizeHtml(initText, {
-      allowedAttributes: {
-        a: ['href', 'target'],
-        button: ['class'],
-        div: ['class'], // PORTALS-1450: including 'style' in the allow-list will cause string values to come through, which crashes the app when used (because it uses jsx).
-        h1: ['toc'],
-        h2: ['toc'],
-        h3: ['toc'],
-        h4: ['toc'],
-        h5: ['toc'],
-        h6: ['toc'],
-        li: ['class'],
-        ol: ['class'],
-        span: ['*'],
-        table: ['class'],
-        th: ['colspan'],
-        thead: ['class'],
-        ul: ['class'],
-        img: ['src', 'alt'],
-      },
-      allowedTags: [
-        'span',
-        'code',
-        'h1',
-        'h2',
-        'h3',
-        'h4',
-        'h5',
-        'p',
-        'b',
-        'i',
-        'em',
-        'strong',
-        'a',
-        'id',
-        'table',
-        'tr',
-        'td',
-        'tbody',
-        'th',
-        'thead',
-        'button',
-        'div',
-        'img',
-        'image',
-        'ol',
-        'ul',
-        'li',
-        'svg',
-        'g',
-        'br',
-        'hr',
-        'summary',
-        'details',
-        'strong',
-      ],
-    })
+    const cleanText = xss(initText, xssOptions)
     return { __html: cleanText }
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/containers/markdown/MarkdownSynapse.tsx
```diff
@@ -179,63 +180,7 @@ export default class MarkdownSynapse extends React.Component<
...
-        div: ['class'], // PORTALS-1450: including 'style' in the allow-list will cause string values to come through, which crashes the app when used (because it uses jsx).
-        h1: ['toc'],
-        h2: ['toc'],
-        h3: ['toc'],
-        h4: ['toc'],
-        h5: ['toc'],
-        h6: ['toc'],
-        li: ['class'],
-        ol: ['class'],
-        span: ['*'],
```
nickgros, line 195: I confirmed that event handlers that could run scripts (e.g. `onclick`) were removed, but this `*` still makes me deeply uncomfortable.

src/lib/containers/markdown/MarkdownSynapse.tsx
```diff
@@ -179,63 +180,7 @@ export default class MarkdownSynapse extends React.Component<
...
-        'h3',
-        'h4',
-        'h5',
-        'p',
-        'b',
-        'i',
-        'em',
-        'strong',
-        'a',
-        'id',
```
nickgros, line 216: Removed--this is not a known HTML tag

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
+import { IFilterXSSOptions } from 'xss'
+
+// PORTALS-1450: including 'style' in the allow-list will cause string values to come through, which crashes the app when used (because it uses jsx).
+export const xssOptions: IFilterXSSOptions = {
```
nickgros, line 4: Copied from SWC with a few additions per to match the existing `santize-html` config

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+    aside: [],
+    audio: ['autoplay', 'controls', 'loop', 'preload', 'src'],
+    b: [],
+    bdi: ['dir'],
+    bdo: ['dir'],
+    big: [],
+    blockquote: ['cite'],
+    body: [],
+    br: [],
+    button: ['class'],
```
nickgros, line 20: added tag

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+    h5: ['toc'],
+    h6: ['toc'],
+    head: [],
+    header: [],
+    hr: [],
+    html: [],
+    i: [],
+    img: ['src', 'alt', 'title', 'width', 'height'],
+    ins: ['datetime'],
+    li: ['class'],
```
nickgros, line 50: added attr `'class'`

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+    pre: [],
+    s: [],
+    section: [],
+    small: [],
+    span: ['data-widgetparams', 'class', 'id'],
+    sub: [],
+    summary: [],
+    sup: [],
+    strong: [],
+    svg: [],
```
nickgros, line 65: added tag

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+    table: ['width', 'border', 'align', 'valign', 'class'],
+    tbody: ['align', 'valign'],
+    td: ['width', 'rowspan', 'colspan', 'align', 'valign'],
+    tfoot: ['align', 'valign'],
+    th: ['width', 'rowspan', 'colspan', 'align', 'valign', 'class'],
+    thead: ['class', 'align', 'valign'],
+    tr: ['rowspan', 'align', 'valign'],
+    tt: [],
+    u: [],
+    ul: ['class'],
```
nickgros, line 75: added attr `'class'`

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+  },
+  safeAttrValue: function (tag, name, value) {
+    if (tag === 'img' && name === 'src') {
+      if (
+        value &&
+        (value.startsWith('data:image/') || value.startsWith('http'))
+      ) {
+        return value
+      } else {
+        return ''
```
nickgros, line 103: This return did not exist in SWC (effectively returning `undefined`, but caused a type error here because this function must return a string. Returning an empty string preserves the same effective behavior.

src/lib/utils/functions/SanitizeHtmlUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  },
+  stripIgnoreTagBody: true, // filter out all tags not in the whitelist
+  allowCommentTag: false,
+  css: false,
+  onIgnoreTag: function (tag, html, options) {
+    if (tag === '!doctype') {
+      // do not filter doctype
+      return html
+    }
+    return undefined
```
nickgros, line 94: Added this return because our tsconfig requires that all code paths return a value

jay-hodgson (approved):

PR #1811, opened 2022-09-29
PORTALS-2335

yjyoo773:
Screen Shot 2022-09-29 at 9 02 46 AM


nickgros (changes requested):
Fix the query key and add tests for ForumTable and this will be good to go

src/lib/containers/discussion_forum/ForumTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+                  }
+                  onClick={() =>
+                    onSort(DiscussionThreadOrder.NUMBER_OF_REPLIES)
+                  }
+                />
+              </span>
+            </th>
+            <th>
+              <span className="SRC-split">
+                <span>View</span>
```
nickgros, line 101:
```suggestion
                <span>Views</span>
```

nickgros (changes requested):
One super picky change to try to keep our test logs clean (which we historically have not been good at). Otherwise looks fantastic

nickgros (approved):

PR #1812, opened 2022-09-29
EntityBundle type narrowing based on request

nickgros:
Attempt to make the `EntityBundle` type more strict based off of a passed `EntityBundleRequest` constant.


src/lib/containers/EntityBadgeIcons.tsx
```diff
@@ -109,8 +110,8 @@ export const EntityBadgeIcons = (props: EntityBadgeIconsProps) => {
 
   const { data: bundle } = useGetEntityBundle(
     entityId,
-    undefined,
     versionNumber,
+    ALL_ENTITY_BUNDLE_FIELDS,
```
nickgros, line 114: Reordered the arguments for `useGetEntityBundle`.

It was `(entityId, fields, versionNumber, queryOptions)`, now it's `(entityId, versionNumber, fields, queryOptions)`

src/lib/utils/synapseTypes/EntityBundle.ts
```diff
@@ -45,3 +51,64 @@ export type EntityBundleRequest = {
...
+ * Otherwise, the resulting type is never, indicating that the field is not present in the result.
+ */
+type IncludeOnlyIfRequested<
+  TRequestConst extends Record<string, boolean>,
+  TRequestField extends keyof TRequestConst,
+  TResultObject,
+  TResultField extends keyof TResultObject,
+> = TRequestConst[TRequestField] extends true
+  ? TResultObject[TResultField]
+  : never
```
nickgros, line 95: This might cause other problems with dynamic request objects or not passing a generic (neither of which we are doing today), so I'm wondering if this should be `TResultObject[TResultField] | undefined`, at the cost of losing the strictness of `never`

nickgros: Could also consider

```
TRequestConst[TRequestField] extends true
  ? TResultObject[TResultField] :
 	(TResultObject[TResultField] extends boolean ? 
	 TResultObject[TResultField] | undefined
	  : never)
```

src/lib/containers/HasAccessV2.tsx
```diff
@@ -111,11 +111,11 @@ export function useGetFileHandleDownloadType(
   const { accessToken } = useSynapseContext()
   const { data: entityBundle, error: entityFetchError } = useGetEntityBundle(
     entityId,
+    parseInt(entityVersionNumber ?? ''),
```
jay-hodgson, line 114: Is NaN (parseInt('') handled properly by useGetEntityBundle()?

nickgros: Not sure, I'll change it to make it more clear/safe

src/lib/utils/synapseTypes/EntityBundle.ts
jay-hodgson: Difficult to read with the nested ternary operators, but logic is sound

src/lib/utils/synapseTypes/EntityBundle.ts
```diff
@@ -45,3 +51,64 @@ export type EntityBundleRequest = {
...
+
+/**
+ * Bundle to transport an Entity and related data objects. See [Synapse REST API](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundle.html).
+ *
+ * We have utilized TypeScript to add additional type safety based on the value of the request object.
+ * @typeParam T - A constant EntityBundleRequest object.
+ * @returns An EntityBundle type that only includes the fields that are requested in the request object.
+ */
+export type EntityBundle<
```
jay-hodgson, line 104: Very clever type definition

jay-hodgson (approved):

jay-hodgson:
There will be a bit more work to do if/when the entity bundle is extended, but I think it's well worth the effort for all of the compile-time type safety that we get from this enhancement.  Thanks @nickgros !

src/lib/utils/synapseTypes/EntityBundle.ts
nickgros: Unfortunately TypeScript doesn't give us a ton of tools to make types more readable 😵‍💫

PR #1813, opened 2022-10-03
PORTALS-2316: HtmlPreview component

nickgros:
Adds HtmlPreview component
- Given raw HTML and a file handle, renders the raw HTML (if the user is on the trusted team) or renders sanitized HTML.
- If the rendered HTML is different than the raw HTML, the user can open the untrusted content in a new window.

- Component has not yet been exposed via MarkdownSynapse - this PR does not include: 
  - Parsing the `preview` widget descriptor
  - Fetching the file referenced by the preview descriptor
  - Asking the user to log in if they can't download the referenced file as Anonymous
  - Determining how to preview the file (e.g. download the file contents or a preview file handle?)
  - Downloading the file handle contents.

image

### DIFF
--- a/src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
+++ b/src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
@@ -1,0 +1,76 @@
+import React, { useMemo } from 'react'
+import xss from 'xss'
+import { xssOptions } from '../../../utils/functions/SanitizeHtmlUtils'
+import { useGetTeamMembers } from '../../../utils/hooks/SynapseAPI/team/useTeamMembers'
+import { TRUSTED_HTML_USERS_TEAM_ID } from '../../../utils/SynapseConstants'
+import { FileHandle } from '../../../utils/synapseTypes'
+import { SynapseSpinner } from '../../LoadingScreen'
+import { SanitizedWarning } from './SanitizedWarning'
+
+/**
+ * @param options
+ * @returns HTML that is cleaned if it is not trusted. undefined if `isLoading` is true.
+ */
+function useCleanHtml(options: {
+  rawHtml: string
+  isLoading: boolean
+  isTrusted: boolean
+}): string | undefined {
+  const { rawHtml, isLoading, isTrusted } = options
+  return useMemo(() => {
+    if (isLoading) {
+      return undefined
+    }
+    if (isTrusted) {
+      return rawHtml
+    } else {
+      return xss(rawHtml, xssOptions)
+    }
+  }, [isLoading, isTrusted, rawHtml])
+}
+
+export type HtmlPreviewProps = {
+  fileHandle: FileHandle
+  rawHtml: string
+}
+
+/**
+ * Renders raw HTML. Uses file handle data to determine if the content should be sanitized.
+ * @param props
+ * @returns
+ */
+export default function HtmlPreview(props: HtmlPreviewProps) {
+  const { fileHandle, rawHtml } = props
+  const { data: trustedHtmlUsersData, isLoading } = useGetTeamMembers(
+    TRUSTED_HTML_USERS_TEAM_ID,
+  )
+
+  const trustedHtmlUserIds = (trustedHtmlUsersData?.results ?? []).map(
+    member => member.member.ownerId,
+  )
+
+  const htmlIsCreatedByTrustedUser = trustedHtmlUserIds.includes(
+    fileHandle.createdBy,
+  )
+
+  const cleanHtml = useCleanHtml({
+    rawHtml,
+    isLoading,
+    isTrusted: htmlIsCreatedByTrustedUser,
+  })
+
+  if (isLoading) {
+    return <SynapseSpinner />
+  }
+
+  return (
+    <>
+      {rawHtml !== cleanHtml && <SanitizedWarning rawHtml={rawHtml} />}
+      <iframe srcDoc={cleanHtml} style={{ border: 0, width: '100%' }} />
+    </>
+  )
+}
+
+export const EXPORTED_FOR_UNIT_TESTING = {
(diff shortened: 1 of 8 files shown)


src/lib/utils/SynapseClient.ts
```diff
@@ -1779,11 +1780,14 @@ export const getFileHandleContentFromID = (
 export const getFileHandleContent = (
   fileHandle: FileHandle,
   presignedUrl: string,
+  maxFileSizeBytes?: number,
```
nickgros, line 1783:
```suggestion
```

src/lib/utils/SynapseClient.ts
```diff
@@ -1779,11 +1780,14 @@ export const getFileHandleContentFromID = (
...
 ): Promise<string> => {
   // get the presigned URL, download the data, and send that back (via resolve())
   return new Promise((resolve, reject) => {
-    // sanity check!  must be less than 5MB
-    if (fileHandle.contentSize < MAX_JS_FILE_DOWNLOAD_SIZE) {
+    // sanity check!  must be less than 5MB (unless overridden)
+    if (
+      fileHandle.contentSize < (maxFileSizeBytes ?? MAX_JS_FILE_DOWNLOAD_SIZE)
+    ) {
```
nickgros, line 1790:
```suggestion
    // sanity check!  must be less than 5MB
    if (fileHandle.contentSize < MAX_JS_FILE_DOWNLOAD_SIZE) {
```

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { data: trustedHtmlUsersData, isLoading } = useGetTeamMembers(
+    TRUSTED_HTML_USERS_TEAM_ID,
+  )
+
+  const trustedHtmlUserIds = (trustedHtmlUsersData?.results ?? []).map(
+    member => member.member.ownerId,
+  )
+
+  const htmlIsCreatedByTrustedUser = trustedHtmlUserIds.includes(
+    fileHandle.createdBy,
```
jay-hodgson, line 53: Instead of a fileHandle, the HtmlPreview could just take a userId

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Renders raw HTML. Uses file handle data to determine if the content should be sanitized.
+ * @param props
+ * @returns
+ */
+export default function HtmlPreview(props: HtmlPreviewProps) {
+  const { fileHandle, rawHtml } = props
+  const { data: trustedHtmlUsersData, isLoading } = useGetTeamMembers(
```
jay-hodgson, line 44: Will this get the first 50 members only?

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <>
+      {rawHtml !== cleanHtml && <SanitizedWarning rawHtml={rawHtml} />}
+      <iframe srcDoc={cleanHtml} style={{ border: 0, width: '100%' }} />
+    </>
+  )
+}
+
+export const EXPORTED_FOR_UNIT_TESTING = {
```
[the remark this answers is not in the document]
jay-hodgson: What do you think about starting a page that describes some of our design pattern decisions?  For example, it would contain our current thinking about the use (and testing) of hooks.

src/lib/utils/SynapseConstants.ts
```diff
@@ -118,3 +120,9 @@ export const ENTITY_HEADER_STORAGE_KEY = 'INFO_FROM_IDS_ENTITY_HEADER'
 
 export const DATETIME_UTC_COOKIE_KEY =
   'org.sagebionetworks.synapse.datetime.utc'
+
+export const BASE = 1024,
```
jay-hodgson, line 124: Looks like you started to think about the file size restriction, but I don't think it's mentioned in this PR

jay-hodgson: Related to a new ticket - https://sagebionetworks.jira.com/browse/SWC-6291
I'd like to use the SRC HtmlPreview in SWC

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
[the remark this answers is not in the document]
nickgros: I like the idea but am worried those practices change more frequently than we would update the page

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
nickgros: I think it's hard-capped at 20. Not an issue today since the team is not that large, but I can update this to fetch all members

src/lib/utils/SynapseConstants.ts
nickgros: No reason we couldn't export this and use it there. The file size limit shouldn't have been added here but it will be in the next PR.

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
nickgros: The team has 22 members so if it is hard-capped at 20, it will cause an issue today

nickgros: But the cap is indeed 50, so no issue. Still worth changing.

jay-hodgson (approved):

PR #1816, opened 2022-10-04
PORTALS-2316: Add FileHandleContentRenderer

nickgros:
Simply renders the contents of a file handle based on the predetermined type of preview shown.

 The next PR will include a component that will determine which file handle to render for a particular entity.

### DIFF
--- a/src/__tests__/lib/containers/markdown/FilePreview/HtmlPreview/HtmlPreview.test.tsx
+++ b/src/__tests__/lib/containers/markdown/FilePreview/HtmlPreview/HtmlPreview.test.tsx
@@ -10,21 +10,21 @@ import React from 'react'
 import HtmlPreview, {
   EXPORTED_FOR_UNIT_TESTING,
   HtmlPreviewProps,
-} from '../../../../../lib/containers/FilePreview/HtmlPreview/HtmlPreview'
-import { createWrapper } from '../../../../../lib/testutils/TestingLibraryUtils'
-import { TEAM_ID_MEMBER_ID } from '../../../../../lib/utils/APIConstants'
+} from '../../../../../../lib/containers/FilePreview/HtmlPreview/HtmlPreview'
+import { createWrapper } from '../../../../../../lib/testutils/TestingLibraryUtils'
+import { TEAM_ID_MEMBER_ID } from '../../../../../../lib/utils/APIConstants'
 import {
   BackendDestinationEnum,
   getEndpoint,
-} from '../../../../../lib/utils/functions/getEndpoint'
-import { TRUSTED_HTML_USERS_TEAM_ID } from '../../../../../lib/utils/SynapseConstants'
-import { TeamMember } from '../../../../../lib/utils/synapseTypes/TeamMember'
-import { rest, server } from '../../../../../mocks/msw/server'
+} from '../../../../../../lib/utils/functions/getEndpoint'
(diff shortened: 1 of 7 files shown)


src/__tests__/lib/containers/markdown/FilePreview/HtmlPreview/HtmlPreview.test.tsx
```diff
@@ -10,21 +10,21 @@ import React from 'react'
 import HtmlPreview, {
```
nickgros, line 10: Test file just moved to match main structure

src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
```diff
@@ -0,0 +1,34 @@
...
+      batchFileResult.requestedFiles[0].preSignedURL!,
+      maxFileSizeBytes,
+    )
+    return data
+  }
+  return useQuery<string, SynapseClientError>(
+    ['presignedUrlContent', fileHandle, request],
+    queryFn,
+    {
+      staleTime: Infinity,
```
nickgros, line 30: Configure cached data to never expire

src/lib/containers/FilePreview/FileHandleContentRenderer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  /** Informs how to render the file data */
+  previewType: PreviewRendererType
+}
+
+/**
+ * Fetches the content for and renders the contents of a file handle.
+ * @param props
+ * @returns
+ */
+export default function FileHandleContentRenderer(
```
nickgros, line 29: Pretty simple component-get the data for a file handle, then render it.

src/lib/containers/ErrorBanner.tsx
```diff
@@ -92,7 +89,7 @@ export const ErrorFallbackComponent: React.FunctionComponent<FallbackProps> = ({
   resetErrorBoundary,
 }) => {
   return (
-    <div role="alert" className="bootstrap-4-backport SRC-marginBottomTop">
+    <div className="bootstrap-4-backport SRC-marginBottomTop">
```
nickgros, line 92: This already contains an `alert`, so this role is not necessary

src/lib/containers/ErrorBanner.tsx
```diff
@@ -31,20 +31,17 @@ export const ClientError = (props: { error: SynapseClientError }) => {
     issuePriority: '3',
   })
 
-  return (
```
nickgros, line 34: Same logic, just more clear that the states are mutually exclusive

src/lib/containers/FilePreview/FileHandleContentRenderer.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const batchFileRequest: BatchFileRequest = {
+    requestedFiles: [fileHandleAssociation],
+    includePreSignedURLs: true,
+    includeFileHandles: false,
+    includePreviewPreSignedURLs: false,
+  }
+
+  const { data: content, isLoading } = useGetPresignedUrlContent(
+    fileHandle,
```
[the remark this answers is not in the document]
nickgros: While it would be nice to abstract the details of the request away, I don't want the component or the hook to be FileEntity only, so it could be reused in other places where we might need to download file handle contents (e.g. Wiki attachments).

PR #1818, opened 2022-10-04
PORTALS-2313: ability to set the loading UI row count

jay-hodgson:


jay-hodgson:
For large wikis (such as a dedicated About page in a portal), it would be nice to have a large number of lines while the content is loading.  What do you think about this functionality?

nickgros:
This will be great for making the Portals layout shifts a little less jarring. 

Perhaps not super important, but I think it would be slightly better if we added a SkeletonParagraph where the first n-1 lines are 100% width and the nth line is something like 20-60% width. Then in wikis specify the number of skeleton paragraphs to show. Not necessary to add this, though.

nickgros (approved):

jay-hodgson:
> This will be great for making the Portals layout shifts a little less jarring.
> 
> Perhaps not super important, but I think it would be slightly better if we added a SkeletonParagraph where the first n-1 lines are 100% width and the nth line is something like 20-60% width. Then in wikis specify the number of skeleton paragraphs to show. Not necessary to add this, though.

That's a good idea, then this could be changed to be a loadingUiParagraphCount (for example)

PR #1820, opened 2022-10-05
PORTALS-2313: render column value, but use rowId when creating link href

jay-hodgson:


jay-hodgson:
In the ARK portal I want to link the Dataset name column to the DatasetDetails page (where we will show the dataset info and dataset members using the Synapse ID from the search params)

src/lib/containers/CardContainerLogic.tsx
```diff
@@ -41,8 +41,11 @@ export interface CardLink {
   baseURL: string
   // the key that will go into the url
   URLColumnName: string
-  // the column name who's value will be used
+  // the column name who's value will be used as the display text
   matchColumnName: string
+  // If set, use the rowID as the column value.  Otherwise, use the value in the matchColumnName column
+  overrideValueWithRowID?: boolean
```
nickgros, line 47: In general this entire interface is confusing to me...I would love if we could come up with a new one that accurately captures the complexity we need. But that's a project for a later day

nickgros (approved):

PR #1824, opened 2022-10-06
PORTALS-2340: propagate scroll event inside the react-flow graph to window

jay-hodgson:


src/lib/containers/provenance/ProvenanceGraph.tsx
```diff
@@ -375,6 +375,15 @@ const ProvenanceReactFlow = (props: ProvenanceProps) => {
     onEdgesChangedListener,
   ])
 
+  const onPaneScrollFunction: (event?: React.WheelEvent) => void = useCallback(
+    event => {
```
jay-hodgson, line 379: Is there a better way to do this.  I don't want this function to be recreated on each render so I utilized useCallback, but I don't really have a valid dependency array (something that causes re-rendering)

src/lib/containers/provenance/ProvenanceGraph.tsx
```diff
@@ -375,6 +375,15 @@ const ProvenanceReactFlow = (props: ProvenanceProps) => {
...
+      if (event && event.deltaY && rootEntityRefs.length > 0) {
+        window.scrollTo(window.scrollX, window.scrollY + event?.deltaY)
+      }
+    },
+    [rootEntityRefs],
```
jay-hodgson, line 384: So I artificially depended on rootEntityRefs

jay-hodgson:
Proposing we force the user to use the Controls (or pinch) to zoom in and out instead of scrolling.

src/lib/containers/provenance/ProvenanceGraph.tsx
nickgros: If this truly does not depend on rootEntityRefs, does it work if you remove ` && rootEntityRefs.length > 0` and set the dependencies to `[]`? It doesn't seem like this function needs to be re-created, but maybe I'm misunderstanding.

Alternatively, you could define the function outside of the component.

jay-hodgson: Removed dependency.  This solution works _pretty_ well.  If you try to scroll inside of a node, it still eats the event.  So there still might be some frustration when attempting to scroll past the graph, but with this change it's now likely that you'll be able to scroll to where you want to be (instead of accidentally zooming in/out the graph).

nickgros (approved):

PR #1825, opened 2022-10-07
PORTALS-2316: FileEntityPreview

nickgros:
Add FileEntityPreview and PreviewUtils


src/lib/containers/FilePreview/FileEntityPreview.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * content type and file name of the file handle(s) associated with the FileEntity.
+ * @param props
+ */
+export default function FileEntityPreview(props: FileEntityPreviewProps) {
+  const { bundle } = props
+  const { accessToken } = useSynapseContext()
+  const isLoggedIn = !!accessToken
+
+  if (!isFileEntity(bundle.entity)) {
+    throw new Error(`Entity ${bundle.entity.id!} is not a FileEntity`)
```
nickgros, line 34: Let's wrap all Markdown widgets in an error boundary (coming in next PR). There are too many places where things can go wrong in each layer, and it doesn't make sense to duplicate error handling logic in all of those places. Much easier to just throw an Error.

src/lib/containers/FilePreview/FileEntityPreview.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <Alert
+      dismissible={false}
+      show={true}
+      variant={'danger'}
+      transition={false}
+    >
+      Could not render a preview for entity: {bundle.entity.id!}
+    </Alert>
+  )
+}
```
nickgros, line 115:
```suggestion
  // We can't render the file, so throw an Error so the ErrorBoundary will show an alert.
  throw new Error(`Could not render a preview for entity: ${bundle.entity.id!}`)
```

src/lib/containers/FilePreview/FileEntityPreview.tsx
jay-hodgson: I agree

src/lib/containers/FilePreview/PreviewUtils.ts
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Given a file handle that represents a link to original data, return the type of renderer that we could use to
+ * display the file contents in the browser.
+ *
+ * @returns the PreviewRendererType that would be used to render the file contents, or NONE if the file cannot be rendered
+ */
+export function getOriginalFileHandleRenderer(
+  dataFileHandle: FileHandle,
+): PreviewRendererType {
+  // TODO: Could this be changed to CloudProviderFileHandle?
```
jay-hodgson, line 89: I believe so if the auth check has been delegated to Synapse for all CloudProviderFileHandle types

jay-hodgson (approved):

PR #1826, opened 2022-10-07
PORTALS-2316: Add EntityPreview and register it with Markdown processor

nickgros:
Depends on #1825 

* Add `EntityPreview`. `EntityPreview` takes a SynID and renders the appropriate preview based on the Entity type. For now, it only handles `FileEntity` and `Link`, if the `Link` resolves to a `FileEntity`.
* Add `MarkdownEntityPreview` which renders an `EntityPreview`, and register the component to handle the `preview` widget descriptor

### DIFF
--- a/src/lib/containers/FilePreview/EntityPreview.tsx
+++ b/src/lib/containers/FilePreview/EntityPreview.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+import { useResolveLinkEntity } from '../../utils/functions/ResolveLinkEntity'
+import useGetEntityBundle from '../../utils/hooks/SynapseAPI/entity/useEntityBundle'
+import { ALL_ENTITY_BUNDLE_FIELDS } from '../../utils/SynapseConstants'
+import { FileEntity, isFileEntity } from '../../utils/synapseTypes'
+import { SynapseSpinner } from '../LoadingScreen'
+import FileEntityPreview from './FileEntityPreview'
+
+export type EntityPreviewProps = {
+  entityId: string
+  versionNumber?: number
+}
+
+/**
+ * Returns a preview of an entity. Currently only works for FileEntities, and Links that resolve to FileEntities.
+ * @param props
+ * @returns
+ */
+export default function EntityPreview(props: EntityPreviewProps) {
+  let { entityId, versionNumber } = props
+
+  // If the entity is a Link, follow it until we get a non-link
+  const { data: resolvedLink, isLoading: isLoadingLinkResolution } =
+    useResolveLinkEntity(entityId, versionNumber)
+  const isResolvedLinkFileEntity = resolvedLink && isFileEntity(resolvedLink)
+
+  const { data: bundle, isLoading: isLoadingBundle } = useGetEntityBundle(
+    resolvedLink?.id!,
+    (resolvedLink as FileEntity | undefined)?.versionNumber,
+    ALL_ENTITY_BUNDLE_FIELDS,
+    {
+      enabled: !!isResolvedLinkFileEntity,
+      useErrorBoundary: true,
+    },
+  )
+
+  const isLoading = isLoadingLinkResolution || isLoadingBundle
+
+  if (isLoading) {
+    return <SynapseSpinner />
+  }
+
+  // If the entity is NOT a FileEntity, throw an error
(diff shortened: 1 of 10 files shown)


src/lib/utils/functions/ResolveLinkEntity.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @returns the ID of the entity at the end of the chain of links
+ * @throws an error if a link is found that does not resolve, or if a cycle is found
+ */
+export async function resolveLinkEntity(
+  synId: string,
+  versionNumber?: number,
+  accessToken?: string,
+): Promise<Entity> {
+  // Track visited entities to prevent infinitely fetching a cycle of Links
+  const seen: Set<string> = new Set()
```
nickgros, line 18: I don't think we have cycle detection in SWC, but adding it here

src/lib/containers/FilePreview/EntityPreview.tsx
```diff
@@ -0,0 +1,49 @@
...
+  entityId: string
+  versionNumber?: number
+}
+
+/**
+ * Returns a preview of an entity. Currently only works for FileEntities, and Links that resolve to FileEntities.
+ * @param props
+ * @returns
+ */
+export default function EntityPreview(props: EntityPreviewProps) {
```
nickgros, line 19: Just takes a synID and ensures it's a file or a link that points to a file

src/lib/containers/markdown/MarkdownWidget.tsx
```diff
@@ -104,8 +114,19 @@ export default function MarkdownWidget(props: MarkdownWidgetProps) {
...
 }
+
+export default function MarkdownWidgetWithWrapper(props: MarkdownWidgetProps) {
+  // Wrap the widget in an error boundary.
+  return (
+    <SynapseErrorBoundary>
+      <MarkdownWidget {...props} />
+    </SynapseErrorBoundary>
+  )
+}
```
nickgros, line 132: Add an error boundary so we can throw Errors with reckless abandon

src/lib/containers/entity_finder/EntityFinder.tsx
```diff
@@ -168,19 +168,9 @@ export const EntityFinder: React.FunctionComponent<EntityFinderProps> = ({
 
   useEffect(() => {
     if (searchTerms?.length === 1) {
-      const synIdMatch = SYNAPSE_ENTITY_ID_REGEX.exec(searchTerms[0])
```
nickgros, line 171: Refactored/moved to RegularExpressions.ts, and reused by MarkdownEntityPreview

src/lib/containers/FilePreview/EntityPreview.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { useResolveLinkEntity } from '../../utils/functions/ResolveLinkEntity'
+import useGetEntityBundle from '../../utils/hooks/SynapseAPI/entity/useEntityBundle'
+import { ALL_ENTITY_BUNDLE_FIELDS } from '../../utils/SynapseConstants'
+import { FileEntity, isFileEntity } from '../../utils/synapseTypes'
+import { SynapseSpinner } from '../LoadingScreen'
+import FileEntityPreview from './FileEntityPreview'
+
+export type EntityPreviewProps = {
+  entityId: string
```
jay-hodgson, line 10: Could be a "Reference" if you want to reuse the backend data model object

src/lib/containers/FilePreview/EntityPreview.tsx
jay-hodgson: I see how this works for a Link that points to a FileEntity, but I don't see how this works for an ID that points to a FileEntity.  Could you explain?

jay-hodgson: Oh, I see.  It looks like resolveLinkEntity() will get the FileEntity and just return it

jay-hodgson (approved):

src/lib/containers/FilePreview/EntityPreview.tsx
jay-hodgson: Then again, if this entityId could have the dot notation then this is not a good fit.

PR #1828, opened 2022-10-10
PORTALS-2369

nickgros:


src/__tests__/lib/containers/UserCardTests.test.tsx
```diff
@@ -57,314 +57,322 @@ const renderAvatar = (props: Omit<UserCardProps, 'size'>) => {
...
+      await screen.findByRole('img')
+    })
 
-  it('renders a small card', () => {
-    renderSmallUserCard(props)
-    screen.getByText('@' + MOCK_USER_NAME)
-  })
+    it('renders a small card', async () => {
+      renderSmallUserCard(props)
+      await screen.findByText('@' + MOCK_USER_NAME)
```
nickgros, line 82: Use await ...find instead of get

src/__tests__/lib/containers/UserCardTests.test.tsx
```diff
@@ -57,314 +57,322 @@ const renderAvatar = (props: Omit<UserCardProps, 'size'>) => {
...
-      { exact: false },
-    )
-    await screen.findByText('ORCID', { exact: false })
+    it('renders a large card', async () => {
+      renderLargeUserCard(props)
+      await screen.findByText(
+        mockUserProfileData.firstName + ' ' + mockUserProfileData.lastName,
+        { exact: false },
+      )
+      await screen.findByText('ORCID', { exact: false })
```
nickgros, line 100: `await waitFor(() => screen.get...` => `await screen.find...`

src/__tests__/lib/containers/UserCardTests.test.tsx
```diff
@@ -57,314 +57,322 @@ const renderAvatar = (props: Omit<UserCardProps, 'size'>) => {
...
+      await screen.findByText(`@${mockUserProfileData.userName}`)
+      expect(screen.queryByRole('img')).not.toBeInTheDocument()
+    })
 
-    // There is no medium user card, so we shouldn't be able to find the full name anywhere
-    expect(
-      screen.queryByText(
+    it('shows a medium user card when mouse enters', async () => {
+      jest.useFakeTimers()
+      const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
```
nickgros, line 200: Changing
```
      const user = userEvent.setup({ advanceTimers: jest.runAllTimers })
```
to
```
      const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
```
 seemed to fix a `not wrapped in act` warning

src/lib/containers/StorybookComponentWrapper.tsx
```diff
@@ -22,10 +25,16 @@ export async function sessionChangeHandler() {
...
     if (err instanceof SynapseClientError && err.status === 401) {
+      console.error(
+        'Encountered error fetching profile: ',
+        err,
+        'Signing out...',
+      )
       await signOut(() => {})
       accessToken = undefined
     }
-    console.error('Encountered error fetching profile: ', err, 'Signing out...')
```
nickgros, line 28: Move the error message inside, otherwise it was inaccurate (we didn't necessarily sign out)

src/lib/containers/StorybookComponentWrapper.tsx
```diff
@@ -22,10 +25,16 @@ export async function sessionChangeHandler() {
...
+    // Otherwise rethrow
+    throw err
```
nickgros, line 37: this should never happen, so I think it's ok to throw in storybook

src/lib/containers/StorybookComponentWrapper.tsx
```diff
@@ -50,16 +61,20 @@ export const StorybookComponentWrapper = props => {
...
+  const context: SynapseContextType = useMemo(
+    () => ({
+      accessToken: accessToken,
+      isInExperimentalMode: SynapseClient.isInSynapseExperimentalMode(),
+      utcTime: SynapseClient.getUseUtcTimeFromCookie(),
+      withErrorBoundary: true,
+      downloadCartPageUrl: '/?path=/story/download-downloadcartpage--demo',
+    }),
+    [accessToken],
+  )
```
nickgros, line 74: Try to stabilize this, just in case

src/lib/containers/UserCard.tsx
```diff
@@ -62,95 +67,75 @@ export const UserCard: React.FunctionComponent<UserCardProps> = (
...
-        },
-      )
-    }
-    return () => {
-      isCanceled = true
-    }
-  }, [userProfile, alias, accessToken])
+  const isLoading = isLoadingAliasLookup || isLoadingProfile
 
-  function Card(props: {
```
nickgros, line 123: This fn component was defined inside of the UserCard fn component, so we probably had excessive re-renders here.

src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
```diff
@@ -32,3 +32,31 @@ export function useGetPresignedUrlContent(
...
+export function useGetProfileImage(
+  userId: string,
+  options?: Omit<UseQueryOptions<Blob | null, SynapseClientError>, 'staleTime'>,
+) {
+  const queryFn = async () => {
+    const presignedUrl = await SynapseClient.getProfilePicPreviewPresignedUrl(
+      userId,
+    )
+    if (presignedUrl) {
+      const data = await fetch(presignedUrl, {
```
nickgros, line 45:
```suggestion
      // Fetch the presigned URL right away because it will expire
      const data = await fetch(presignedUrl, {
```

src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
```diff
@@ -32,3 +32,31 @@ export function useGetPresignedUrlContent(
     },
   )
 }
+
+export function useGetProfileImage(
```
nickgros, line 36:
```suggestion
/**
 * Get a blob containing the image data for the avatar of a Synapse user. Returns null if the user does not have a profile image.
 */
export function useGetProfileImage(
```

src/lib/utils/hooks/usePreFetchResource.ts
```diff
@@ -19,15 +18,16 @@ export function releaseResourceUrl(resourceUrl: string) {
 export default function usePreFetchResource(
```
nickgros, line 18: Split this up. Now, this hook takes a URL and returns a blob.

New fn `useCreateUrlForData` takes a Blob and returns a data URL.

src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
jay-hodgson: Nice comment, recommend this is committed.

src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
jay-hodgson: Nice comment, recommend this is committed.

PR #1829, opened 2022-10-10
PORTALS-2339: Expandable table data cells

nickgros:
- Add ExpandableTableDataCell, which shows a button to toggle show/hide overflow in a table cell in SynapseTable
- Remove ts-jest. I had to modify the Jest config with these changes, so I decided to simplify the configuration. I think this may also speed up the full test suite

https://user-images.githubusercontent.com/17580037/194906597-b90d36bf-5b84-4f56-8210-457339fddae7.mov


package.json
```diff
@@ -22,6 +22,7 @@
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@popperjs/core": "^2.10.2",
     "@react-google-maps/api": "^2.13.1",
+    "@react-hook/resize-observer": "^1.2.6",
```
nickgros, line 25: Way easier to use for this case than SizeMe (which doesn't have hooks)

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -262,22 +255,7 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
     case ColumnType.INTEGER:
     case ColumnType.BOOLEAN:
     case ColumnType.LARGETEXT: {
-      const isShort = isShortString(columnValue)
```
nickgros, line 265: No need to do all of this truncation

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -262,22 +255,7 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
...
-      if (isShort) {
-        return <p className={isBold}> {columnValue}</p>
-      } else {
-        return (
-          <div className={isBold}>
-            <ElementWithTooltip
```
nickgros, line 271: No more tooltip

src/lib/containers/synapse_table_functions/SynapseTableCell.tsx
```diff
@@ -262,22 +255,7 @@ export const SynapseTableCell: React.FC<SynapseTableCellProps> = ({
     case ColumnType.INTEGER:
     case ColumnType.BOOLEAN:
     case ColumnType.LARGETEXT: {
-      const isShort = isShortString(columnValue)
-      if (isShort) {
-        return <p className={isBold}> {columnValue}</p>
-      } else {
-        return (
-          <div className={isBold}>
```
nickgros, line 270: Dropped this div (which I'm guessing might have been used to style the tooltip?)

jest.config.js
```diff
@@ -1,26 +1,19 @@
...
+const esModules = [
+  '@juggle/resize-observer',
+  '@react-hook',
+  'lodash-es',
+  'nanoid',
+].join('|')
+
+/** @type {import('jest').Config} */
 module.exports = {
-  preset: 'ts-jest',
```
nickgros, line 3: Probably didn't need to remove ts-jest but I think it will speed things up, and simplify the configuration here

jest.config.js
```diff
@@ -1,26 +1,19 @@
...
   testEnvironment: 'jsdom',
   moduleNameMapper: {
     '\\.(css|less|scss)$': 'identity-obj-proxy',
     // Mock SVGs, loaded by SVGR https://react-svgr.com/docs/jest/
     '\\.svg$': '<rootDir>/src/mocks/svg.js',
   },
-  transformIgnorePatterns: [
-    '<rootDir>/node_modules/(?!(nanoid|lodash-es|jest*)/)',
-  ],
+  transformIgnorePatterns: [`/node_modules/(?!${esModules})`],
```
nickgros, line 16: Building the pattern using a list of modules is much more readable

src/lib/style/components/_expandable_table_data.scss
```diff
@@ -0,0 +1,27 @@
+.ExpandableTableData {
+  max-width: 250px;
+  vertical-align: middle;
+
+  p {
+    text-overflow: ellipsis;
+    margin: 0px 5px;
+  }
+
+  &[aria-expanded='true'] {
```
jay-hodgson, line 10: Maybe apply a css transition to this to smooth it out?

babel.config.js
```diff
@@ -1,4 +1,8 @@
 // Babel configuration currently only used by Jest.
 module.exports = {
-  presets: ['@babel/preset-env', '@babel/preset-react'],
+  presets: [
+    '@babel/preset-env',
+    ['@babel/preset-typescript', { allowDeclareFields: true }],
```
jay-hodgson, line 5: why allowDeclareFields=true?

jay-hodgson (approved):

babel.config.js
nickgros: We use the TS `declare` keyword in class components to use React context

src/lib/style/components/_expandable_table_data.scss
nickgros: I don't think that can be done in a simple way, since we're toggling styles that cannot be animated. We might be able to do some hacky stuff with `max-height`

PR #1831, opened 2022-10-11
SWC-6287

nickgros:
Option to show detailed error on 403

https://user-images.githubusercontent.com/17580037/195097937-8c171735-1247-4658-a70c-1ed22f1be08f.mov

### DIFF
--- a/src/lib/containers/ErrorBanner.tsx
+++ b/src/lib/containers/ErrorBanner.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { Alert, Button } from 'react-bootstrap'
 import {
   ErrorBoundary,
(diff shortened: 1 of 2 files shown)


src/lib/containers/ErrorBanner.tsx
```diff
@@ -42,7 +44,23 @@ export const ClientError = (props: { error: SynapseClientError }) => {
...
+        <Button
+          variant={'tertiary'}
+          style={{ fontSize: '12px' }}
+          onClick={() => setShowDetailedError(show => !show)}
+        >
+          {showDetailedError ? 'Hide' : 'Show'} details
+        </Button>
+        <div>
+          <Collapse in={showDetailedError}>
+            <code>{error.reason}</code>
```
jay-hodgson, line 59: I'm a little shocked that Design was ok with the code style.  I'm a little concerned that our syntax highlighter is going to find this code element and do more damage to the message style!
What do you think about putting this in an unassuming grey box instead?

nickgros: @jay-hodgson I didn't know we had a syntax highlighter for all `code` blocks. Could we change this to `pre`? I figure at some point a stack trace could end up in here so I'd like to have it preserve whitespace.

PR #1834, opened 2022-10-11
PORTALS-2346

yjyoo773:


src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -0,0 +1,171 @@
...
+const FOLLOWING_TEXT = 'You are following this topic. Click to stop following.'
+const UNFOLLOWING_TEXT = 'You are not following this topic. Click to follow.'
+
+export function DiscussionThread(props: DiscussionThreadProps) {
+  const { threadId, limit } = props
+
+  const { accessToken } = useSynapseContext()
+  const [datePosted, setDatePosted] = useState(true)
+  const [message, setMessage] = useState<string>()
+  const [subscribed, setSubscribed] = useState<Subscription>()
```
nickgros, line 42: There's a chance to reduce complexity here, and to reduce code duplication since we are fetching and updating a Subscription in multiple components.

Instead of having a state hook, an effect hook, and a complex event handler in this component (and the forum table component), can we extract a custom hook that would look something like this?

```tsx
const { subscription, onToggleSubscription, isLoading } = useSubscription(id, subscriptionObjectType);
```

Then this custom hook could handle the logic for fetching, subscribing, unsubscribing to a particular subscription, and the component just uses those fields that it returns.

It would be especially cool if it used `react-query` 🙂

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+const FOLLOWING_TEXT = 'You are following this topic. Click to stop following.'
+const UNFOLLOWING_TEXT = 'You are not following this topic. Click to follow.'
+
+export function DiscussionThread(props: DiscussionThreadProps) {
+  const { threadId, limit } = props
+
+  const { accessToken } = useSynapseContext()
+  const [datePosted, setDatePosted] = useState(true)
```
nickgros, line 40: Not clear that this is a sort. Can you rename to something like `sortRepliesAscending`?

nickgros (changes requested):

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const { data: threadData } = useGetThread(threadId)
+  const { mutate: updateSubscription } = usePostSubscription()
+  const { mutate: deleteSubscription } = useDeleteSubscription()
+
+  useEffect(() => {
+    getSubscription(accessToken, threadId, SubscriptionObjectType.THREAD).then(
+      result => setSubscribed(result),
+    )
+  }, [accessToken, threadId, handleFollowBtn])
```
yjyoo773, line 57: I tried using react-query, but when using it returns an empty array. So ended up using the useEffect and useState.

src/lib/utils/hooks/SynapseAPI/subscription/useSubscription.ts
```diff
@@ -0,0 +1,89 @@
...
+      subscriptionRequest,
+    )
+    return subscriptionList.results[0]
+  }
+  return useQuery<Subscription, SynapseClientError>(
+    ['subscription', objectId, objectType],
+    queryFn,
+    options,
+  )
+}
```
yjyoo773, line 44: As mentioned above this hook returns undefined.

nickgros (approved):

PR #1835, opened 2022-10-12
PORTALS-2368

jay-hodgson:


src/lib/containers/table/table-top/ProgrammaticOptions.tsx
```diff
@@ -1,60 +1,55 @@
...
+    isLoading: queryIsLoading,
+    error,
+    isPreviousData: newQueryIsFetching,
+  } = useGetQueryResultBundleWithAsyncStatus(
+    {
+      ...queryBundleRequest,
+      partMask: SynapseConstants.BUNDLE_MASK_COMBINED_SQL,
+    },
+    {
+      keepPreviousData: true,
```
nickgros, line 29:
```suggestion
      keepPreviousData: true,
      useErrorBoundary: true,
```

src/lib/containers/table/table-top/ProgrammaticOptions.tsx
```diff
@@ -1,60 +1,55 @@
...
+      // Change the SQL to "SELECT * ..."
+      const indexOfFrom = combinedSql.toUpperCase().indexOf('FROM SYN')
+      const selectStarTransformedSql = `SELECT * ${combinedSql.substring(
+        indexOfFrom,
+      )}`
+      setGeneratedSql(selectStarTransformedSql)
+    }
+  }, [combinedSql, isLoadingNewBundle])
+
+  useEffect(() => {
```
nickgros, line 48: Adding `useErrorBoundary` above will throw the error. Then you can delete this effect. Unless there's a case where there is an error only when `isLoadingNewBundle` is true?

nickgros (approved):
One place where you can replace an effect with a `useQuery` option

PR #1843, opened 2022-10-17
SWC-6265c: add limit for number of fetch retries

yjyoo773:


src/lib/utils/SynapseClient.ts
```diff
@@ -327,11 +327,16 @@ const fetchWithExponentialTimeout = async <TResponse>(
...
 ): Promise<TResponse> => {
   let response = await fetch(url, options)
+  let numOfTry = 1
   while (response.status && RETRY_STATUS_CODES.includes(response.status)) {
     await delay(delayMs)
     // Exponential backoff if we re-fetch
     delayMs = delayMs * 2
     response = await fetch(url, options)
+    numOfTry++
+    if (numOfTry == MAX_RETRY) {
```
jay-hodgson, line 337: The max_retry count should only apply to status codes 502 and 503.
429 is a Too Many Requests throttle response, 504 is a Gateway Timeout, and 0 is typically returned when the network connection has been interrupted. In these cases we should keep retrying (with an exponential backoff).

jay-hodgson (approved):

PR #1846, opened 2022-10-19
SWC-6265: prevent modal closing when error occurs

yjyoo773:
https://user-images.githubusercontent.com/60489495/196778967-306fc187-2d91-44bc-836d-5272ac3405e0.mov


nickgros:
I'd like if the second modal would put a shadow over the first modal. I think we're doing something with CSS to accomplish this in another place, so I'll track that down and get back to you

nickgros:
@yjyoo773 I couldn't find a case where we are doing modal-on-modal. Here are some potential changes. I figured the easiest thing to do would be to just create a PR on your feature branch.

https://github.com/yjyoo773/Synapse-React-Client/pull/44

nickgros (approved):

PR #1853, opened 2022-10-24
PORTALS-2157 - Rename QueryFilter component to FacetFilterControls and small bugfix

nickgros:
- Rename QueryFilter component to FacetFilterControls to avoid confusion with the [QueryFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/QueryFilter.html) object. The QueryFilter object is not used by the FacetFilterControls, since QueryFilters are not used with facets.
- Fix case where a facet chip would be shown when the corresponding facet control could not be shown

### DIFF
--- a/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
+++ b/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
@@ -23,7 +23,7 @@ import { FacetChip } from './FacetChip'
 import { RangeFacetFilter } from './RangeFacetFilter'
 
 export type FacetFilterControlsProps = {
-  facetFiltersToShow?: string[]
+  facetsToFilter?: string[]
 }
 
 const convertFacetToFacetColumnValuesRequest = (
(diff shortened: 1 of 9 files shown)


src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -137,19 +137,23 @@ export const applyChangesToRangeColumn = (
...
+  const { facetFiltersToShow } = props
   const { data, isLoadingNewBundle, getLastQueryRequest, executeQueryRequest } =
     useQueryContext()
 
-  const facets = data?.facets
+  const facets = data?.facets?.filter(
+    facet =>
+      // Don't show facets where there are no values
+      !isSingleNotSetValue(facet),
+  )
```
nickgros, line 149: Here's the change. The filter was previously applied just on the control, so the chip would render but the control wouldn't be shown when the chip was clicked.

jay-hodgson (approved):

PR #1856, opened 2022-10-25
Portals 2157 all

nickgros:
- DetailsPage search params that add filters now always use `additionalFilters` instead of altering the SQL query
- Removed unused/buggy ability to link to "disaggregated" query when selecting a COUNT aggregation
- Remove `sql-parser` (no longer using after these changes)
- Remove `vite-plugin-externals` (only using to work around bugs in sql-parser)


nickgros:
Closed in favor of #1868 and others

PR #1858, opened 2022-10-25
PORTALS-2370

jay-hodgson:


jay-hodgson:
Screen Shot 2022-10-25 at 10 29 52 AM

jay-hodgson:
Extend existing useGetInfoFromIds to support Evaluation IDs (in addition to user IDs and entity IDs).
Hook did not read cached results, so removed this code complexity.  Better solution is to utilize react-query, but this code change is to fix the immediate bug.

src/lib/utils/hooks/useGetInfoFromIds.ts
```diff
@@ -91,46 +125,13 @@ export default function useGetInfoFromIds<
...
-      sessionStorage.setItem(
-        storageKey(type),
-        JSON.stringify(dataInStorageAsObjectArr),
-      )
-    } catch (e) {
-      sessionStorage.setItem(storageKey(type), JSON.stringify(data))
-    }
-  }
-
-  // Alina TODO: check if the items are already in Local Storage before making server call.
```
jay-hodgson, line 132: Note Alina's comment.  She was right, local storage cache is not used, so there's no point in saving the results to web storage

jay-hodgson: There is a point!  TotalQueryResults uses the cache to figure out the display name to be rendered.

src/lib/containers/TotalQueryResults.tsx
```diff
@@ -142,24 +153,31 @@ const TotalQueryResults: FunctionComponent<TotalQueryResultsProps> = ({
         facet.facetValues.forEach(facetValue => {
           if (facetValue.isSelected) {
             let displayValue = facetValue.value
-            if (
```
jay-hodgson, line 145: Changed to a switch statement, and added EVALUATIONID

nickgros (approved):
All of my issues with this code are not new...so I'll approve

src/lib/containers/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -50,6 +55,11 @@ function valueToLabel(
     displayValue = eh ? eh.name : `unknown (${value})`
   }
 
+  const evaluation = evaluations.find(evaluation => evaluation.id === value)
+  if (evaluation?.name) {
+    displayValue = evaluation.name
+  }
+
   return `${displayValue}`
```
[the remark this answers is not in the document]
jay-hodgson: 💯

PR #1863, opened 2022-10-26
PORTALS-2381: basic commands for markdown editor

yjyoo773:


src/lib/containers/IconSvg.tsx
```diff
@@ -455,6 +462,16 @@ const getIcon = (options: IconSvgOptions) => {
...
+    case 'strikethrough':
+      return <StrikethroughS style={muiSvgStyle} />
+    case 'latex':
+      return <span>TeX</span>
+    case 'image':
+      return <Image style={muiSvgStyle} />
+    case 'superscript':
+      return <span>X&sup2;</span>
+    case 'subscript':
+      return <span>X&#8322;</span>
```
yjyoo773, line 474: Could not find appropriate icon for LaTeX, subscript, and superscript. There seems to be icons for subscript and superscript in MUI V5 which we can later update?

yjyoo773:
Screen Shot 2022-10-26 at 10 29 11 AM

src/lib/containers/IconSvg.tsx
nickgros: What you have works!

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
   const [currentTab, setCurrentTab] = useState<MarkdownEditorTabs>(
     MarkdownEditorTabs.WRITE,
   )
-  const [text, setText] = useState<string>()
+  const [text, setText] = useState<string>('')
+  const [selectionStart, setSelectionStart] = useState<number>(0)
+  const commandList = [
```
nickgros, line 22: Pull this constant out of the component, it doesn't need to be re-created every render

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+    'italic',
+    'strikethrough',
+    'code',
+    'latex',
+    'subscript',
+    'superscript',
+    'link',
+    'image',
+  ]
+
```
nickgros, line 34:
```suggestion
  ] as const
  
  type CommandListType = typeof commandList

```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+  const textAreaRef = useRef<HTMLTextAreaElement>(null)
+
+  useEffect(() => {
+    const input = textAreaRef.current
+    if (input) {
+      input.setSelectionRange(selectionStart, selectionStart)
+    }
+  }, [textAreaRef, selectionStart])
+
+  const handleCommands = (command: string) => {
```
nickgros, line 44: Stronger typing
```suggestion
  const handleCommands = (command: CommandListType) => {
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+    if (textVal) {
+      const start = textVal.selectionStart
+      const end = textVal.selectionEnd
+      const selected = text.substring(start, end)
+      const first = text.substring(0, start)
+      const later = text.substring(end, text.length)
+
+      let openSyntax = ''
+      let closeSyntax = ''
+      if (command === 'bold') {
```
nickgros, line 55: Prefer switch/case for a list of many controlled values

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+        setText(newText)
+        textAreaRef.current.focus()
+        // adds one due to new line
+        setSelectionStart(start + openSyntax.length + 1)
+        return
+      }
+
+      const newText = `${first}${openSyntax}${selected}${closeSyntax}${later}`
+      textAreaRef.current.focus()
+      setSelectionStart(start + openSyntax.length)
```
nickgros, line 95: Can you just do this and get rid of the useEffect and useState? 
```suggestion
      const finalCursorPosition = start + openSyntax.length
      textAreaRef.current.setSelectionRange(finalCursorPosition, finalCursorPosition)
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -42,18 +120,13 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
-            </button>
-            <button>
-              <IconSvg options={{ icon: 'italic' }} />
-            </button>
-            <button>
-              <IconSvg options={{ icon: 'code' }} />
-            </button>
+            {commandList.map(type => {
+              return (
+                <button key={type} onClick={() => handleCommands(type)}>
```
nickgros, line 125: If you run into type issues where it conflicts with the Icon type, you could use a type cast

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +17,85 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+      const start = textVal.selectionStart
+      const end = textVal.selectionEnd
+      const selected = text.substring(start, end)
+      const first = text.substring(0, start)
+      const later = text.substring(end, text.length)
+
+      let openSyntax = ''
+      let closeSyntax = ''
+      if (command === 'bold') {
+        openSyntax = closeSyntax = '**'
```
nickgros, line 56: I think it might be cleaner to just create a new file with a data object like
```ts
type MarkdownEditorCommandData = { // or a better name if you can come up with one
	openSyntax: string
	closeSyntax: string | null
}

type MarkdownEditorCommands = Record<CommandListType, MarkdownEditorCommandData>

export const MARKDOWN_COMMANDS_DATA: MarkdownEditorCommands: {
	bold: {
		openSyntax: "**",
		closeSyntax: "**"
	},
	title: {
		openSyntax: "###",
		closeSyntax: null
	},
	// ...
}
```

Then just pull it in here and use it.

nickgros (changes requested):

src/lib/containers/markdown/MarkdownEditor.tsx
yjyoo773: This was my initial attempt. However when doing so the cursor would still go to the end of the text. Whereas the useEffect would put the cursor right next to the openSyntax.

nickgros: I see; the `setSelectionRange` call needs to happen after the `text` state updates, and `useEffect` ensures that that happens in the right order.

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +22,62 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
   const [currentTab, setCurrentTab] = useState<MarkdownEditorTabs>(
     MarkdownEditorTabs.WRITE,
   )
-  const [text, setText] = useState<string>()
+  const [text, setText] = useState<string>('')
+  const [selectionStart, setSelectionStart] = useState<number>(0)
+  const textAreaRef = useRef<HTMLTextAreaElement>(null)
+
+  useEffect(() => {
```
nickgros, line 29:
```suggestion
  /**
   * Updates the position of the cursor. This must occur within `useEffect` because we need to ensure that the
   * cursor position is updated after the state of `text` is updated.
   */
  useEffect(() => {
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +22,62 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+    }
+  }, [textAreaRef, selectionStart])
+
+  const handleCommands = (command: CommandListType) => {
+    const textVal = textAreaRef.current
+    if (textVal) {
+      const start = textVal.selectionStart
+      const end = textVal.selectionEnd
+      const selected = text.substring(start, end)
+      const first = text.substring(0, start)
```
nickgros, line 42:
```suggestion
      const textBeforeSelection = text.substring(0, start)
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +22,62 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+      const later = text.substring(end, text.length)
```
nickgros, line 43:
```suggestion
      const textAfterSelection = text.substring(end, text.length)
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,7 +22,62 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
+
+      switch (command) {
+        case 'code': {
+          const newText = `${first}
+          ${openSyntax}
+          ${selected}
+          ${closeSyntax}
+          ${later}`
+            .split('\n')
+            .map(s => s.trim())
```
nickgros, line 56: I don't think trimming whitespace is right, this would turn 

```
{
  "hello": "world"
}
```

into 

```
``
{
"hello": "world"
}
``
```

What case were you trying to solve here?

nickgros (changes requested):

nickgros (approved):

PR #1876, opened 2022-11-01
SWC-6180: Upgrade to MUI v5

nickgros:
- Upgrade to MUI v5
- Convert Tooltip, Typography custom styled components to default MUI configuration

Most of the changes were applied using a migration codemod provided by MUI


package.json
```diff
@@ -17,9 +17,10 @@
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@brainhubeu/react-carousel": "1.19.26",
-    "@material-ui/core": "^4.12.3",
-    "@material-ui/icons": "^4.11.2",
-    "@material-ui/lab": "^4.0.0-alpha.57",
+    "@emotion/react": "^11.10.5",
+    "@emotion/styled": "^11.10.5",
```
nickgros, line 21: Peer dependency of MUI, and also a library we can use to make our own styled components

package.json
```diff
@@ -17,9 +17,10 @@
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@brainhubeu/react-carousel": "1.19.26",
-    "@material-ui/core": "^4.12.3",
-    "@material-ui/icons": "^4.11.2",
-    "@material-ui/lab": "^4.0.0-alpha.57",
```
nickgros, line 22: Skeleton is the only thing in lab we're using, and it's in @mui/material in v5. So lab is gone

src/lib/utils/theme/useTheme.tsx
```diff
@@ -0,0 +1,27 @@
+import { createTheme, StyledEngineProvider, ThemeOptions } from '@mui/material'
+import { ThemeProvider as MuiThemeProvider } from '@mui/material/styles'
+import { merge } from 'lodash-es'
+import React, { useMemo } from 'react'
+import defaultMuiTheme from './DefaultTheme'
+
```
nickgros, line 6: I got rid of the `SynapseTheme`. To override the theme, just use the MUI ThemeOptions type. We'll merge it with a default set of ThemeOptions

src/lib/utils/theme/typography/Typography.ts
```diff
@@ -0,0 +1,105 @@
+import { TypographyOptions } from '@mui/material/styles/createTypography'
+import { Palette } from '@mui/material/styles/createPalette'
+
+export const typographyOptions: (
```
nickgros, line 4: Moved Typography styles here. They get added directly to the MUI theme and then we can just use MUI typography instead of our own component wrapper.

src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -0,0 +1,35 @@
...
+        arrow: ({ ownerState, theme }) => ({
+          color: theme.palette.common.black,
+        }),
+        tooltip: ({ ownerState, theme }) => ({
+          fontSize: '14px',
+          borderRadius: '2px',
+          backgroundColor: theme.palette.common.black,
+        }),
+      },
+    },
```
nickgros, line 25: Instead of applying in a wrapper component over tooltip, we set these in the theme and then they are always applied to the Tooltip component. Wrapper component removed.

src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -0,0 +1,35 @@
...
+    MuiTypography: {
+      defaultProps: {
+        fontStyle: 'normal',
+        letterSpacing: '0px',
+      },
```
nickgros, line 30: props shared by all Typography variants.

src/lib/utils/theme/ThemeTypes.ts
```diff
@@ -0,0 +1,60 @@
...
+  | 'dataFieldKey'
+
+// We create this type because we have to use interfaces in the module augmentations, which don't allow mapped types.
+type RecordWithCustomVariantKeys<Value> = {
+  [key in CustomTypographyVariants]: Value
+}
+
+// To add custom Typography variants to the MUI theme, we have to extend the MUI theme type.
+// See https://mui.com/material-ui/customization/theming/#theme-configuration-variables CTRL+F TypeScript
+declare module '@mui/material/styles' {
```
nickgros, line 31: This "module augmentation" extends the MUI types and allows us to have static typing with our own extensions to the MUI theme (e.g. Typography variants, custom color definitions)

src/lib/containers/HasAccessV2.tsx
```diff
@@ -40,53 +40,50 @@ export enum FileHandleDownloadTypeEnum {
...
     icon: 'accessOpen',
-    color: theme => theme.colors.success,
+    color: theme => theme.palette.success.main,
     tooltipText: '',
   },
 }
 
 function AccessIcon(props: { downloadType: FileHandleDownloadTypeEnum }) {
   const { downloadType } = props
+  const theme = useTheme()
```
nickgros, line 76: Use the @mui/material theme, not the SynapseTheme (which was removed)

src/lib/containers/table/table-top/DownloadOptions.tsx
```diff
@@ -67,12 +67,17 @@ export const DownloadOptions: React.FunctionComponent<
           </Dropdown.Item>
           {isFileViewOrDataset && (
             <Tooltip
-              title="A draft version of a dataset cannot be added to the Download Cart"
+              title={
+                disableDownload
+                  ? 'A draft version of a dataset cannot be added to the Download Cart'
+                  : null
+              }
```
nickgros, line 74: Tooltip isn't shown if title is `''`, `null`, or `undefined`

src/lib/containers/table/table-top/DownloadOptions.tsx
```diff
@@ -67,12 +67,17 @@ export const DownloadOptions: React.FunctionComponent<
...
               placement="left"
               enterNextDelay={300}
-              disableHoverListener={!disableDownload}
+              describeChild={true}
```
nickgros, line 77: I think in v5 they changed it so the child's `aria-label` attribute is set to the Tooltip's `title` prop, unless you set `describeChild` to be true (which is appropriate in this case, because it is _describing_ the child, not _labeling_ it). This change fixed broken tests.

src/lib/utils/hooks/useTheme.tsx
```diff
@@ -1,86 +0,0 @@
...
-      main: '#469285',
-    },
-    success: '#32a330',
-    info: '#017fa5',
-    warning: '#cc9f00',
-    error: '#c13415',
-  },
-}
-
-export type SynapseTheme = {
```
nickgros, line 25: Removed in favor of just using the MUI ThemeOptions

src/lib/utils/hooks/useTheme.tsx
```diff
@@ -1,86 +0,0 @@
...
-    warning: string
-    error: string
-  }
-}
-
-/**
- * This must be exported to use the context in class components.
- */
-export const ThemeContext =
-  React.createContext<SynapseTheme>(defaultSynapseTheme)
```
nickgros, line 44: Removed, just use the MUI theme context

src/lib/utils/hooks/useTheme.tsx
```diff
@@ -1,86 +0,0 @@
-import { createTheme, MuiThemeProvider } from '@material-ui/core'
-import { merge } from 'lodash-es'
-import React, { useContext, useMemo } from 'react'
-import DeepPartial from '../types/DeepPartial'
-
```
nickgros, line 5: File was moved and had many changes, so git thinks it was deleted.

esbuild.config.mjs
```diff
@@ -54,7 +55,7 @@ const esBuildOptions = {
     GlobalsPlugin(globals),
   ],
   external: [
-    'react',
+    '^react$', // Use regex ^$ because we do want to bundle 'react/jsx-transform'
```
nickgros, line 58: MUI 5 and emotion ship code that uses the automatic JSX transform, so their source code -> distribution looks kind of like this:
```jsx
	<Component {...props}>
		{children}
	</Component>
```

```js
	import jsx from 'react/jsx-transform'
	jsx(Component, props, children) // instead of React.createElement
```

We have `react` set as an external package, and the React UMD bundle doesn't include the JSX transform module, so this would break. Using regex allows us to keep `'react'` imports external but still bundle `'react/jsx-transform'`, fixing this case.

package.json
```diff
@@ -115,7 +116,7 @@
     "@googlemaps/jest-mocks": "^2.7.5",
     "@juggle/resize-observer": "^3.3.1",
     "@rollup/plugin-inject": "^4.0.4",
-    "@sage-bionetworks/esbuild-plugin-svgr": "^0.0.4",
+    "esbuild-plugin-svgr": "^1.0.1",
```
nickgros, line 119: I had created a PR from our fork that got merged some time ago, so let's go back to the upstream version

package.json
```diff
@@ -163,10 +164,10 @@
     "babel-loader": "^8.2.5",
     "buffer": "^6.0.3",
     "core-js": "^3.1.4",
-    "esbuild": "^0.14.2",
+    "esbuild": "^0.15.12",
     "esbuild-plugin-globals": "^0.1.1",
     "esbuild-plugin-node-polyfills": "^1.0.2",
-    "esbuild-sass-plugin": "^1.8.0",
+    "esbuild-sass-plugin": "^2.4.0",
     "eslint": "^8.18.0",
```
nickgros, line 171: Upgrading esbuild and dependencies in general

src/lib/utils/theme/typography/Typography.ts
```diff
@@ -0,0 +1,128 @@
...
+    marginTop: '24px',
+    fontStyle: 'italic',
+  },
+  body2: {
+    ...defaultStyles,
+    fontSize: '12px',
+    fontWeight: 400,
+    lineHeight: '24px',
+    marginTop: '24px',
+  },
```
nickgros, line 55: `body2` is used internally by MUI for some components like Pagination. We really want it to just be a smaller version of body1.

nickgros, line 55: `body2` is used internally by MUI for some components like Pagination. We really want it to just be a smaller version of body1.

src/lib/utils/theme/typography/Typography.ts
```diff
@@ -0,0 +1,128 @@
...
+  body1Italic: {
+    ...defaultStyles,
+    fontSize: '16px',
+    fontWeight: 400,
+    lineHeight: '24px',
+    marginTop: '24px',
+    fontStyle: 'italic',
+  },
```
nickgros, line 48: `body1Italic` is new. Has the same properties as old `body2`.

jay-hodgson (approved):
Assuming icons load properly and typography has expected rendering, these changes look great!

PR #1878, opened 2022-11-02
PORTALS-2348: create and edit discussion thread

yjyoo773:


src/lib/utils/hooks/SynapseAPI/forum/useForum.ts
```diff
@@ -54,18 +61,128 @@ export function useGetForumInfinite(
   )
 }
 
-export function useGetThread(
+export function useGetThread(threadId: string) {
+  const { data: threadData, isLoading: isLoadingBundle } =
+    useGetThreadBundle(threadId)
+  const { data: threadBody, isLoading: isLoadingBody } = useGetThreadBody(
+    threadData?.messageKey ?? '',
```
yjyoo773, line 68: The parameter for this hook requires the response from the above hook. However, when first instantiated it runs with an empty parameter throwing a 400 error. The expected response is eventually returned but is there any other suggestions?

nickgros: Does this fix it?

```suggestion
  const { data: threadBody, isLoading: isLoadingBody } = useGetThreadBody(
    threadData?.messageKey ?? '',
    { enabled: !!(threadData?.messageKey) }
```

src/lib/utils/hooks/SynapseAPI/forum/useForum.ts
```diff
@@ -54,18 +61,128 @@ export function useGetForumInfinite(
...
+  options?: UseQueryOptions<string, SynapseClientError>,
+) {
+  const { accessToken } = useSynapseContext()
+  const messageUrl = SynapseClient.getThreadMessageUrl(messageKey, accessToken)
+  return useQuery<string, SynapseClientError>(
+    ['thread', messageKey, accessToken],
+    async () => getMessage((await messageUrl).messageUrl),
+    options,
+  )
+}
```
nickgros, line 97: Consistently use the thread ID in the second field.

You could also define a type and use an object instead of an array to keep this organized. If you want to do that, take a look at queryKeys.ts which is used for entities.

```suggestion
export function useGetThreadBody(
  threadBundle: DiscussionThreadBundle,
  options?: UseQueryOptions<string, SynapseClientError>,
) {
  const { accessToken } = useSynapseContext()
  const messageUrl = SynapseClient.getThreadMessageUrl(messageKey, accessToken)
  return useQuery<string, SynapseClientError>(
    ['thread', threadBundle.id, messageKey, accessToken],
    async () => getMessage((await messageUrl).messageUrl),
    options,
  )
}
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,17 +18,28 @@ export enum MarkdownEditorTabs {
 
 export type MarkdownEditorProps = {
   placeholder?: string
+  forumId?: string
+  editTitle?: string
```
nickgros, line 22: Let's keep the markdown editor generic and not forum-specific.
```suggestion
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -17,17 +18,28 @@ export enum MarkdownEditorTabs {
...
+  editText?: string
+  onSave: (text: string, title?: string) => void
+  onCancel: () => void
```
nickgros, line 25: Let's also make this fully controlled component:
```suggestion
  text?: string
  setText?: (text: string) => void
```

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -107,6 +124,15 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
     <div className="bootstrap-4-backport MarkdownEditor">
+      {forumId && (
+        <input
+          placeholder="Title"
+          type="text"
+          style={{ width: '100%', padding: '4px' }}
+          value={title}
+          onChange={e => setTitle(e.target.value)}
+        />
+      )}
```
nickgros, line 135: This should be in a `ForumThreadEditor` component, and rendered next to a `MarkdownEditor`, instead of inside of it, since the title field is forum-thread specific.

src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -161,7 +187,16 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
...
           'Nothing to preview'
         )}
+        <div style={{ float: 'right' }}>
+          <Button onClick={onCancel} variant="light">
+            Cancel
+          </Button>
+          <Button onClick={() => handleSave()} variant="primary">
+            Save
+          </Button>
+        </div>
```
nickgros, line 197: Same here--these will be in a `ForumThreadEditor component`

src/lib/containers/discussion_forum/ForumTable.tsx
```diff
@@ -205,13 +224,34 @@ export const ForumTable: React.FC<ForumTableProps> = ({
...
+      >
+        <Modal.Header>
+          <Modal.Title>New Thread</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <MarkdownEditor
+            onSave={onSave}
+            forumId={forumId}
+            onCancel={() => setThreadModal(false)}
+          />
```
nickgros, line 251: This should use a `ForumThreadEditor` component. See comments on MarkdownEditor.tsx for more details

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -139,12 +149,37 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+          <Modal.Title>New Thread</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <MarkdownEditor
+            onSave={onSave}
+            forumId={forumId}
+            onCancel={() => setThreadModal(false)}
+            editTitle={threadData?.title}
+            editText={threadBody}
+          />
```
nickgros, line 179: This should use a ForumThreadEditor component. See comments on MarkdownEditor.tsx for more details

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -57,23 +63,20 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+  const onSave = (text: string, title?: string) => {
+    updateTitle({
+      title: title!,
+      threadId: threadId,
+    })
+    updateMessage({
+      messageMarkdown: text,
+      threadId: threadId,
+    })
+  }
```
nickgros, line 75: This logic should be moved to a ForumThreadEditor component. See other comments for details.

src/lib/containers/discussion_forum/ForumTable.tsx
```diff
@@ -55,6 +62,15 @@ export const ForumTable: React.FC<ForumTableProps> = ({
...
   )
 
+  const onSave = (text: string, title?: string) => {
+    const request: CreateDiscussionThread = {
+      forumId: forumId,
+      title: title!,
+      messageMarkdown: text,
+    }
+    postThread(request)
+  }
```
nickgros, line 72: This logic should be moved to a ForumThreadEditor component. See other comments for details.

nickgros (changes requested):

src/lib/containers/discussion_forum/ForumThreadEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+  useCreateThread,
+  useUpdateThreadTitle,
+  useUpdateThreadMessage,
+} from '../../utils/hooks/SynapseAPI/forum/useForum'
+import { CreateDiscussionThread } from '../../utils/synapseTypes/DiscussionBundle'
+import { MarkdownEditor } from '../markdown/MarkdownEditor'
+
+export type ForumThreadEditorProps = {
+  editTitle?: string
+  editText?: string
```
nickgros, line 13: Do you think these are better prop names?
```suggestion
  initialTitle?: string
  initialMessage?: string
```

src/lib/containers/discussion_forum/ForumThreadEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+  id: string
+  onCancel: () => void
```
nickgros, line 15: `onClose` is a better name since you also call it after saving

```suggestion
  onClose: () => void
```

nickgros (approved):

PR #1879, opened 2022-11-03
PORTALS-2349: reply to existing thread

yjyoo773:


src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -124,15 +126,41 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+          placeholder="Write a reply..."
+          onClick={() => setShowReplyEditor1(true)}
+        />
+      ) : (
+        <ForumThreadEditor
+          id={threadId}
+          isReply={true}
+          onClose={() => setShowReplyEditor1(false)}
+        />
+      )}
```
nickgros, line 141: Do you need this here and `showReplyEditor1`? It seems like the other one will always work if you remove `replies.length > 0 &&`

src/lib/containers/discussion_forum/ForumThreadEditor.tsx
```diff
@@ -13,58 +15,103 @@ export type ForumThreadEditorProps = {
...
+  const { mutate: updateMessage, isLoading: isLoadingMessage } =
+    useUpdateThreadMessage({
+      onSuccess: () => onClose(),
+    })
+  const { mutate: postThread, isLoading: isLoadingThread } = useCreateThread({
+    onSuccess: () => onClose(),
+  })
+  const { mutate: postReply, isLoading: isLoadingReply } = usePostReply({
+    onSuccess: () => onClose(),
+  })
```
nickgros, line 39: Being super picky but I think this is slightly easier to understand

```suggestion
  const { mutate: createThread, isLoading: isLoadingThread } = useCreateThread({
    onSuccess: () => onClose(),
  })
  const { mutate: createReply, isLoading: isLoadingReply } = usePostReply({
    onSuccess: () => onClose(),
  })
```

src/lib/utils/hooks/SynapseAPI/forum/useForum.ts
```diff
@@ -203,7 +205,7 @@ export function useGetRepliesInfinite(
     PaginatedResults<DiscussionReplyBundle>,
     SynapseClientError
   >(
-    ['forumthread', 'infinite', threadId, limit, filter, sort, ascending],
+    ['thread', 'infinite', threadId, limit, filter, sort, ascending],
```
nickgros, line 208: If you use the thread ID second here...
```suggestion
    ['thread', threadId, 'infinite', limit, filter, sort, ascending],
```

src/lib/utils/hooks/SynapseAPI/forum/useForum.ts
```diff
@@ -224,3 +226,61 @@ export function useGetRepliesInfinite(
...
+    DiscussionReplyBundle,
+    SynapseClientError,
+    CreateDiscussionReply
+  >(
+    (request: CreateDiscussionReply) =>
+      SynapseClient.postReply(request, accessToken),
+    {
+      ...options,
+      onSuccess: async (newReply, variables, ctx) => {
+        await queryClient.invalidateQueries(['thread'])
```
nickgros, line 250: ...then you can invalidate _just_ the thread that was updated here
```suggestion
        await queryClient.invalidateQueries(['thread', request.threadId])
```

src/lib/utils/hooks/SynapseAPI/forum/useForum.ts
```diff
@@ -224,3 +226,61 @@ export function useGetRepliesInfinite(
...
+    DiscussionReplyBundle,
+    SynapseClientError,
+    UpdateDiscussionReply
+  >(
+    (request: UpdateDiscussionReply) =>
+      SynapseClient.putReply(request, accessToken),
+    {
+      ...options,
+      onSuccess: async (newReply, variables, ctx) => {
+        await queryClient.invalidateQueries(['thread'])
```
nickgros, line 279:
```suggestion
        await queryClient.invalidateQueries(['thread', request.threadId])
```

nickgros (changes requested):

src/lib/containers/discussion_forum/DiscussionThread.tsx
yjyoo773: I was imitating as much as Synapse was doing. It originally has a textarea which renders the editor onClick which is what the `showReplyEditor` is doing. There was also another separate textarea at the bottom when there were existing replies in the thread.

nickgros (approved):

PR #1881, opened 2022-11-03
SWC-6202

jay-hodgson:
Initially attempted to force update storybook to the latest release (6.5.13, released 11 days ago).  But since that didn't update the target sub-dependency (trim), I went ahead and added a resolution and manually tested Storybook.


package.json
```diff
@@ -198,6 +198,7 @@
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
     "timers-browserify": "^2.0.12",
+    "trim": "^0.0.3",
```
nickgros, line 201: I don't think we need to specify this in `devDependencies` is we aren't directly using it. Keep in `resolutions`

nickgros (changes requested):
Can you remove `trim` from devDependencies?

nickgros (approved):
Pending CI

PR #1880, opened 2022-11-03
SWC-6180 (followup): IconSvg improvements

nickgros:
- (BREAKING) Change IconSvg signature to use sx (so you can change the colors of the icon using the theme) and not nest everything in an object
- Modify usage of SvgIcon to utilize theming
- Remove ContentCopy and Login (MUI Icons supplies these)

PR #1883, opened 2022-11-04
PORTALS-2353: delete a discussion thread

yjyoo773:


src/lib/containers/IconSvg.tsx
```diff
@@ -74,6 +74,9 @@ import {
   StrikethroughS,
   Image,
   EditTwoTone,
+  Superscript,
+  Subscript,
+  AlternateEmail,
```
yjyoo773, line 79: Update icons based on MUI V5

jay-hodgson: Conflicts with https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1880 ?

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -35,15 +37,27 @@ export function DiscussionThread(props: DiscussionThreadProps) {
   const [showThreadModal, setShowThreadModal] = useState(false)
   const [showReplyEditor1, setShowReplyEditor1] = useState(false)
   const [showReplyEditor2, setShowReplyEditor2] = useState(false)
+  const [showDeleteModal, setShowDeleteModal] = useState(false)
 
   const { threadData, threadBody } = useGetThread(threadId)
   const { data: currentUserProfile } = useGetCurrentUserProfile()
```
jay-hodgson, line 43: You should also be thinking about the anonymous user experience.  User might be able to READ the forum, but no user actions should be available (following, pinning, replying, ...)

src/lib/containers/discussion_forum/DiscussionThread.tsx
```diff
@@ -35,21 +37,39 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+
+  console.log(currentUserProfile)
 
   function handleFollowBtn() {
-    try {
-      toggleSubscribed()
-    } catch (err: any) {
-      displayToast(err.reason as string, 'danger')
+    if (currentUserProfile?.userName == 'anonymous') {
+      alert('You will need to sign in for access to that resource.')
```
yjyoo773, line 66: Prevents anonymous user to use functionality such as subscribing or posting a comment. Would there be a better way than using `alert`?

jay-hodgson: There is the SRC_SIGN_IN_CLASS (see HasAccessV2 for complications in using this) if you want to attempt to automatically invoke the sign-in UI

jay-hodgson: I would refactor to make the message a reusable constant.
There may be an opportunity here to introduce an abstraction for all of preflight checks (that a user must pass before doing certain operations). This could start as a signed-in check, but could be expanded to check for certification status or a validated user profile.  Just a thought.

jay-hodgson: I would not use the native browser alert functionality, as it is not aligned in style.  I think using displayToast is preferable.

jay-hodgson (approved):

PR #1890, opened 2022-11-08
SWC-6321: Increase react-query stale time defaults

nickgros:
- react-query default stale time increased from 1m to 5m
- Table query stale time set to 30m
- default cacheTime increased from 5m to 30m.

### DIFF
--- a/src/lib/utils/SynapseContext.tsx
+++ b/src/lib/utils/SynapseContext.tsx
@@ -11,7 +11,7 @@ import { ThemeOptions } from '@mui/material'
 export const defaultQueryClientConfig: QueryClientConfig = {
   defaultOptions: {
     queries: {
-      staleTime: 60 * 1000, // 60s
+      staleTime: 1000 * 60 * 5, // 5 min
       retry: false, // SynapseClient knows which queries to retry
     },
   },
--- a/src/lib/utils/SynapseContext.tsx
+++ b/src/lib/utils/SynapseContext.tsx
@@ -12,6 +12,7 @@ export const defaultQueryClientConfig: QueryClientConfig = {
   defaultOptions: {
     queries: {
       staleTime: 1000 * 60 * 5, // 5 min
+      cacheTime: 1000 * 60 * 30, // 30 min
       retry: false, // SynapseClient knows which queries to retry
     },
   },
--- a/src/lib/utils/SynapseContext.tsx
+++ b/src/lib/utils/SynapseContext.tsx
@@ -14,6 +14,7 @@ export const defaultQueryClientConfig: QueryClientConfig = {
       staleTime: 1000 * 60 * 5, // 5 min
       cacheTime: 1000 * 60 * 30, // 30 min
       retry: false, // SynapseClient knows which queries to retry
+      refetchOnWindowFocus: true,
     },
   },
 }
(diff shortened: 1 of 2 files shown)


src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
```diff
@@ -23,6 +23,7 @@ import {
 import { entityQueryKeys } from './queryKeys'
 
 const sharedQueryDefaults = {
+  staleTime: 1000 * 60 * 30, // 30 minutes
   refetchOnWindowFocus: false,
```
nickgros, line 27:
```suggestion
```

src/lib/utils/SynapseContext.tsx
```diff
@@ -11,7 +11,8 @@ import { ThemeOptions } from '@mui/material'
 export const defaultQueryClientConfig: QueryClientConfig = {
   defaultOptions: {
     queries: {
-      staleTime: 60 * 1000, // 60s
+      staleTime: 1000 * 60 * 5, // 5 min
+      cacheTime: 1000 * 60 * 30, // 30 min
       retry: false, // SynapseClient knows which queries to retry
```
nickgros, line 16:
```suggestion
      retry: false, // SynapseClient knows which queries to retry
      refetchOnWindowFocus: true,
```

src/lib/utils/SynapseContext.tsx
```diff
@@ -11,8 +11,10 @@ import { ThemeOptions } from '@mui/material'
 export const defaultQueryClientConfig: QueryClientConfig = {
   defaultOptions: {
     queries: {
-      staleTime: 60 * 1000, // 60s
+      staleTime: 1000 * 60 * 5, // 5 min
+      cacheTime: 1000 * 60 * 30, // 30 min
```
nickgros, line 15: So data will be considered "fresh" (won't be auto-refetched) for 5 minutes, and stale data could appear before being immediately refetched for 30 minutes.

src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
```diff
@@ -23,7 +23,7 @@ import {
 import { entityQueryKeys } from './queryKeys'
 
 const sharedQueryDefaults = {
-  refetchOnWindowFocus: false,
+  staleTime: 1000 * 60 * 30, // 30 minutes
```
nickgros, line 26: For table queries, max out the stale time to match the cache time, i.e. don't auto-refetch data for 30 minutes

src/lib/utils/SynapseContext.tsx
```diff
@@ -11,8 +11,10 @@ import { ThemeOptions } from '@mui/material'
...
       retry: false, // SynapseClient knows which queries to retry
+      refetchOnWindowFocus: true,
```
nickgros, line 17:
```suggestion
      refetchOnWindowFocus: false,
```

src/lib/utils/hooks/SynapseAPI/entity/useGetQueryResultBundle.ts
jay-hodgson: Does this mean that after a Table is updated in SWC the user will not see the change for 30 minutes?

nickgros: They will see the change if they refresh the page (unless we move the cache to localStorage or sessionStorage), or if we invalidate the cache. The SWC EntityPresenter currently clears data for all entities from the cache when we fire the EntityUpdatedEvent.

PR #1891, opened 2022-11-08
PORTALS-2407: proof of concept

jay-hodgson:


src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -9,6 +9,18 @@ const defaultMuiTheme: ThemeOptions = {
   palette: palette,
   components: {
     MuiLink: linkTheme,
+    MuiButton: {
+      styleOverrides: {
+        root: {
+          disableRipple: true,
```
nickgros, line 15: I think `disableRipple` should be in `defaultProps` rather than style overrides. Prefer theming in `defaultProps` when possible because it's easier to override in special cases.

nickgros (approved):
Looks good! One small suggestion

PR #1904, opened 2022-11-17
SWC-6323: Remove marginTop from Typography styles

nickgros:
- Remove marginTop from all Typography styles because it was breaking styling for builtin MUI components (namely the dropdown menu)
- Re-apply margins to select components with SCSS mixin
- Fix typo in DownloadCartPage
- Add stories for a few components that I needed to spot-check: BlockingLoader, MeetAccessRequirementCard, RequestDownloadCard


src/lib/containers/download_list_v2/DownloadCartPage.tsx
```diff
@@ -160,7 +160,7 @@ export const DownloadCartPage: React.FunctionComponent<
                         Packages)
                       </Typography>
                       <HelpPopover
-                        markdownText="This will allow you to create a .zip file that contains eligible files. Files greater that 100 MB, external links, or files which are not stored on Synapse native storage are ineligible. In most cases, ineligible files can be downloaded individually. External links will require navigation to an external site, which may require a seperate login process."
+                        markdownText="This will allow you to create a .zip file that contains eligible files. Files greater that 100 MB, external links, or files which are not stored on Synapse native storage are ineligible. In most cases, ineligible files can be downloaded individually. External links will require navigation to an external site, which may require a separate login process."
```
nickgros, line 163: "seperate" -> "separate"

src/lib/style/abstracts/_mixins.scss
```diff
@@ -194,3 +194,35 @@ $link-underline-color-default: adjust-color(
     @include disabled($ignoreLink: true);
   }
 }
+
+// Always top margins to the MuiTypography variants causes problems with some built-in MUI components
+// Use this mixin within a selector to apply the margins
+@mixin MuiTypographyTopMargins() {
```
nickgros, line 200: I am not sure if we should use this mixin when developing new components, but it seemed to be the easiest way to apply the margin change to the handful of components that relied on them.

src/lib/utils/theme/typography/Typography.ts
```diff
@@ -15,43 +15,37 @@ export const typographyOptions: (
...
     fontWeight: 700,
     lineHeight: '20px',
-    marginTop: '20px',
   },
   body1: {
     ...defaultStyles,
     fontSize: '16px',
     fontWeight: 400,
     lineHeight: '24px',
-    marginTop: '24px',
```
nickgros, line 39: In particular, having this top margin on body1 caused each dropdown menu item to also get the margin, which looked weird. I suspect this have similar effects on other components, so it seems like this is the wrong place to apply the margin.

jay-hodgson (approved):

PR #1905, opened 2022-11-17
SWC-6323: Add generic DropdownMenu component

nickgros:
https://user-images.githubusercontent.com/17580037/202484591-38df024a-fde5-495a-a8f8-38a47f4f4dc8.mov

Thanks to MUI it also has keyboard navigation


yjyoo773 (approved):
This is great. Nice work!

PR #1908, opened 2022-11-18
PORTALS-2387: replace momentjs

yjyoo773:


src/lib/containers/entity/annotations/CustomDateTimeWidget.tsx
```diff
@@ -14,13 +14,9 @@ export const CustomDateTimeWidget: Widget = ({
   return (
     <CalendarWithIconFormGroup
       disabled={disabled}
-      value={value ? moment(value) : ''}
-      setterCallback={(newValue: string | Moment) => {
-        if (newValue == null || typeof newValue === 'string') {
-          onChange(newValue)
-        } else {
-          onChange(newValue.toISOString())
```
yjyoo773, line 22: Removed due to throwing error when typing a date

src/lib/containers/row_renderers/ObservationCard.tsx
```diff
@@ -46,7 +45,10 @@ export const ObservationCard: React.FunctionComponent<ObservationCardProps> = ({
       {time && (
         <div className="ObservationCard__time">
           <IconSvg icon="time" />
-          <span>{getFriendlyPreciseDuration(time, timeUnits)}</span>
+          <span>
+            {`${time} ${timeUnits}`}
+            {Number(time) > 1 ? 's' : ''}
+          </span>
```
yjyoo773, line 51: Removing `getFriendltyPreciseDuration` and replacing with above being this is the only place used and not compatible with dayjs

src/lib/containers/evaluation_queues/CalendarWithIconFormGroup.tsx
```diff
@@ -1,54 +1,46 @@
...
+            disabled={disabled}
+            onChange={setterCallback}
+            renderInput={params => (
+              <TextField
+                disabled={disabled}
+                id={id}
+                className="form-control calendar-date-time-input rounded-right"
+                variant="outlined"
+                {...params}
+              />
```
yjyoo773, line 39: MUI's DateTimePicker does not have a timezone functionality hence was omitted. Another possibility would be printing the result and timezone somewhere else than the input?

src/lib/containers/entity/annotations/CustomDateTimeWidget.tsx
```diff
@@ -14,13 +14,9 @@ export const CustomDateTimeWidget: Widget = ({
...
-        }
+      value={value ? dayjs(value) : ''}
+      setterCallback={(newValue: string | Dayjs | null) => {
+        onChange(newValue)
```
yjyoo773, line 19: Removed `onChange(newValue.toISOString()` as it was throwing errors when typing in dates with keyboard.

src/lib/containers/IDUReportSubmissionInfo.tsx
```diff
@@ -26,7 +26,7 @@ export const SubmissionInfoCard: React.FunctionComponent<
           <strong>
             {' '}
             Intended Data Use Statement &#40;accepted on{' '}
-            {formatDate(moment(info.modifiedOn), 'M/D/YYYY')}&#41;:
+            {formatDate(dayjs(info.modifiedOn, 'M/D/YYYY'))}&#41;:
```
nickgros, line 29: Wouldn't the default pattern in `formatDate` end up overriding the one specified here?
```suggestion
            {formatDate(dayjs(info.modifiedOn), 'M/D/YYYY')}&#41;:
```

src/lib/style/components/evaluation_queue/_calendar_with_icon_input_group.scss
```diff
@@ -21,4 +21,13 @@
...
   }
+  .css-12wm9o4-MuiInputBase-root-MuiOutlinedInput-root {
+    margin-top: 0;
+    padding: 8px;
+    border: 1px #dcdcdc solid;
+    border-radius: 2px;
+    .css-ihdtdm {
+      width: 0;
+    }
+  }
```
nickgros, line 32: These CSS selectors are auto-generated and will change, so this won't be applied.

src/lib/containers/row_renderers/ObservationCard.tsx
nickgros: How is it not compatible with dayjs? It seems like the behavior here will change, but I can't tell for sure because the tests were removed. Please keep the function and tests. If the function can't be trivially re-implemented or the expected test values need to change, then let's work through that.

src/lib/containers/evaluation_queues/CalendarWithIconFormGroup.tsx
nickgros: Maybe append the current user timezone after the input? This is important so we can't leave it out.

Quick mockup

image

src/lib/containers/entity/annotations/CustomDateTimeWidget.tsx
nickgros: What happens if it's a Dayjs object? I think `onChange` is expecting a `string` or `null`, even if the type accepts `Dayjs`

nickgros (changes requested):

src/lib/containers/row_renderers/ObservationCard.tsx
yjyoo773: After replacing momentjs with dayjs for getFriendlyPreciseDuration, one of the tests fail as the following.
```
    expect(received).toBe(expected) // Object.is equality

    Expected: "1 year 1 month"
    Received: "1 year 25 days"

      17 |     expect(getFriendlyPreciseDuration('4', 'week')).toBe('28 days')
      18 |     expect(getFriendlyPreciseDuration('6', 'week')).toBe('1 month 12 days')
    > 19 |     expect(getFriendlyPreciseDuration('13', 'month')).toBe('1 year 1 month')
         |                                                       ^
      20 |   })
      21 | })
```

src/lib/containers/entity/annotations/CustomDateTimeWidget.tsx
yjyoo773: If you mean when using the DateTimePicker it works. As soon as the user types into the input an error is thrown `Invalid time value`.

src/lib/containers/evaluation_queues/CalendarWithIconFormGroup.tsx
yjyoo773: Screen Shot 2022-11-18 at 1 23 06 PM
Changed to look like this

src/lib/containers/evaluation_queues/CalendarWithIconFormGroup.tsx
```diff
@@ -26,19 +33,22 @@ export const CalendarWithIconFormGroup: React.FunctionComponent<
...
                 id={id}
-                className="form-control calendar-date-time-input rounded-right"
                 variant="outlined"
                 {...params}
               />
             )}
           />
+          <span style={{ margin: 'auto 8px' }}>
+            {dayjs().tz(dayjs.tz.guess()).format('z')}
+          </span>
```
nickgros, line 51: Is it possible to change the DateTimePicker and this rendered text if the user has UTC mode on?

nickgros: Actually...was the old picker always in local time? This is probably fine if so

src/lib/containers/entity/annotations/CustomDateTimeWidget.tsx
nickgros: We decided we'll open a follow-up ticket.

nickgros (approved):

PR #1909, opened 2022-11-18
PORTALS-2410: modified package.json resolutions to regenerate yarn.lo…

jay-hodgson:
…ck file that uses patched loader-utils in dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -236,7 +236,9 @@
     "js-yaml": "3.13.1",
     "@types/react": "18.0.17",
     "@types/webpack": "^5.28.0",
-    "trim": "^0.0.3"
+    "trim": "^0.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -236,7 +236,9 @@
     "js-yaml": "3.13.1",
     "@types/react": "18.0.17",
     "@types/webpack": "^5.28.0",
-    "trim": "^0.0.3"
+    "trim": "^0.0.3",
+    "webpack/**/loader-utils": "^1.4.2",
+    "storybook/**/loader-utils": "^2.0.4"
```
nickgros, line 241: Does it work with just the lockfile changes, and without the changes in package.json ?

nickgros (approved):

PR #1914, opened 2022-11-29
PORTALS-2430

nickgros:
- Fix case where a bad filter would cause us to mis-index a USERID column, and make a malformed request to the backend
- Add mock data and test to verify that users with/out fake data are not filtered
- Pass null fake userProfile ID instead of empty string, so we don't make further invalid calls to the backend

### DIFF
--- a/src/lib/containers/CardContainer.tsx
+++ b/src/lib/containers/CardContainer.tsx
@@ -111,8 +111,7 @@ export const CardContainer = (props: CardContainerProps) => {
       )
     }
     const listIds = data.queryResult!.queryResults.rows.map(
-      el =>
-        el.values.filter((id): id is string => id !== null)[userIdColumnIndex],
+      el => el.values[userIdColumnIndex],
     )
     cards = <UserCardList data={data} list={listIds} size={MEDIUM_USER_CARD} />
   } else {
--- a/src/lib/containers/UserCardList.tsx
+++ b/src/lib/containers/UserCardList.tsx
@@ -1,17 +1,18 @@
-import * as React from 'react'
+import React from 'react'
 import { getUserProfileWithProfilePicAttached } from '../utils/functions/getUserData'
 import { UserProfileList } from '../utils/SynapseClient'
 import { MEDIUM_USER_CARD } from '../utils/SynapseConstants'
 import {
+  ColumnType,
   QueryResultBundle,
   UserProfile,
-  ColumnType,
 } from '../utils/synapseTypes/'
 import UserCard, { UserCardSize } from './UserCard'
 import { without } from 'lodash-es'
 
 export type UserCardListProps = {
-  list: string[]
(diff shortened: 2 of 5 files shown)


src/lib/containers/CardContainer.tsx
```diff
@@ -111,8 +111,7 @@ export const CardContainer = (props: CardContainerProps) => {
       )
     }
     const listIds = data.queryResult!.queryResults.rows.map(
-      el =>
-        el.values.filter((id): id is string => id !== null)[userIdColumnIndex],
+      el => el.values[userIdColumnIndex],
```
nickgros, line 114: This bad filter caused the issue. Null data would be dropped and the index would be wrong! We'd then send this data over to Synapse assuming it's a list user IDs and it turns out to be a list of arbitrary strings.

We don't actually want to filter anything; nulls are OK and represent fake profiles.

src/lib/containers/UserCardList.tsx
```diff
@@ -53,8 +54,10 @@ export default class UserCardList extends React.Component<
     }
   }
 
-  update(list: string[]) {
-    getUserProfileWithProfilePicAttached(list.filter(el => el))
+  update(list: (string | null)[]) {
+    getUserProfileWithProfilePicAttached(
+      list.filter((el): el is string => !!el),
```
nickgros, line 59: For this call, filter the null values (they obviously won't have profiles we can fetch)

src/lib/containers/UserCardList.tsx
```diff
@@ -93,13 +96,13 @@ export default class UserCardList extends React.Component<
...
     )
-    return nullOwnerIdsRows.map<UserProfile>(el => {
+    return nullOwnerIdsRows.map<Omit<UserProfile, 'ownerId'>>(el => {
       const values = el.values
       return {
         firstName: values[firstNameIndex] ?? '',
         lastName: values[lastNameIndex] ?? '',
         company: values[institutionIndex] ?? undefined,
-        ownerId: '',
+        ownerId: null,
```
nickgros, line 105: If we make this the empty string, the UserCard component will try to fetch the profile, which emits all kinds of errors and warnings to the console. It should really be null.

src/lib/containers/UserCardList.tsx
```diff
@@ -150,7 +153,8 @@ export default class UserCardList extends React.Component<
                   disableLink={true}
                   hideEmail={true}
                   size={size}
-                  userProfile={fauxUserProfile}
+                  // TODO: Modify UserCard to accept a fake profile with no ownerId field, and remove this cast.
+                  userProfile={fauxUserProfile as UserProfile}
```
nickgros, line 157: UserCard works fine if this is null, but I think making that change type-safe is probably complicated. Not doing it now for the sake of keeping the scope of this PR small.

jay-hodgson (approved):

PR #1924, opened 2022-12-01
SWC-6321: Allow disabled IconButton

nickgros:
- Pass `disabled` through to IconButton
- Specify flex alignItems
- Move dropdown tooltip placement


jay-hodgson (approved):
Ok

PR #1928, opened 2022-12-05
PORTALS-2437: focus on user/team search for markdown editor

yjyoo773:


src/lib/containers/UserSearchBoxV2.tsx
```diff
@@ -24,6 +25,7 @@ export type UserSearchBoxProps = {
   typeFilter?: TYPE_FILTER
   filterPredicate?: (item: UserGroupHeader) => boolean
   placeholder?: string
+  focusOnSelect?: boolean
```
yjyoo773, line 28: In case there are multiple SearchBoxes and do not want to focus on any or focus on one.

src/lib/containers/UserSearchBoxV2.tsx
```diff
@@ -108,6 +110,13 @@ const UserSearchBoxV2: React.FC<UserSearchBoxProps> = props => {
     typeFilter,
   )
 
+  const selectRef = React.useRef<any>(null)
```
yjyoo773, line 113: Set this as `any` because
- when left empty the below code `selectRef.current.focus()` would have a type `never`.
- when setting it as SelectInstance as the docs showed, the propTypes were incompatible with the `customSelectComponents`.
- vice versa when matching the `customSelectComponents` was incompatible with the RefAttributes.

src/lib/containers/UserSearchBoxV2.tsx
```diff
@@ -108,6 +110,13 @@ const UserSearchBoxV2: React.FC<UserSearchBoxProps> = props => {
...
+  React.useEffect(() => {
+    if (focusOnSelect) {
+      selectRef.current && selectRef.current.focus()
+    }
+  })
```
nickgros, line 118: Does it still work with this dependency array? I think you might have issues if you try to set the focus on every render.
```suggestion
  }, [focusOnSelect])
```

src/lib/containers/UserSearchBoxV2.tsx
```diff
@@ -86,6 +87,7 @@ const UserSearchBoxV2: React.FC<UserSearchBoxProps> = props => {
     filterPredicate,
     typeFilter,
     placeholder,
+    focusOnSelect,
```
nickgros, line 90: Setting a default value here shouldn't affect the logic but it's good practice to not allow a null/undefined value if it has no meaning
```suggestion
    focusOnSelect = false,
```

src/lib/containers/UserSearchBoxV2.tsx
yjyoo773: No it doesn't work with the dependency array.

nickgros (approved):

PR #1937, opened 2022-12-09
PORTALS-2441: Fix flaky test

nickgros:
Change SynapseTable.test.tsx so it doesn't log errors
Add `jest-fail-on-console` and use it in SynapseTable to prevent future issues. We should try to add this to as many tests as possible (or setupFilesOnEnv)


jay-hodgson (approved):

jay-hodgson:
Adding the handler prevents a race condition where MSW logs an error that we didn’t have a handler for a request _after_ the test suite has finished (which causes the "Cannot log after tests were done" error).

PR #1943, opened 2022-12-12
PORTALS-2418

yjyoo773:
Conditionally render the view terms button with the default to show button. This is due to match the design for PORTALS-2418 which shows the view terms beneath the Accept and Continue button.


src/lib/containers/TermsAndConditions.tsx
```diff
@@ -145,8 +148,8 @@ const TermsAndConditions: React.FunctionComponent<TermsAndConditionsProps> = ({
       <h3 className="page-header">Synapse Pledge</h3>
       <form>
         <label>
-          I affirm my commitment to all Synapse Governance policies for
-          responsible research and data handling (linked below), including:
+          I affirm my commitment to all <Link>Synapse Governance policies</Link>{' '}
```
yjyoo773, line 151: The link to Sage Governance Policies does not exist but will be a Confluence page

yjyoo773: Yes, I thinks the link provided would work

jay-hodgson: Could you verify with Adam that this link is ok (who should verify this with the Governance team)?

yjyoo773: Confirmed the link with Adam

src/lib/containers/TermsAndConditions.tsx
```diff
@@ -94,10 +95,13 @@ Please refer to our full <a target="_blank" href="https://help.synapse.org/docs/
 
 const TermsAndConditions: React.FunctionComponent<TermsAndConditionsProps> = ({
   onFormChange,
+  showTcTerm = true,
```
nickgros, line 98: 1. Not a great name, how about `showLinkToFullTC`?
2. We haven't been consistent about this in this codebase, but it is often better to make optional boolean props default to false. If the default behaviour here should be to show, then switch `show` to `hide` and make the default `false`. Then when rendering the component in JSX, if you wanted `hideLinkToFullTC` to be true, you can just write:

```jsx
<TermsAndConditions hideLinkToFullTC />
```

```suggestion
  hideLinkToFullTC = false,
```

nickgros (approved):
Only one issue with a variable name, please fix but doesn't block approval

PR #1962, opened 2023-01-10
SWC-6351

jay-hodgson:


src/lib/utils/SynapseClient.ts
```diff
@@ -364,7 +367,13 @@ const fetchWithExponentialTimeout = async <TResponse>(
   options: RequestInit,
   delayMs = 1000,
 ): Promise<TResponse> => {
-  let response = await fetch(url, options)
+  let response
+  try {
+    response = await fetch(url, options)
+  } catch (err) {
```
nickgros, line 373: Per MDN: 

> The Promise returned from fetch() won't reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing.

So this is definitely the right thing to do!

nickgros (approved):

PR #1963, opened 2023-01-11
SWC-6363

nickgros:
- Add ErrorChip component + stories
- Use ErrorChip component for EntityLink
- Add story for EntityLink
- Move error-related components into 'error' subfolder

https://user-images.githubusercontent.com/17580037/211850426-e89f22d7-cd13-495b-a1ce-b6a69dc72e05.mov

Current ACT dashboard:

image

After change:

image


src/lib/containers/error/ErrorChip.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const variant = loginError ? 'warning' : 'error'
+  let avatar
+  let icon
+  if (loginError || accessDenied) {
+    avatar = (
+      <Avatar sx={{ bgcolor: `${variant}.main` }}>
+        <LockOutlined sx={{ color: 'white', width: '18px' }} />
+      </Avatar>
+    )
```
nickgros, line 40: The avatar component is the easiest way in MUI to give an icon a circular background

src/lib/containers/error/ErrorChip.tsx
```diff
@@ -0,0 +1,64 @@
+import { Avatar, Chip, Tooltip } from '@mui/material'
+import { ErrorOutlined, LockOutlined } from '@mui/icons-material'
```
nickgros, line 2: Not using IconSvg because it causes some strange spacing issues compared to just using the icons straight from MUI

src/lib/containers/EntityLink.tsx
```diff
@@ -75,6 +76,9 @@ export const EntityLink = (props: EntityLinkProps) => {
       )
     }
   } else if (error) {
+    if (entityId) {
+      return <ErrorChip chipText={entityId} error={error} />
+    }
```
nickgros, line 81: Here's the first place we'll use this component. Show the new error chip if we encounter an error and have an ID that we can show in the chip.

jay-hodgson (approved):

PR #1971, opened 2023-01-20
Attempt to fix flaky tests: FacetNav

nickgros:
Tests in FacetNav are failing for me in TravisCI, not entirely sure why. I'm refactoring the tests (and one component change) to get rid of error logs that are either triggering or obscuring the failure reason.

- Use react-query instead of useEffect in FacetNavPanel, mostly to prevent error logs in test
- Attempt to make tests less flaky by refactoring to avoid error and warning logs

### DIFF
--- a/src/__tests__/lib/containers/widgets/facet-nav/FacetNav.test.tsx
+++ b/src/__tests__/lib/containers/widgets/facet-nav/FacetNav.test.tsx
@@ -21,6 +21,7 @@ import {
 } from '../../../../../lib/utils/synapseTypes'
 import testData from '../../../../../mocks/mockQueryResponseDataWithManyEnumFacets'
 import { server } from '../../../../../mocks/msw/server'
+import failOnConsole from 'jest-fail-on-console'
 
 const lastQueryRequest: QueryBundleRequest = {
   concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
@@ -93,6 +94,7 @@ function init(
 }
 
 describe('facets display hide/show', () => {
+  failOnConsole()
   beforeAll(() => server.listen())
   afterEach(() => server.restoreHandlers())
   afterAll(() => server.close())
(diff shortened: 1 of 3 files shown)


src/__tests__/lib/containers/widgets/facet-nav/FacetNav.test.tsx
```diff
@@ -93,21 +94,22 @@ function init(
 }
 
 describe('facets display hide/show', () => {
+  failOnConsole()
```
nickgros, line 97: In both test suites, add `failOnConsole`.

src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -368,27 +368,27 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = (
...
-    }
-    return () => {
-      isMounted = false
-    }
-  }, [facetToPlot, data, index, plotType, accessToken, getColumnType])
+      ),
+    {
+      enabled: !!facetToPlot,
+    },
+  )
```
nickgros, line 391: One small logic change; the `useEffect` + `isMounted` pattern can cause cause delayed updates which triggers a warning in tests. react-query has already solved this general problem, so let's just wrap our async call in `useQuery`.

src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -542,6 +542,7 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = (
           )}
           <div
             className={`FacetNavPanel__body${isModalView ? '--expanded' : ''}`}
+            role="graphics-object"
```
nickgros, line 545: adding a reasonable accessible role for test purposes

jay-hodgson (approved):

PR #1975, opened 2023-02-03
SWC-6384

jay-hodgson:


jay-hodgson:
Pull over [code from the old SWC HtmlPreview implementation](https://github.com/Sage-Bionetworks/SynapseWebClient/commit/35b7498e068534b6b87168831a6b6d8af7867d8d#diff-edd8aca66c9deca98ff92619ddd37932595a865e52798fc5a8eba128800291cbL139) to dynamically set the iframe height.

jay-hodgson:
Integration tests
Before: 
Screen Shot 2023-02-02 at 5 06 39 PM
After (copying newly built SRC into running SWC):
Screen Shot 2023-02-02 at 5 07 46 PM

src/lib/containers/FilePreview/HtmlPreview/HtmlPreview.tsx
```diff
@@ -40,12 +40,30 @@ export type HtmlPreviewProps = {
...
+      if (!frameHeight || Math.abs(newHeightPx - frameHeight) > 70) {
+        setFrameHeight(newHeightPx + 50)
+      }
+    }
+  }
+  useEffect(() => {
+    setInterval(() => {
+      updateHeight()
+    }, 500)
+  }, [])
```
nickgros, line 66: Should probably rerun this effect if the raw HTML changes in case we do something funny, like pass an empty string and change it out after the resource loads.

```suggestion
  }, [rawHtml])
```

nickgros:
Moving over here: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/13

nickgros (approved):

